PR #25, opened 2023-10-02
[Proto] Mutate state in PyQOperation

dominikandreasseitz:
Instead of performing expensive tensordotting when propagating the state through primitive gates, lets use operations which mutate the state.


dominikandreasseitz:
moving to pyq

PR #28, opened 2023-10-03
[Infra] Add Publish Step

dominikandreasseitz:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,41 @@
+name: Upload release to PyPI
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
```
Roland-djee, line 10: I guess we use the 22.04 image in other actions. Might be best to harmonise to `latest`.

Roland-djee (approved):
Thanks @nmheim. LGTM.

.github/workflows/publish.yml
nmheim: good point, will do!

madagra (approved):
Thanks, @nmheim, LGTM.

PR #57, opened 2023-10-09
[Docs,Infra] Build documentation

nmheim:
It was easiest to pre-build/commit to a branch via [mike](https://github.com/jimporter/mike), which gives us, in addition to not having to set up anything special on RTD, versioned docs. Given that that is already directly compatible with github pages I decided to host our docs there for now.


dominikandreasseitz:
thanks a lot for this!

Roland-djee (approved):
LGTM and thanks do much @nmheim. Just a question: do we envision to use the dockerhub image at some point ?

PR #70, opened 2023-10-11
[Testing] Use pytest-xdist

dominikandreasseitz:


madagra (approved):
Finally, long-sought. Thanks, @dominikandreasseitz.

PR #77, opened 2023-10-11
[Docs] Minor docs fixes

### LINKED ISSUE #71: Plotting of examples in the documentation fails
All the code blocks with 'plot' in them don't work out of the box (like the ones in [Digital-analog quantum computing]) for:

`NameError: name 'plot' is not defined. Did you mean: 'float'?`

Also, in some of the blocks torch is used but not loaded at each block (and I guess they should be independent?)

Used with matplotlib is 3.7.3.

nmheim:
Closes #71 

- [x] prevent plots from popping up during local docs build
- [x] show plot in qcl example
- [x] show plotting function definitions in qcl example

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   test_qadence_ubuntu:
-    name: Test Qadence docs (ubuntu)
+    name: docs (ubuntu)
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test_examples.yml
nmheim, line 1: renaming so that the pipelines in MRs are more readable

dominikandreasseitz (approved):

PR #78, opened 2023-10-11
Add transpile.reverse

dominikandreasseitz:


qadence/transpile/reverse.py
```diff
@@ -0,0 +1,43 @@
...
+    """Reverses a block if its a ChainBlock."""
+    if isinstance(block, ChainBlock):
+        blk = chain(*(reverse(b) for b in reversed(block.blocks)))
+        return tag(blk, block.tag) if block.tag is not None else blk
+    else:
+        return block
+
+
+@reverse.register(QuantumCircuit)  # type: ignore[attr-defined]
+def _(circuit: QuantumCircuit) -> QuantumCircuit:
```
nmheim, line 37: we have a `reverse_program` in `transpile/circuit.py` (which is not implemented). Should we call this function `reverse_program` and remove the other not-implemented one? Would also argue for a name change because `reversed` already exists in python

qadence/transpile/reverse.py
```diff
@@ -0,0 +1,43 @@
...
+    """This method reverses a circuit.
+
+    Returns:
+        A reversed QuantumCircuit.
+    """
```
nmheim, line 42: I'm not sure if these doc strings show up in `help(reverse)` (I am suspecting only the top level one might). Could we add an example here as well?

nmheim (changes requested):

qadence/transpile/reverse.py
dominikandreasseitz: removed

Roland-djee (approved):
Thanks @dominikandreasseitz

madagra (approved):
Looks good to me.

tests/qadence/test_dagger.py
```diff
@@ -147,3 +149,9 @@ def test_all_self_adjoint_blocks(block: Tuple[int, AbstractBlock]) -> None:
 )
 def test_composite_blocks_no_fails(block: AbstractBlock) -> None:
     assert isinstance(block.dagger(), AbstractBlock)
+
+
+@given(st.restricted_circuits())
+@settings(deadline=None)
+def test_circuit_dagger(circuit: QuantumCircuit) -> None:
+    circuit == circuit.dagger().dagger()
```
nmheim, line 157: I think there was a test somewhere that did not just do `.dagger().dagger()`, but explicitly checked a few (nested) circuits. could we get that back? here the `dagger()` could also just be a no-op

nmheim (changes requested):

tests/qadence/test_dagger.py
dominikandreasseitz: added explicit test

dominikandreasseitz:
@nmheim recheck if you want

nmheim (approved):
LGTM!

PR #91, opened 2023-10-12
[Docs] Fixed and improved QCL example

madagra:
Fixed the output of the QCL example after introducing the 2-dimensional output in the quantum model.

### DIFF
--- a/docs/qml/qcl.md
+++ b/docs/qml/qcl.md
@@ -9,7 +9,7 @@ shows how to fit the $\sin(x)$ function in the $[-1, 1]$ domain.
 
 In the following, train and test data are defined.
 
-```python exec="on" source="material-block" session="qcl" result="json"
+```python exec="on" source="material-block" session="qcl"
 import torch
 from torch.utils.data import random_split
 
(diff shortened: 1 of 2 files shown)


docs/qml/qcl.md
```diff
@@ -10,21 +10,16 @@ shows how to fit the $\sin(x)$ function in the $[-1, 1]$ domain.
 In the following, train and test data are defined.
 
 ```python exec="on" source="material-block" session="qcl" result="json"
```
nmheim, line 12:
```suggestion
```python exec="on" source="material-block" session="qcl"
```

nmheim: removes the empty output box of the code block

docs/qml/qcl.md
```diff
@@ -87,25 +84,27 @@ using a mean-square error loss, Adam optimizer. Training is performend on the GP
...
 
     optimizer.zero_grad()
 
-    loss = mse_loss(model(input_values), y_train)
+    # given a `n_batch` number of input points and a `n_observables`
+    # number of input observables to measure, the QNN returns
+    # an output of the following shape: [n_batch x n_observables]
+    # given that there is only one observable, a squeeze is applied to get
+    # a 1-dimensional tensor
+    loss = mse_loss(model(values=x_train).squeeze(), y_train)
```
nmheim, line 107: do we want to add an `assert loss < 1e-3` at the end to make sure this will error if it does not converge in a future docs build?

nmheim (approved):
tiny comments, pre-approving this :)

PR #103, opened 2023-10-16
[Fix] TransformedModule serialization

dominikandreasseitz:


tests/ml_tools/test_transformed_module.py
```diff
@@ -42,7 +42,7 @@ def quantum_circuit(n_qubits: int = 2, depth: int = 1) -> QuantumCircuit:
 
 
 def get_qnn(n_qubits: int, depth: int) -> QNN:
-    observable = total_magnetization(n_qubits)
+    observable = hamiltonian_factory(n_qubits, detuning=Z)
```
Roland-djee, line 45: Do we go fo replacing all `total_magnetization` constructors with calls to `hamiltonian_factory` ?

Roland-djee (approved):
LGTM.

tests/ml_tools/test_transformed_module.py
dominikandreasseitz: yes, i will do it in the MR where i clean up the tests

qadence/ml_tools/models.py
```diff
@@ -181,7 +181,7 @@ def _transform_x(self, x: dict[str, torch.Tensor] | Tensor) -> dict[str, Tensor]
 
         else:
             assert isinstance(self.model, torch.nn.Module) and isinstance(x, Tensor)
-            return self._input_scaling * (x + self._input_shifting)
+            return self._input_scaling * x + self._input_shifting
```
nmheim, line 184: why do we want to make this change? this will break backwards compatibility...

dominikandreasseitz: readded!

PR #107, opened 2023-10-16
[Fix] corner case of `chain_single_qubit_ops`

nmheim:
When the krons in a chain of krons where not covering the whole qubit support `chain_single_qubit_ops` was failling

### DIFF
--- a/qadence/transpile/block.py
+++ b/qadence/transpile/block.py
@@ -406,7 +406,14 @@ def chain_single_qubit_ops(block: AbstractBlock) -> AbstractBlock:
     ```
     """
     if is_chain_of_primitivekrons(block):
-        return kron(*map(lambda bs: chain(*bs), zip(*block)))  # type: ignore[misc]
+        try:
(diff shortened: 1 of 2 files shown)


qadence/transpile/block.py
```diff
@@ -407,22 +406,7 @@ def chain_single_qubit_ops(block: AbstractBlock) -> AbstractBlock:
...
-            chains.append(chain(*prim_gates))
-        try:
-            return kron(*chains)
-        except Exception as e:
-            logger.debug(
-                f"Unable to transpile {block} using chain_single_qubit_ops\
-                         due to {e}. Returning original circuit."
-            )
-            return block
+        return kron(*map(lambda bs: chain(*bs), zip(*block)))  # type: ignore[misc]
```
madagra, line 409: Cool :)

madagra (approved):
Thanks for fixing this, @nmheim. I think this bugfix deserves a new release v1.0.2.

dominikandreasseitz (approved):

Issue #93, opened 2023-10-13
Add differentiable backend using Jax


madagra:
## Motivations

Currently, we only support PyTorch as `DifferentiableBackend`. However, it is useful to have a more general interface to support other backends such as Jax. The benefits of Jax over PyTorch are several:

- fully functional interface
- out-of-the-box JIT compilation of models also with complex number
- usually performance of Jax models is significantly higher than PyTorch even without compilation

## Proposal

Adapt the current `DifferentiableBackend` interface to support Jax as a differentiable backend for the code. In essence, one should be able to write an abstract program with `qadence` and its execution should yield a Jax array in case that Jax differentiable backend is selected. In this way, `qadence` models will be compatible with any Jax-based optimization workflow and library.

To bear in mind while performing this implementation:

- [ ] the `QuantumModel` and generally high-level interface of the package should not change
- [ ] It is probably necessary to introduce a generic Tensor datatype that handles tensors in both Jax and PyTorch. If possible, do not proceed the Pennylane way, i.e. using the [autoray ](https://github.com/jcmgray/autoray)package. The interface is not clean and the package is not really maintained currently.
- [ ] make sure that also the GPSR differentiation mode is taken into account. If too long, this can be handled in a separate issue.

dominikandreasseitz:
@Roland-djee

Roland-djee:
@dominikandreasseitz This can be closed right ?

PR #111, opened 2023-10-17
[Engines, Backends] Add JAX Engine and Horqrux Backend

dominikandreasseitz:
Addresses https://github.com/pasqal-io/qadence/issues/93

1. This MR adds an `Engine` to qadence which is uses to instantiate a DifferentiableBackend which can be either JAX or torch.
2. Each backend gets a `engine` field which is uses by `backend_factory` to import either a torch or jax differentiable backend.
3. Add a new backend called 'horqrux'  https://github.com/pasqal-io/horqrux, a differntiable state vector simulator in JAX 
4. Add diff_modes GPSR and AD to horqrux backend
5. Add low-level tutorial to train qadence circuits in JAX


qadence/engines/jax/differentiable_expectation.py
```diff
@@ -0,0 +1,99 @@
...
+                self.circuit.abstract.n_qubits, "0" * self.circuit.abstract.n_qubits
+            )
+
+        def _expectation_fn(
+            state: Array, values: dict, psr_params: dict, uuid_to_eigen: dict
+        ) -> Array:
+            wf = self.circuit.native.forward(state, values)
+            return observable.native.forward(wf, values)
+
+        @custom_vjp
```
atiyo, line 56: If only using `grad` then `custom_vjp` is enough, but if you wanted to support forward mode AD as well it makes sense to implement a `custom_jvp` which would support both forward and backward mode.

Jax would automatically calculate the VJP from the JVP by transposition.

qadence/engines/jax/differentiable_expectation.py
```diff
@@ -0,0 +1,99 @@
...
+        def _expectation(
+            state: Array, values: dict, psr_params: dict, uuid_to_eigen: dict
+        ) -> Array:
```
atiyo, line 59: Is this intended to be used for finite shots too? If so, then how would `jax` get the relevant `PRNGKey` to introduce noise to the shots?

dominikandreasseitz: i am planning to add it but it wont be in the first version of the JAX engine in qadence

qadence/engines/jax/differentiable_expectation.py
dominikandreasseitz: great idea @atiyo thank you :)

tests/backends/horqrux/test_quantum_horqrux.py
```diff
@@ -0,0 +1,102 @@
...
+        def _exp_fn(value: Array) -> Array:
+            vals = {list(param_names)[0]: value}
+            return hq_bknd.expectation(hq_circ, hq_obs, vals)
+
+        d1fdx = grad(_exp_fn)
+        d2fdx = grad(d1fdx)
+        d3fdx = grad(d2fdx)
+        jd3fdx = jit(d3fdx)
+        grad_dict[diff_mode] = jd3fdx(jnp.pi / 2)
+    assert jnp.allclose(grad_dict["ad"], grad_dict["gpsr"])
```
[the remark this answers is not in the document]
dominikandreasseitz: definitely! thanks :)

qadence/engines/jax/differentiable_expectation.py
```diff
@@ -0,0 +1,99 @@
...
+
+        shift = jnp.pi / 2
+        spectral_gap = 2.0
+
+        def _expectation_bwd(
+            res: Tuple[Array, ParamDictType, dict, dict[str, Array]], v: Array
+        ) -> Any:
+            state, values, psr_params, uuid_to_eigen = res
+            grads = {}
+            for param_name, eigenvals in uuid_to_eigen.items():
```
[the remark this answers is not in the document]
GJBoth: @atiyo Indeed you're wrong :-). You don't write higher-dimensional jacobians, but the jacobian for the first-order function.

pyproject.toml
```diff
@@ -41,7 +41,15 @@ dependencies = [
...
-    "matplotlib"
+    "matplotlib",
+    "jax",
+    "flax",
+    "optax",
+    "jaxopt",
+    "einops",
+    "numpy",
+    "sympy2jax",
+    "horqrux @ git+https://github.com/pasqal-io/horqrux"
```
nmheim, line 52: should these be optional?

qadence/parameters.py
```diff
@@ -204,7 +205,15 @@ def heaviside_func(x: Tensor, _: Any) -> Tensor:
...
-    return SymPyModule(expressions=[sympy.N(expr)], extra_funcs=extra_funcs)
+    return torchSympyModule(expressions=[sympy.N(expr)], extra_funcs=extra_funcs)
+
+
+def make_differentiable(expr: Expr, engine: Engine) -> torchSympyModule | JaxSympyModule:
+    if engine == Engine.JAX:
+        return JaxSympyModule(expr)
+    else:
+        extra_funcs = {sympy.core.numbers.ImaginaryUnit: 1.0j}
+        return torchSympyModule(expressions=[sympy.N(expr)], extra_funcs=extra_funcs)
```
nmheim, line 216: this seems to be part of the `Heaviside` docs failure

dominikandreasseitz: yeah i noticed, will fix later. but there was something else failing in the docs

nmheim, line 216: `Heaviside` is missing in the `extra_funcs`

pyproject.toml
dominikandreasseitz: that was being discussed before, if qadence was either full torch or jax, we could do that. but now its kinda both

qadence/engines/jax/differentiable_expectation.py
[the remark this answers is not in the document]
atiyo: Yeah, I got myself confused by wondering how higher-order gradients would get access to different random keys in the case of finite shots and shot noise. Nevermind me 😛

qadence/engines/jax/differentiable_backend.py
```diff
@@ -0,0 +1,151 @@
+from __future__ import annotations
+
+from collections import Counter
+from typing import Any
+
+from jax import Array
+
+from qadence.backend import Backend as QuantumBackend
```
Roland-djee, line 8: Not sure this is necessary, unless I'm mistaken. In `DifferentiableBackend`, it is `Backend`.

qadence/engines/jax/differentiable_backend.py
```diff
@@ -0,0 +1,151 @@
...
+            mitigation=mitigation,
+            endianness=endianness,
+        )
+
+    def circuit(self, circuit: QuantumCircuit) -> ConvertedCircuit:
+        parametrized_blocks = list(uuid_to_block(circuit.block).values())
+        non_prim_blocks = filter(lambda b: not isinstance(b, PrimitiveBlock), parametrized_blocks)
+        if len(list(non_prim_blocks)) > 0:
+            raise ValueError(
+                "The circuit contains non-primitive blocks that are currently not supported by the "
```
Roland-djee, line 122: Out of curiosity, what would be an example of a non-primitive parametrised block ? Multi-parametrised ?

qadence/engines/jax/differentiable_backend.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> None:
+        super().__init__(backend=backend, engine=backend.engine)
+        self.diff_mode = diff_mode
+        self.psr_args = psr_args
+
+    def run(
+        self,
+        circuit: ConvertedCircuit,
+        param_values: ParamDictType = {},
+        state: Array | None = None,
```
Roland-djee, line 45: Shouldn't that be `ArrayLike` ? At least, it is the case in `TorchBackend`.

tests/backends/horqrux/test_quantum_horqrux.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@pytest.mark.parametrize("block", [RX(0, 1.0), RY(0, 3.0), RZ(0, 4.0), X(0), Y(0), Z(0)])
```
Roland-djee, line 73: Isn't it possible to use a strategy to pick up randomly supported gates in Horqrux ?

Roland-djee (commented):
Thank you so much for this @dominikandreasseitz it feels really clean. Minor comments from my side. Happy to approve whenever commented and addressed.

pyproject.toml
dominikandreasseitz: its split now

qadence/backends/api.py
```diff
@@ -46,7 +47,15 @@ def backend_factory(
     set_backend_config(backend_inst, diff_mode)
 
     if diff_mode is not None:
-        backend_inst = DifferentiableBackend(backend_inst, DiffMode(diff_mode))
+        if backend_inst.engine == Engine.TORCH:
```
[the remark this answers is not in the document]
dominikandreasseitz: true, but in contrast to Backends, we dont really plan to have more than 2 engines, do we?

qadence/blocks/embedding.py
```diff
@@ -16,24 +14,38 @@
...
     return list(set(x))
 
 
 def embedding(
-    block: AbstractBlock, to_gate_params: bool = False
-) -> tuple[StrTensorDict, Callable[[StrTensorDict, StrTensorDict], StrTensorDict],]:
-    """Construct embedding function.
+    block: AbstractBlock, to_gate_params: bool = False, engine: Engine = Engine.TORCH
+) -> tuple[ParamDictType, Callable[[ParamDictType, ParamDictType], ParamDictType],]:
+    """Construct embedding function which maps user-facing parameters to either *expression-level*.
```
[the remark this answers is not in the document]
dominikandreasseitz: its the pydocstringformatter :D

qadence/backends/api.py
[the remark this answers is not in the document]
dominikandreasseitz: added

qadence/engines/jax/differentiable_backend.py
dominikandreasseitz: good question

tests/backends/horqrux/test_quantum_horqrux.py
dominikandreasseitz: sure, but we're already picking random gates in the test_backends tests. and i want to make sure all are tested

qadence/backends/jax_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+def values_to_jax(param_values: dict[str, Tensor]) -> dict[str, Array]:
+    return {key: jnp.array(value.detach().numpy()) for key, value in param_values.items()}
+
+
+def jaxify(expr: Expr) -> JaxSympyModule:
+    return JaxSympyModule(expr)
+
+
+def unhorqify(state: Array) -> Array:
```
GJBoth, line 34: loving this name. However, doesn't pyqtorch do the same?

qadence/backends/jax_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    """Convert a state of shape [2] * n_qubits + [batch_size] to (batch_size, 2**n_qubits)."""
+    return jnp.ravel(state)
+
+
+def uniform_batchsize(param_values: ParamDictType) -> ParamDictType:
```
GJBoth, line 39: So wait, this allows people to explicitly batch a certain parameter, and the rest follows? Are we sure we want to allow this? Cause to me this feels like a foot gun...

qadence/backends/jax_utils.py
```diff
@@ -0,0 +1,45 @@
...
+from jax import Array, device_get
+from sympy import Expr
+from sympy2jax import SymbolicModule as JaxSympyModule
+from torch import Tensor, cdouble, from_numpy
+
+from qadence.types import ParamDictType
+
+
+def jarr_to_tensor(arr: Array, dtype: Any = cdouble) -> Tensor:
+    return from_numpy(device_get(arr)).to(dtype=dtype)
```
GJBoth, line 15: Are you sure you need a `device_get` here? Usually you only use that when you're explicitly doing things...

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -252,12 +252,12 @@ def sparse_operation(state: Tensor, values: dict[str, Tensor] = None) -> Tensor:
                 convert_block(block, n_qubits, config),
             )
 
-    def forward(self, state: Tensor, values: dict[str, Tensor]) -> Tensor:
-        return pyq.overlap(state, self.operation(state, values))
-
     def run(self, state: Tensor, values: dict[str, Tensor]) -> Tensor:
         return self.operation(state, values)
 
+    def forward(self, state: Tensor, values: dict[str, Tensor]) -> Tensor:
```
GJBoth, line 258: Not sure why you changed this in this PR but fine for me?

qadence/backends/api.py
```diff
@@ -44,9 +49,22 @@ def backend_factory(
 
     # Set backend configurations which depend on the differentiation mode
     set_backend_config(backend_inst, diff_mode)
-
+    # Wrap the quantum Backend in a DifferentiableBackend if a diff_mode is passed.
     if diff_mode is not None:
-        backend_inst = DifferentiableBackend(backend_inst, DiffMode(diff_mode))
+        try:
```
GJBoth, line 54: We'll always need an engine, independent of diffmode no?

examples/backends/low_level/horqrux_backend.py
```diff
@@ -0,0 +1,67 @@
...
+    hq_circ, hq_obs, hq_embedfn, hq_init_params = hq_bknd.convert(circ, obs)
+    inputs = {}
+    embedded_params = hq_embedfn(hq_init_params, inputs)
+    optimizer = optax.adam(learning_rate=0.001)
+
+    param_names = embedded_params.keys()
+    param_values = embedded_params.values()
+    init_array = jnp.array(jnp.concatenate([arr for arr in param_values]))
+    opt_state = optimizer.init(init_array)
+    param_array = init_array
```
GJBoth, line 31: This is all not necessary in Jax... we can just give the embedded params directly to the optimiser, no need to unpack.

examples/backends/low_level/horqrux_backend.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def optimize_step(params: Array, opt_state: Array, grads: Array):
+        updates, opt_state = optimizer.update(grads, opt_state, params)
+        new_params = optax.apply_updates(params, updates)
+        return new_params, opt_state
+
+    def _exp_fn(values: Array) -> Array:
```
GJBoth, line 38: Why all the _?

examples/backends/low_level/horqrux_backend.py
```diff
@@ -0,0 +1,67 @@
...
+        vals = {k: v for k, v in zip(param_names, values)}
+        return hq_bknd.expectation(hq_circ, hq_obs, vals)
+
+    init_pred = _exp_fn(init_array)
+
+    def _loss(param_arr: Array, y_true: Array) -> Array:
+        expval = _exp_fn(param_arr)
+        return (expval - y_true) ** 2
+
+    loss_and_grads = value_and_grad(_loss)
```
GJBoth, line 48: This should just go in the train step :)

examples/backends/low_level/horqrux_backend.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _train_step(param_array: Array, opt_state, y_true: Array) -> tuple:
+        loss, grads = loss_and_grads(param_array, y_true)
+        updated_param_array, updated_opt_state = optimize_step(param_array, opt_state, grads)
+        return loss, updated_param_array, updated_opt_state
+
+    train_step = jit(_train_step)
```
GJBoth, line 55: You can jit the trainstep with a decorator!

qadence/blocks/embedding.py
```diff
@@ -16,24 +14,38 @@
...
+from qadence.types import ArrayLike, DifferentiableExpression, Engine, ParamDictType, TNumber
 
-StrTensorDict = dict[str, Tensor]
+
+def _concretize_parameter(engine: Engine) -> Callable:
+    if engine == Engine.JAX:
+        import jax.numpy as jnp
+
+        def concretize_parameter(value: TNumber, requires_grad: bool) -> ArrayLike:
+            fn, dtype = (jnp.array, jnp.float64) if requires_grad else (np.array, np.float64)
```
GJBoth, line 26: I think jax automatically promotes inputs to jax, so just returning a numpy array *will not* do what you want it here...

qadence/engines/differentiable_backend.py
```diff
@@ -0,0 +1,144 @@
...
+from qadence.circuit import QuantumCircuit
+from qadence.measurements import Measurements
+from qadence.mitigations import Mitigations
+from qadence.noise import Noise
+from qadence.types import ArrayLike, DiffMode, Endianness, Engine, ParamDictType
+
+
+@dataclass(frozen=True, eq=True)
+class DifferentiableBackend(ABC):
+    """The abstract class that defines the interface for differentaible backends."""
```
GJBoth, line 20: typo - should be differentiable :)

qadence/engines/jax/differentiable_backend.py
```diff
@@ -0,0 +1,73 @@
...
+    DifferentiableBackend as DifferentiableBackendInterface,
+)
+from qadence.engines.jax.differentiable_expectation import DifferentiableExpectation
+from qadence.measurements import Measurements
+from qadence.mitigations import Mitigations
+from qadence.noise import Noise
+from qadence.types import ArrayLike, DiffMode, Endianness, Engine, ParamDictType
+
+
+class DifferentiableBackend(DifferentiableBackendInterface):
```
GJBoth, line 14: Why not call it JaxDifferentiableBackend, or something? The import renames the thing with the same name, can be confusing :)

qadence/engines/jax/differentiable_expectation.py
```diff
@@ -0,0 +1,88 @@
...
+                state,
+                values,
+                psr_params,
+            )
+
+        def _expectation_bwd(res: Tuple[Array, ParamDictType, dict], v: Array) -> Any:
+            state, values, psr_params = res
+            grads = {}
+            # Hardcoding the single spectral_gap to 2. for jax.lax jitting reasons.
+            spectral_gap = jnp.array(2.0, dtype=jnp.float64)
```
GJBoth, line 74: Not sure why you need to hardcode the spectral gap? You can try and set it as a static_argument if it changes shape (i.e. more spectral gaps.)

qadence/engines/jax/differentiable_expectation.py
```diff
@@ -0,0 +1,88 @@
...
+            shift = jnp.pi / 2
+            for param_name, _ in psr_params.items():
+                shifted_values = values.copy()
```
GJBoth, line 77: This works but I think you should be able do to something nicer with `tree_map` (no priority though)

GJBoth (changes requested):
Awesome work, very happy to see this :-)

Overall it looks pretty good I think, I made a few commments.
One main thing I'm missing (but also because I haven't worked on qadence in a while) is the general architecture - we now have engines, backends, multiple diff methods etc. I'd appreciate some comments/ documentation on how all of this now communicates. For example, can I use Pyqtorch with the Jax engine?

Another one is examples - I think we can improve the given example and make it more concise.

One final thing is that jax handles all kinds of datastructures easily. So instead of flattening -> making into an array -> ops -> go back, we can just give the dictionary / whatever data structure.

examples/backends/low_level/horqrux_backend.py
dominikandreasseitz: fixed

examples/backends/low_level/horqrux_backend.py
dominikandreasseitz: jep, changed

examples/backends/low_level/horqrux_backend.py
dominikandreasseitz: done

examples/backends/low_level/horqrux_backend.py
dominikandreasseitz: done

qadence/backends/api.py
dominikandreasseitz: hm no because you can run things in "no_grad" mode

qadence/backends/horqrux/backend.py
```diff
@@ -0,0 +1,219 @@
...
+        circuit: ConvertedCircuit,
+        param_values: ParamDictType = {},
+        state: Any = None,
+        endianness: Endianness = Endianness.BIG,
+        horqify_state: bool = True,
+        unhorqify_state: bool = True,
+    ) -> ArrayLike:
+        n_qubits = circuit.abstract.n_qubits
+        if state is None:
+            state = prepare_state(n_qubits, "0" * n_qubits)
```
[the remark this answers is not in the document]
dominikandreasseitz: haha yea fully, agree. some horqrux legacy i didnt want to change

qadence/backends/jax_utils.py
dominikandreasseitz: i think i added it initially because stuff was failing without :D i can check it

qadence/backends/jax_utils.py
dominikandreasseitz: yeah, but for torch tensors :D

qadence/backends/jax_utils.py
dominikandreasseitz: hm not really a certain one. its more like that we have fixed parameters in the param dict. and those will have batch_size =1 , if we want to vmap over batches of featureparameters, we need to "blow up" the other parameters to the same batch_size

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: haha you know while youre cleaning the kitchen you end up cleaning the bathroom too...

qadence/engines/differentiable_backend.py
dominikandreasseitz: thanks

qadence/engines/jax/differentiable_backend.py
dominikandreasseitz: i had it exactly like that :D but we do the same thing for the backends (meaning calling each individual backend class "Backend" instead of "{X}Backend"). i am open to both but i think @Roland-djee prefers it this way

qadence/engines/jax/differentiable_expectation.py
dominikandreasseitz: saved for later

qadence/engines/jax/differentiable_expectation.py
dominikandreasseitz: haha because jit didnt like me not telling it what shapes the spectral gap arrays will have. thanks for the suggestion

qadence/engines/jax/differentiable_expectation.py
dominikandreasseitz: yeah this will be done if we adopt a proper data type like DAG etc for the quantum circuits. for now i would leave it

qadence/backend.py
```diff
@@ -114,6 +114,7 @@ class Backend(ABC):
     is_remote: bool
     with_measurements: bool
     native_endianness: Endianness
+    engine: Engine
```
nmheim, line 117: can we add an entry in the docstring for this one? weird that the docs linting is not failing here, actually.

dominikandreasseitz: oh good catch, yeah, indeed weird..

qadence/backends/horqrux/backend.py
```diff
@@ -0,0 +1,216 @@
...
+        ops = convert_block(circuit.block, n_qubits=circuit.n_qubits, config=self.config)
+        return ConvertedCircuit(
+            native=HorqruxCircuit(ops), abstract=circuit, original=original_circ
+        )
+
+    def observable(self, observable: AbstractBlock, n_qubits: int) -> ConvertedObservable:
+        transpilations = [
+            flatten,
+            scale_primitive_blocks_only,
+        ]
```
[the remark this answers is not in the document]
dominikandreasseitz: @madagra was against that

qadence/backends/horqrux/config.py
```diff
@@ -0,0 +1,26 @@
...
+from qadence.transpile import (
+    blockfn_to_circfn,
+    flatten,
+    scale_primitive_blocks_only,
+)
+
+logger = get_logger(__name__)
+
+
+def default_passes(config: Configuration) -> list[Callable]:
```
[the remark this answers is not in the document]
dominikandreasseitz: lets do that in another MR since it affects all backends

qadence/engines/differentiable_backend.py
```diff
@@ -0,0 +1,144 @@
...
+            state: Initial state.
+            measurement: Optional measurement protocol. If None, use
+                exact expectation value with a statevector simulator.
+            noise: A noise model to use.
+            mitigation: The error mitigation to use.
+            endianness: Endianness of the resulting bit strings.
+        """
+        raise NotImplementedError(
+            "A DifferentiableBackend needs to override the expectation method."
+        )
```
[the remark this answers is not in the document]
dominikandreasseitz: all childclasses must override this method

qadence/backend.py
dominikandreasseitz: actually several other, non-MR related args are missing too

qadence/backends/api.py
dominikandreasseitz: we tell the backend_factory essentially "i want to compute grads" by passing a diff_mode, if none is passed, then torch stuff is run in no_grad mode and for JAX its irrelevant since you only do grads by manually invoking "grad" "value_and_grad" etc. however, if you choose gpsr as diffmode, you need the PSR Expectation which sits in the DiffBackend. lastly,  see "engine" more as an alias for "give me the differentiablebackend for backend "X" which uses autodiff engine "Y")

qadence/types.py
```diff
@@ -370,3 +373,12 @@ class OpName(StrEnum):
...
     """The projector operation."""
+
+
+class Engine(StrEnum):
+    TORCH = "torch"
+    JAX = "jax"
+
+
+ParamDictType = dict[str, ArrayLike]
+DifferentiableExpression = Callable[..., ArrayLike]
```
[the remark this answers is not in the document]
dominikandreasseitz: i like having type aliases like this since it makes the code more readable. i could also just use "Callable" if you prefer.

i introduced it to be able to type:
```
    embeddings: dict[sympy.Expr, DifferentiableExpression] = {
        expr: make_differentiable(expr=expr, engine=engine)
        for expr in unique_expressions
        if not expr.is_number
    }
```
but again, i can just do "Callable" (which is not really correct right)

qadence/backends/horqrux/backend.py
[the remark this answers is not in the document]
nmheim: Hmm I don’t see why it’s better to have them hardcoded by I might be missing something

qadence/engines/differentiable_backend.py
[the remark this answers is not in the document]
nmheim: Yes, but why is it not implemented? I thought horqrux was supporting expectations

qadence/backends/horqrux/backend.py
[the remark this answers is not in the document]
dominikandreasseitz: citing @madagra " But the thing is that observable could get different passes than the circuit and I decided not to allow that in the configuration. Observable transpilation is not really standard AFAIK and it does not make much sense to expose to the user."

qadence/engines/differentiable_backend.py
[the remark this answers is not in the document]
dominikandreasseitz: i think youre looking at the wrong file. this is the abstract diff backend class. horqrux implements everything pyq does too

qadence/extensions.py
```diff
@@ -18,6 +18,20 @@
...
+    """Fallback function for native Qadence available engines if extensions is not present."""
+    res = {}
+    for engine in Engine.list():
+        module_path = f"qadence.engines.{engine}.differentiable_backend"
+        try:
+            module = importlib.import_module(module_path)
+            DifferentiableBackendCls = getattr(module, "DifferentiableBackend")
+            res[engine] = DifferentiableBackendCls
+        except (ImportError, ModuleNotFoundError):
+            pass
```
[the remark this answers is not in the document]
dominikandreasseitz: added it to the logging

qadence/backends/horqrux/backend.py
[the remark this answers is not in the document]
madagra: Happy to be contradicted here but observables transpilation looks really specific to me and should in general be disallowed.

qadence/extensions.py
```diff
@@ -6,13 +6,28 @@
...
-from qadence.types import BackendName, DiffMode
+from qadence.types import BackendName, DiffMode, Engine
 
 backends_namespace = Template("qadence.backends.$name")
 
 logger = get_logger(__name__)
 
 
+def _available_engines() -> dict:
+    """Fallback function for native Qadence available engines if extensions is not present."""
```
madagra, line 17: This docstring should be updated.

examples/backends/low_level/horqrux_backend.py
```diff
@@ -0,0 +1,71 @@
...
+from qadence.types import BackendName, DiffMode
+
+backend = BackendName.HORQRUX
+
+num_epochs = 10
+n_qubits = 4
+depth = 1
+
+fm = feature_map(n_qubits)
+circ = QuantumCircuit(n_qubits, hea(n_qubits, depth=depth))
```
madagra, line 22: Why are you not using the feature map here?

qadence/engines/differentiable_backend.py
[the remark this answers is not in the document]
dominikandreasseitz: clear @nmheim ?

qadence/backends/horqrux/backend.py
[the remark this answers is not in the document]
dominikandreasseitz: not relevant to MR imo, we can track it in an issue @nmheim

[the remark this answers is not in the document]
nmheim: sure, good with me 👍

qadence/engines/differentiable_backend.py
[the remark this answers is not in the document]
nmheim: ahh, right, sorry my mistake.

nmheim (approved):

tests/backends/horqrux/test_quantum_horqrux.py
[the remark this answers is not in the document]
dominikandreasseitz: added

pyproject.toml
```diff
@@ -57,13 +58,29 @@ visualization = [
   # "latex2svg @ git+https://github.com/Moonbase59/latex2svg.git#egg=latex2svg",
   # "scour",
 ]
+horqrux = [
+    "horqrux==0.3.0",
+    "jax",
```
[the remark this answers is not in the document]
dominikandreasseitz: lets leave it for now until we have a new release

qadence/backends/api.py
dominikandreasseitz: les move this discussion somewhere else if there are some remaining questions

qadence/backends/horqrux/backend.py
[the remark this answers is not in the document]
dominikandreasseitz: https://github.com/pasqal-io/qadence/issues/246

qadence/backends/horqrux/backend.py
```diff
@@ -0,0 +1,219 @@
...
+                _probs=probs,
+                n_shots=n_shots,
+                endianness=endianness,
+                n_qubits=circuit.abstract.n_qubits,
+            ),
+        ]
+
+        return samples
+
+    def assign_parameters(self, circuit: ConvertedCircuit, param_values: ParamDictType) -> Any:
```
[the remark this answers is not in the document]
dominikandreasseitz: not possible since mypy will complain that the return types are different in every child class

qadence/backends/horqrux/convert_ops.py
```diff
@@ -0,0 +1,273 @@
...
+    OpName.RY: Ry,
+    OpName.RZ: Rz,
+    OpName.CRX: Rx,
+    OpName.CRY: Ry,
+    OpName.CRZ: Rz,
+    OpName.CNOT: NOT,
+    OpName.I: I,
+}
+
+supported_gates = list(set(list(ops_map.keys())))
```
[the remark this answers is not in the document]
dominikandreasseitz: yes

qadence/types.py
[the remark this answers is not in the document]
dominikandreasseitz: i am happy to hear about alternative aliases. `DifferentiableFunction` is weird

qadence/backends/jax_utils.py
dominikandreasseitz: yes, im sure we need it: ` TypeError: expected np.ndarray (got ArrayImpl)`

qadence/backends/horqrux/convert_ops.py
```diff
@@ -0,0 +1,273 @@
...
+                        control=[b.qubit_support[0] for b in block.blocks],
+                    )
+                ]
+            else:
+                ops = [HorqruxCircuit(ops)]
+
+    elif isinstance(block, CNOT):
+        native_op = ops_map[block.name]
+        ops = [
+            HorqCNOTGate(native_op, block.qubit_support[1], block.qubit_support[0])
```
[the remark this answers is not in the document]
dominikandreasseitz: done

qadence/backends/horqrux/convert_ops.py
```diff
@@ -0,0 +1,273 @@
...
+from itertools import chain as flatten
+from operator import add
+from typing import Any, Callable, Dict
+
+import jax.numpy as jnp
+from horqrux.gates import NOT, H, I, Rx, Ry, Rz, X, Y, Z
+from horqrux.ops import apply_gate
+from horqrux.types import Gate
+from horqrux.utils import overlap
+from jax import Array
```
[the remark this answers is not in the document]
dominikandreasseitz: hm i would agree in a file where both numpy and jax arrays are used, but here its not the case. would resolve

qadence/engines/jax/differentiable_expectation.py
dominikandreasseitz: i have another PSR version which i will merge in a feature MR. for now the way im doing it works since its only single gap. stored as a FIXME

GJBoth (approved):
LGTM, thanks so much for this :-)

PR #117, opened 2023-10-19
Cloud interface implementation in pulser backend

vytautas-a:
Cloud interface implementation in pulser backend.

### DIFF
--- a/qadence/backends/pulser/backend.py
+++ b/qadence/backends/pulser/backend.py
@@ -25,6 +25,7 @@ from qadence.register import Register
 from qadence.utils import Endianness
 
 from .channels import GLOBAL_CHANNEL, LOCAL_CHANNEL
+from .cloud import get_client
 from .config import Configuration
 from .convert_ops import convert_observable
 from .devices import Device, IdealDevice, RealisticDevice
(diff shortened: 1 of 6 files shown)


qadence/backends/pulser/backend.py
```diff
@@ -95,18 +96,37 @@ def make_sequence(circ: QuantumCircuit, config: Configuration) -> Sequence:
...
+            sampling_rate=config.sampling_rate,
+            config=config.sim_config,
+            with_modulation=config.with_modulation,
+        )
+        if state is not None:
+            simulation.set_initial_state(qutip.Qobj(state.cpu().numpy()))
+
+        sim_result = simulation.run(nsteps=config.n_steps_solv, method=config.method_solv)
+        if n_shots is not None:
+            return sim_result.sample_final_state(n_shots)
```
madagra, line 127: Is this returning a `Counter` instance as stated in the return type?

qadence/backends/pulser/cloud.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import annotations
+
+import os
+from functools import lru_cache
+from typing import Optional
+
+from pasqal_cloud import AUTH0_CONFIG, PASQAL_ENDPOINTS, SDK, Auth0Conf, Endpoints, TokenProvider
+
+DEFAULT_CLOUD_ENV = "prod"
```
madagra, line 9: Can you make a `StrEnum` with the available environments instead of using a simple string?

qadence/backends/pulser/config.py
```diff
@@ -49,6 +50,12 @@ class Configuration(BackendConfiguration):
     # interaction type
     interaction: Interaction = Interaction.NN
 
+    # platform to execute sequence on on the cloud
+    cloud_platform: Optional[EmulatorType] = None
```
madagra, line 54: Not sure that `cloud_platform` is the best naming choice here? Maybe `cloud_backend` or `cloud_device`?

qadence/backends/pulser/config.py
```diff
@@ -49,6 +50,12 @@ class Configuration(BackendConfiguration):
...
+
+    # credentials for connecting to the cloud platform
+    cloud_credentials: Optional[dict] = None
```
madagra, line 57: It is likely more elegant to define a dataclass for this and add also the `cloud_platform` attribute directly in the dataclass. In this way, the user needs to specify only one field and it is easier to document.

madagra (commented):
Thanks, @vytautas-a. I have a few comments. Additionally, I would also add a paragraph in the Pulser backend tutorial to explain how to use the cloud backend as well as an entry in the API reference documentation.

qadence/backends/pulser/backend.py
vytautas-a: Yes, it is

qadence/backends/pulser/cloud.py
vytautas-a: Sure

dominikandreasseitz (approved):
i guess i am lacking the context for the need of a cloud.py file here but LGTM

madagra:
> i guess i am lacking the context for the need of a cloud.py file here but LGTM

This is for running Pulser backend simulations on the PASQAL emulators.

PR #129, opened 2023-10-23
[Feat] Adding a verbosity option for the display of training metrics

gvelikova:


qadence/ml_tools/config.py
```diff
@@ -40,6 +40,8 @@ class TrainConfig:
     """A boolean function which evaluates a given training stopping metric is satisfied"""
     batch_size: int = 1
     """The batch_size to use when passing a list/tuple of torch.Tensors."""
+    verbose: bool = True
+    """Whether or not to to be verbose during training."""
```
madagra, line 44: I would rephrase it in a more clear way: "Whether or not to print out metrics values during training".

madagra (approved):
Small comment @gvelikova but overall looks good to me.

PR #130, opened 2023-10-23
Pyqtorch Adjoint

dominikandreasseitz:


pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
     "jsonschema",
     "nevergrad",
     "scipy<1.11",
-    "pyqtorch==0.5.0",
+    "pyqtorch @ git+https://github.com/pasqal-io/pyqtorch.git@ds/remake",
```
madagra, line 42: Reminder to change this one back to the latest PyQ release.

qadence/backends/pyqtorch/backend.py
```diff
@@ -56,7 +56,7 @@ def circuit(self, circuit: QuantumCircuit) -> ConvertedCircuit:
 
         abstract = transpile(*transpilations)(circuit)  # type: ignore[call-overload]
         ops = convert_block(abstract.block, n_qubits=circuit.n_qubits, config=self.config)
-        native = pyq.QuantumCircuit(abstract.n_qubits, ops)
+        native = pyq.QuantumCircuit(abstract.n_qubits, ops, diff_mode="adjoint")
```
madagra, line 59: I guess this should be added to the Backend configuration, right? Do you think it makes sense to have adjoint as the default method?

qadence/backends/pyqtorch/backend.py
```diff
@@ -107,7 +107,7 @@ def run(
             else:
                 batch_size = max([len(tensor) for tensor in param_values.values()])
             state = circuit.native.init_state(batch_size=batch_size)
-        state = circuit.native(state, param_values)
+        state = circuit.native.run(state, param_values)
```
madagra, line 110: Why this change?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -76,24 +76,32 @@ def convert_block(
...
                 block=block,
                 config=config,
             )
+        elif block.n_supports == 1:
+            op = getattr(pyq, block.name)(block.qubit_support[0], config.get_param_name(block)[0])
         else:
-            op = ParametricPyQOperation(n_qubits, block, config)
+            op = getattr(pyq, block.name)(
+                block.qubit_support[0], block.qubit_support[1], config.get_param_name(block)[0]
+            )
```
madagra, line 90: This does not look a generic condition. Maybe you can replace it with: 

```python
op = getattr(pyq, block.name)(*block.qubit_support, config.get_param_name(block)[0])
```

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -76,24 +76,32 @@ def convert_block(
...
     elif isinstance(block, MatrixBlock):
         return [PyQMatrixBlock(block, n_qubits, config)]
     elif isinstance(block, PrimitiveBlock):
-        return [PyQOperation(n_qubits, block)]
+        if len(block.qubit_support) == 1:
+            return [getattr(pyq, block.name)(block.qubit_support[0])]
+        elif isinstance(block, U):
+            return [getattr(pyq, block.name)(block.qubit_support[0])]
+        else:
+            return [getattr(pyq, block.name)(block.qubit_support[0], block.qubit_support[1])]
```
madagra, line 100: As above.

madagra (commented):
A few comments, probably not ready yet this one.

qadence/backends/pyqtorch/backend.py
dominikandreasseitz: removed, was just a test. this MR wont contain the adjoint yet, just the new pyq

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: check now

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: redone

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -104,87 +102,46 @@ def convert_block(
...
-    def forward(self, state: torch.Tensor, values: dict[str, torch.Tensor]) -> torch.Tensor:
-        return self._forward(state, values)
 
 
 class PyQMatrixBlock(Module):
     def __init__(self, block: MatrixBlock, n_qubits: int, config: Configuration = None):
         super().__init__()
         self.n_qubits = n_qubits
-        self.qubits = block.qubit_support
+        self.qubits = list(block.qubit_support)
```
Roland-djee, line 144: This is to conform with `PyQTorch` type. So, has the decision been taken to use `list[int]` for qubit supports there ?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -214,13 +171,13 @@ def forward(
...
-        return self.apply(self.matrices(values, batch_size), state)
+        return self.apply(self.unitary(values, batch_size), state)
 
     def apply(self, matrices: torch.Tensor, state: torch.Tensor) -> torch.Tensor:
         batch_size = state.size(-1)
         return _apply_batch_gate(state, matrices, self.qubits, self.n_qubits, batch_size)
 
-    def matrices(self, values: dict[str, torch.Tensor] | None, batch_size: int) -> torch.Tensor:
+    def unitary(self, values: dict[str, torch.Tensor] | None, batch_size: int) -> torch.Tensor:
         perm = (2, 0, 1)  # We permute the dims since torch.bmm expects the batch_dim at 0.
```
Roland-djee, line 181: Is it possible to construct values tensors such that the batch dim is the first dimension ? Just to avoid these implicit permutations.

tests/backends/pyq/test_quantum_pyq.py
Roland-djee, line 1: Is it worth adding tests to the HamEvo block ?

Roland-djee (approved):
Few comments from me otherwise looking good.

tests/backends/pyq/test_quantum_pyq.py
dominikandreasseitz: definitely!

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: in native pyq , the batch dim is always last, its just to conform with this

dominikandreasseitz: we do have to pass them back to pyq, so there is no way around it

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: its back to tuple

qadence/backends/pyqtorch/backend.py
dominikandreasseitz: i like it because its clear that it returns a wf

tests/backends/pyq/test_quantum_pyq.py
dominikandreasseitz: we do however test if (ofc) somewhere else, i can look for it

jpmoutinho: `constructors.test_daqc.py` depends very heavily on the correctness of `HamEvo`

Roland-djee: From a quick glance at the test, if feels like only a subset of the supported ones are actually checked. Unless I'm mistaken, we should cover all of them including the analog ones.

jpmoutinho: And `test_qft.py` also.

jpmoutinho: Yes it only uses a specific kind of generator, but what I mean is that it tests very concrete cases that ensure the matrix exponentiation is being done correctly. Either way I will for sure revamp the analog tests in the emulated analog MR.

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: exactly

tests/backends/pyq/test_quantum_pyq.py
dominikandreasseitz: ok so lets agree to properly revisit the tests in a separate MR @jpmoutinho

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: its just to comply with the pyq state shape which has the batch dim last, if we do it mixed (for unitary matrices first and states last) we have to always permute and eerything becomes messy in pyq

Issue #116, opened 2023-10-19
Revisit definition and implementation of qubit interaction


madagra:
### Motivations

Currently, the definition of qubit interaction for supporting backends (PyQTorch and Pulser) is not very modular and does not properly account for different underlying implementations (using `HamEvo` with PyQTorch and native with Pulser). 

It is thus desirable to separate the definition of the interaction in a higher-level interface which would include all information needed to define the interaction independently on the way it is implemented in the backend.

### Proposal

Define a general `Interaction` class which will take into account the following properties:

- interaction type
- custom interaction function
- interaction strength (e.g. blockade radius)
- (optional) Rydberg level
- (optional) local addressing pattern to be added on top of the interaction itself
- etc. etc.

We can have a modular way of creating a Hamiltonian. We might need a `Hamiltonian` object for this purpose.

This would look something like the following:

```python
@dataclass
class Interaction:

	type: InteractionType = InteractionType.NN
	custom_interaction_fn: Callable | None = None
	interaction_coeff: float = None
	rydberg_level: int = 70
	semi_local_addr_patterns: list[AddressingPatterns] = None
	blockade_radius: float = None
	
	# generator for the interaction given the above parameters
  # it spits out a block where all the sub-blocks have been
  # replaced with HamEvo operations with the right generator
	
	# this can replace the current `add_interaction` method
	def add_to_block(self, block: AbstractBlock | Register | QuantumCircuit) -> AbstractBlock:
	  pass	
```

The interaction can then be added either as a `Backend` argument or as an argument in the `QuantumCircuit` constructor. The first is likely preferable since the interaction is associated with the execution backend independently from the given circuit.

```python
class Backend:
	... other attributes
	interaction: Interaction | None = None

       # validate that the given interaction is executable with the
       # chosen backend and fail otherwise. Not every backend can
       # correctly simulate all interactions
       def validate_interaction(self):
           raise NotImplementedError
```

nmheim:
we should also take care of register information here somehow. `AnalogBlock` eigenvalues in the pulser backend are a bit of a mess currently, because they don't have access to qubit positions.

vytautas-a:
I guess we could have an additional argument `register` in `Interaction` class.

vytautas-a:
[...]

PR #131, opened 2023-10-24
[Refactoring] Calculation of analog block eigenvalues

vytautas-a:
Proposal for refactoring eigenvalue calculation of analog blocks for further use with Pulser/GPSR combination.

Related to issue #116

### DIFF
--- a/qadence/backends/pulser/backend.py
+++ b/qadence/backends/pulser/backend.py
@@ -14,6 +14,7 @@ from pulser_simulation.simresults import SimulationResults
 from pulser_simulation.simulation import QutipEmulator
 from torch import Tensor
 
+from qadence import add_interaction
 from qadence.backend import Backend as BackendInterface
 from qadence.backend import BackendName, ConvertedCircuit, ConvertedObservable
 from qadence.backends.utils import to_list_of_dicts
@@ -86,7 +87,7 @@ def make_sequence(circ: QuantumCircuit, config: Configuration) -> Sequence:
     )
     sequence.add(zero_pulse, GLOBAL_CHANNEL, "wait-for-all")
 
-    add_pulses(sequence, circ.block, config, circ.register, spacing)
+    add_pulses(sequence, circ.block, config)
     sequence.measure()
 
     return sequence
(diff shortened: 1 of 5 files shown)


qadence/transpile/emulate.py
```diff
@@ -275,7 +275,13 @@ def _(block: ConstantAnalogRotation, register: Register, interaction: Callable)
     # convert "global" to indexed qubit suppport so that we can re-use `kron` dispatched function
     b = deepcopy(block)
     b.qubit_support = QubitSupport(*range(register.n_qubits))
-    return _add_interaction(kron(b), register, interaction)
+
+    # convert to HamEvo block
+    hamevo_block = _add_interaction(kron(b), register, interaction)
+
+    # set eigenvalues of the ConstantAnalogRotation block
+    block.eigenvalues_generator = hamevo_block.eigenvalues_generator
```
madagra, line 283: This is a bit risky since you are modifying the block in place. It might give some nasty side effects? Should we maybe rethink a little bit the `add_interaction` function and its behavior? I guess this is something that @jpmoutinho will take care of in his refactoring anyway.

madagra (commented):
Ehi @vytautas-a, I see the point of these changes but they feel a bit risky to me. I would try to address this issue in the more general refactoring planned (as you suggest).

qadence/transpile/emulate.py
vytautas-a: You see, the problem here is that we need to have eigenvalues for the `ConstantAnalogRotation` block itself so this in-place modification is unavoidable with the current block implementation. This calculation of eigenvalues may be done in some different place but in the end you still have to assign them to the analog block itself.

qadence/backends/pulser/backend.py
```diff
@@ -123,6 +124,9 @@ class Backend(BackendInterface):
     config: Configuration = Configuration()
 
     def circuit(self, circ: QuantumCircuit) -> Sequence:
+        # set eigenvalues of any ConstantAnalogRotation blocks in circuit
+        add_interaction(circ, spacing=self.config.spacing)
```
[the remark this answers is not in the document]
vytautas-a: No, I don't need it. I simply need a function that implements the neutral-atom Hamiltonian with all the correct parameters then calls `block_to_tensor` and finally calculates the corresponding eigenvalues.

Roland-djee:
@vytautas-a @jpmoutinho is this still relevant ?

PR #135, opened 2023-10-25
[Refactor] Train loop; make sure qadence runs on GPUs correctly

dominikandreasseitz:
refactor the `train` functions to have consistent behaviour across dataloader/dictdataloaders.
Now, the train loop will simply iterate over the provided dataloader. if that dataloader is constructed with "infinite" datasets we get back the behaviour from the previous dictdataloader case. See the script above for a full example. the important bit is:

```python
x = torch.linspace(0, 1, batch_size).reshape(-1, 1)
y = torch.sin(x)

# with infinite=True, the dataloader will spit out new batches indefinitely
# note that here the batch_size is equal to the whole dataset size
data = to_dataloader(x, y, batch_size=x.size(0), infinite=True)

# with infinite=False, the dataloader will iterate once over the dataset
# leading to 10 iterations
data = to_dataloader(x, y, batch_size=x.size(0)/10, infinite=True)

train(model, data, ...)
```

**Note:** you now HAVE TO supply a dataloader/something that can be iterated over to get batches, in order to use the train loop. tensors/lists are supported via `to_dataloader`.

Further, we now get consistent GPU behaviour via 
https://github.com/pasqal-io/qadence/pull/135/files#diff-24f77cd4c0f3818366ac3978a215833d6082be2518721615d3604a068b8c9fbcR145

### DIFF
--- a/qadence/backends/pyqtorch/backend.py
+++ b/qadence/backends/pyqtorch/backend.py
@@ -30,9 +30,6 @@ from qadence.utils import Endianness, int_to_basis
 from .config import Configuration
 from .convert_ops import convert_block, convert_observable
 
-DEVICE = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
-torch.set_default_device(DEVICE)
-
 
 @dataclass(frozen=True, eq=True)
 class Backend(BackendInterface):
@@ -60,9 +57,7 @@ class Backend(BackendInterface):
         abstract = transpile(*transpilations)(circuit)  # type: ignore[call-overload]
         ops = convert_block(abstract.block, n_qubits=circuit.n_qubits, config=self.config)
         native = pyq.QuantumCircuit(abstract.n_qubits, ops)
-        return ConvertedCircuit(
-            native=native.to(device=DEVICE), abstract=abstract, original=circuit
-        )
+        return ConvertedCircuit(native=native, abstract=abstract, original=circuit)
 
     def observable(self, observable: AbstractBlock, n_qubits: int) -> ConvertedObservable:
         # make sure only leaves, i.e. primitive blocks are scaled
@@ -75,9 +70,7 @@ class Backend(BackendInterface):
         block = transpile(*transpilations)(observable)  # type: ignore[call-overload]
 
         (native,) = convert_observable(block, n_qubits=n_qubits, config=self.config)
-        return ConvertedObservable(
-            native=native.to(device=DEVICE), abstract=block, original=observable
-        )
+        return ConvertedObservable(native=native, abstract=block, original=observable)
 
     def run(
         self,
--- a/qadence/backends/pyqtorch/backend.py
+++ b/qadence/backends/pyqtorch/backend.py
@@ -18,8 +18,6 @@ from qadence.measurements import Measurements
 from qadence.overlap import overlap_exact
 from qadence.states import zero_state
 from qadence.transpile import (
-    add_interaction,
-    blockfn_to_circfn,
     chain_single_qubit_ops,
     flatten,
     scale_primitive_blocks_only,
@@ -27,7 +25,7 @@ from qadence.transpile import (
 )
 from qadence.utils import Endianness, int_to_basis
 
-from .config import Configuration
+from .config import Configuration, default_passes
 from .convert_ops import convert_block, convert_observable
 
 
@@ -46,18 +44,17 @@ class Backend(BackendInterface):
     config: Configuration = Configuration()
 
     def circuit(self, circuit: QuantumCircuit) -> ConvertedCircuit:
-        transpilations = [
-            lambda circ: add_interaction(circ, interaction=self.config.interaction),
-            lambda circ: blockfn_to_circfn(chain_single_qubit_ops)(circ)
-            if self.config.use_single_qubit_composition
-            else blockfn_to_circfn(flatten)(circ),
-            blockfn_to_circfn(scale_primitive_blocks_only),
-        ]
+        passes = self.config.transpilation_passes
+        if passes is None:
+            passes = default_passes(self.config)
 
-        abstract = transpile(*transpilations)(circuit)  # type: ignore[call-overload]
-        ops = convert_block(abstract.block, n_qubits=circuit.n_qubits, config=self.config)
-        native = pyq.QuantumCircuit(abstract.n_qubits, ops)
-        return ConvertedCircuit(native=native, abstract=abstract, original=circuit)
+        original_circ = circuit
+        if len(passes) > 0:
+            circuit = transpile(*passes)(circuit)
+
+        ops = convert_block(circuit.block, n_qubits=circuit.n_qubits, config=self.config)
+        native = pyq.QuantumCircuit(circuit.n_qubits, ops)
+        return ConvertedCircuit(native=native, abstract=circuit, original=original_circ)
 
     def observable(self, observable: AbstractBlock, n_qubits: int) -> ConvertedObservable:
         # make sure only leaves, i.e. primitive blocks are scaled
--- a/qadence/backends/pyqtorch/backend.py
+++ b/qadence/backends/pyqtorch/backend.py
@@ -30,6 +30,9 @@ from qadence.utils import Endianness, int_to_basis
 from .config import Configuration
 from .convert_ops import convert_block, convert_observable
 
+DEVICE = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+torch.set_default_device(DEVICE)
+
 
 @dataclass(frozen=True, eq=True)
 class Backend(BackendInterface):
@@ -57,7 +60,9 @@ class Backend(BackendInterface):
         abstract = transpile(*transpilations)(circuit)  # type: ignore[call-overload]
         ops = convert_block(abstract.block, n_qubits=circuit.n_qubits, config=self.config)
         native = pyq.QuantumCircuit(abstract.n_qubits, ops)
-        return ConvertedCircuit(native=native, abstract=abstract, original=circuit)
+        return ConvertedCircuit(
+            native=native.to(device=DEVICE), abstract=abstract, original=circuit
+        )
 
     def observable(self, observable: AbstractBlock, n_qubits: int) -> ConvertedObservable:
         # make sure only leaves, i.e. primitive blocks are scaled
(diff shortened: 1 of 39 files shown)


qadence/backends/pyqtorch/backend.py
```diff
@@ -30,6 +30,9 @@
 from .config import Configuration
 from .convert_ops import convert_block, convert_observable
 
+DEVICE = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+torch.set_default_device(DEVICE)
```
madagra, line 34: I thought this would move all the Torch objects to the device without user intervention.

madagra (commented):
I am in favor of the release but not in favor of using the `.to()` everywhere. Honestly, I much rather leave the user taking care of where he wants to put his model rather than doing it at the `qadence` level.

qadence/backends/pyqtorch/backend.py
awennersteen: I'm not sure this is a good idea.
Not to mention if you're on the cluster you might get unexpected results since there will be multiple gpus, so this could make multi-gpu training difficult.

this looks to me to be something the user might write, but not us.

qadence/ml_tools/train_grad.py
```diff
@@ -3,25 +3,29 @@
...
+set_default_device(DEVICE)
+
 
 def train(
     model: Module,
-    dataloader: DictDataLoader | DataLoader | list[Tensor] | tuple[Tensor, Tensor] | None,
+    dataloader: DataLoaderType,
     optimizer: Optimizer,
     config: TrainConfig,
     loss_fn: Callable,
```
madagra, line 31: In the `ml_tools` it is fine to actually keep the calls to `.to()`, I think. This is a high-routine that has complete control on everything (data, models, optimizer, etc. etc.) so we can be sure that everything is indeed sent to the right device.

qadence/ml_tools/train_grad.py
```diff
@@ -3,25 +3,29 @@
...
-from qadence.ml_tools.data import DictDataLoader
+from qadence.ml_tools.data import DataLoaderType, DictDataLoader, dataloader_to_device
 from qadence.ml_tools.optimize_step import optimize_step
 from qadence.ml_tools.printing import print_metrics, write_tensorboard
 from qadence.ml_tools.saveload import load_checkpoint, write_checkpoint
 
 logger = get_logger(__name__)
 
+DEVICE = torchdevice("cuda") if cuda.is_available() else torchdevice("cpu")
+set_default_device(DEVICE)
```
madagra, line 23: This is not needed here, I think. We pass the `device` as argument to the `train()` function.

dominikandreasseitz: removed

qadence/ml_tools/train_grad.py
dominikandreasseitz: i did that now

qadence/backends/pyqtorch/backend.py
dominikandreasseitz: agreed and removed

dominikandreasseitz:
@awennersteen @madagra , the MR changed now to ONLY move the data/dataloader and the model to the device requested by the user in train_grad.

qadence/blocks/embedding.py
```diff
@@ -109,6 +109,7 @@ def embedding_fn(params: StrTensorDict, inputs: StrTensorDict) -> StrTensorDict:
         for e in constant_expressions + unique_const_matrices:
             embedded_params[e] = params[stringify(e)]
 
+        embedded_params = {k: v for k, v in embedded_params.items()}
```
Roland-djee, line 112: Not sure this is necessary.

Roland-djee (approved):
LGTM. Minor comment.

qadence/ml_tools/data.py
```diff
@@ -23,10 +26,45 @@ def __iter__(self) -> DictDataLoader:
...
+            for key, dl in self.dataloaders.items()
+        }
+        return self
+
+
+DataLoaderType = Union[DictDataLoader, DataLoader, list[Tensor], tuple[Tensor, Tensor], None]
+
 
-def to_dataloader(x: torch.Tensor, y: torch.Tensor, batch_size: int = 1) -> DataLoader:
+def tensor_to_dataloader(x: Tensor, y: Tensor, batch_size: int = 1) -> DataLoader:
```
nmheim, line 43: this would be a breaking change... I don't think that's necessary

qadence/ml_tools/data.py
```diff
@@ -23,10 +26,45 @@ def __iter__(self) -> DictDataLoader:
...
     """Convert two torch tensors x and y to a Dataloader."""
     return DataLoader(TensorDataset(x, y), batch_size=batch_size)
+
+
+@singledispatch
+def data_to_device(xs: Any, device: str = "cpu") -> Any:
+    return xs
```
nmheim, line 50: should we turn this into a fallback for `torch.Tensor` and use `.to`?

nmheim:
This currently works:
```python

from itertools import count
from pathlib import Path
from typing import Any

import numpy as np
import pytest
import torch
from torch import nn
from torch.utils.data import DataLoader, TensorDataset

from qadence.ml_tools import DictDataLoader, TrainConfig, train_with_grad
from qadence.models import QNN

# import matplotlib.pyplot as plt

class BasicNetwork(nn.Module):
    def __init__(self, n_neurons: int = 5) -> None:
        super().__init__()
        network = [
            nn.Linear(1, n_neurons),
            nn.ReLU(),
            nn.Linear(n_neurons, n_neurons),
            nn.ReLU(),
            nn.Linear(n_neurons, 1),
        ]
        self.network = nn.Sequential(*network)
        self.n_neurons = n_neurons

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        return self.network(x)

batch_size = 25
n_iters = 1000
x = torch.linspace(0, 1, batch_size*n_iters).reshape(-1, 1)
y = torch.sin(x)

dataset = TensorDataset(x, y)
data = DataLoader(dataset, batch_size=batch_size, shuffle=True)

model = BasicNetwork()

cnt = count()
criterion = torch.nn.MSELoss()
optimizer = torch.optim.Adam(model.parameters(), lr=0.1)

def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, dict]:
    next(cnt)
    x, y = data
    out = model(x)
    loss = criterion(out, y)
    return loss, {}


config = TrainConfig(folder=None, max_iter=n_iters, checkpoint_every=100, write_every=100,
                     print_every=10)

train_with_grad(model, data, optimizer, config, loss_fn=loss_fn, device="cuda")
assert next(cnt) == n_iters

x = torch.rand(5, 1)
assert torch.allclose(torch.sin(x), model.to("cpu")(x), rtol=1e-1, atol=1e-1)

# p = model(x)
# plt.plot(x.squeeze(), torch.sin(x.squeeze()))
# plt.plot(x.squeeze(), model(x).squeeze().detach())
# plt.show()
# 
```

nmheim:
I have refactored the `train_with_grad` function in here to have consistent behaviour across dataloader/dictdataloaders.
Now, the train loop will simply iterate over the provided dataloader. if that dataloader is constructed with "infinite" datasets we get back the behaviour from the previous dictdataloader case. See the script above for a full example. the important bit is:

```python
x = torch.linspace(0, 1, batch_size).reshape(-1, 1)
y = torch.sin(x)

# with infinite=True, the dataloader will spit out new batches indefinitely
# note that here the batch_size is equal to the whole dataset size
data = to_dataloader(x, y, batch_size=x.size(0), infinite=True)

# with infinite=False, the dataloader will iterate once over the dataset
# leading to 10 iterations
data = to_dataloader(x, y, batch_size=x.size(0)/10, infinite=True)
```

Further, we now get consistent GPU behaviour via 
https://github.com/pasqal-io/qadence/pull/135/files#diff-24f77cd4c0f3818366ac3978a215833d6082be2518721615d3604a068b8c9fbcR145

qadence/blocks/embedding.py
dominikandreasseitz: i just forgot to remove this

madagra:
> I have refactored the `train_with_grad` function in here to have consistent behaviour across dataloader/dictdataloaders. Now, the train loop will simply iterate over the provided dataloader. if that dataloader is constructed with "infinite" datasets we get back the behaviour from the previous dictdataloader case. See the script above for a full example. the important bit is:
> 
> ```python
> x = torch.linspace(0, 1, batch_size).reshape(-1, 1)
> y = torch.sin(x)
> 
> # with infinite=True, the dataloader will spit out new batches indefinitely
> # note that here the batch_size is equal to the whole dataset size
> data = to_dataloader(x, y, batch_size=x.size(0), infinite=True)
> 
> # with infinite=False, the dataloader will iterate once over the dataset
> # leading to 10 iterations
> data = to_dataloader(x, y, batch_size=x.size(0)/10, infinite=True)
> 
> train(model, data, ...)
> ```
> 
> **Note:** you now HAVE TO supply a dataloader/something that can be iterated over to get batches, in order to use the train loop. tensors/lists are supported via `to_dataloader`.
> 
> Further, we now get consistent GPU behaviour via https://github.com/pasqal-io/qadence/pull/135/files#diff-24f77cd4c0f3818366ac3978a215833d6082be2518721615d3604a068b8c9fbcR145

Looks nice, @nmheim. What about training a model **without** data, though? This is very common in, e.g. chemistry calculations.

nmheim:
> Looks nice, @nmheim. What about training a model **without** data, though? This is very common in, e.g. chemistry calculations.

without data is still working like before by passing None for the dataloader:)

madagra:
> > Looks nice, @nmheim. What about training a model **without** data, though? This is very common in, e.g. chemistry calculations.
> 
> without data is still working like before by passing None for the dataloader:)

Great, then it looks good to me!

madagra (approved):
LGTM.

qadence/ml_tools/data.py
```diff
@@ -12,21 +15,65 @@ class DictDataLoader:
...
+
+    def __iter__(self) -> Iterator:
+        if len(set([t.size(0) for t in self.tensors])) != 1:
+            raise ValueError("Size of first dimension must be the same for all tensors.")
+
+        for idx in cycle(range(self.tensors[0].size(0))):
+            yield tuple(t[idx] for t in self.tensors)
+
+
+def to_dataloader(x: Tensor, y: Tensor, batch_size: int = 1, infinite: bool = False) -> DataLoader:
```
madagra, line 56: Why don't we allow here for arbitrary number of tensors?

```python
to_dataloader(*tensors: torch.Tensor, batch_size = 1, infinite = False):
    pass
```

PR #147, opened 2023-10-27
[Bug fix] Fix pulser initial state passing

jpmoutinho:


tests/backends/test_endianness.py
```diff
@@ -160,19 +161,25 @@ def test_backend_sample_endianness(
...
+        RX(1, torch.pi),
+        RX(2, torch.pi),
     ],
 )
 def test_pulser_run_endianness(
-    circ: QuantumCircuit,
+    init_state: Tensor,
+    block: AbstractBlock,
 ) -> None:
+    register = Register.from_coordinates([(0.0, 10.0), (0.0, 20.0), (0.0, 30.0)])
```
jpmoutinho, line 178: Just FYI, I changed this to a register of "far-away" qubits just so I could "ignore" the atomic interaction in the pulser backend. Otherwise, these tests would only pass for initial state 000, which is what was being done previously. That makes sense because in that scenario the atoms don't interact anyway.

Proper tests comparing pyqtorch with pulser backend in an interacting scenario will come with the work on emulated analog.

qadence/backends/pulser/backend.py
```diff
@@ -35,6 +35,16 @@
 DEFAULT_SPACING = 8.0  # µm (standard value)
 
 
+def _convert_init_state(state: Tensor | None) -> np.ndarray | None:
+    """Flips and squeezes initial state, to be consistent with Pulser."""
```
Roland-djee, line 39:
```suggestion
    """Flip and squeeze initial state consistent with Pulser convention."""
```

qadence/backends/pulser/backend.py
```diff
@@ -35,6 +35,16 @@
...
+    if state is not None:
```
Roland-djee, line 40: I wouldn't do this check here, rather in `run` or `sample` directly.

qadence/backends/pulser/backend.py
```diff
@@ -164,7 +174,7 @@ def run(
         endianness: Endianness = Endianness.BIG,
     ) -> Tensor:
         vals = to_list_of_dicts(param_values)
-
+        state = _convert_init_state(state)
```
Roland-djee, line 177: Check if `state` is not `None` here.

qadence/backends/pulser/backend.py
```diff
@@ -201,6 +211,7 @@ def sample(
             raise ValueError("You can only call sample with n_shots>0.")
 
         vals = to_list_of_dicts(param_values)
+        state = _convert_init_state(state)
```
Roland-djee, line 214: Check `None`.

tests/backends/test_endianness.py
```diff
@@ -160,19 +161,25 @@ def test_backend_sample_endianness(
...
+    circ = QuantumCircuit(register, block)
     for endianness in Endianness:
-        wf_pyq = run(circ, {}, backend="pyqtorch", endianness=endianness)
-        wf_pulser = run(circ, {}, backend="pulser", endianness=endianness)
+        wf_pyq = run(circ, {}, backend="pyqtorch", endianness=endianness, state=init_state)
```
Roland-djee, line 181: Never use `str` or people will hate you ;)

Roland-djee (approved):
Hey @jpmoutinho thanks for this. Few comments otherwise LGTM.

PR #157, opened 2023-11-02
[Feat] Identity-initialized QNN

n-toscano:
In the context of barren plateau mitigation techniques, this model allows to create and initiate the weights of a `QNN` in a similar fashion to what proposed by [Grant et al.](https://arxiv.org/abs/1903.05076).

- [ ] documentation
- [ ] tests
- [ ] tutorial

### DIFF
--- a/qadence/models/identity_qnn.py
+++ b/qadence/models/identity_qnn.py
@@ -1,0 +1,158 @@
+from __future__ import annotations
+
+import itertools
+from typing import Callable
+
+import numpy as np
+from torch import Tensor
+
+from qadence.backend import BackendConfiguration, BackendName
+from qadence.backends.pytorch_wrapper import DiffMode
+from qadence.blocks import AbstractBlock, chain, kron, tag
+from qadence.blocks.utils import parameters
+from qadence.circuit import QuantumCircuit
+from qadence.measurements import Measurements
+from qadence.models.quantum_model import QuantumModel
+from qadence.operations import CNOT, RX, RY
+
+
+def identity_block(
+    n_qubits: int,
+    layer: int = 0,
+    ops: list[type[AbstractBlock]] = [RX, RY],
+    periodic: bool = False,
+) -> AbstractBlock:
+    """
+    Identity block for barren plateau mitigation.
+    The initial configuration of this block is equal to an identity unitary
+    but can be trained in the same fashion as other ansatzes reaching same level
+    of expressivity.
+    """
+
+    left_rotations = [
+        kron(
+            gate(n, "alpha" + f"_{layer}{n + n_qubits*i}")  # type: ignore [arg-type]
+            for n in range(n_qubits)
+        )
+        for i, gate in enumerate(ops)
+    ]
+
+    if not periodic:
+        left_cnots = [chain(CNOT(n, n + 1) for n in range(n_qubits - 1))]
+    else:
+        left_cnots = [chain(CNOT(n, (n + 1) % n_qubits) for n in range(n_qubits))]
+
+    centre_rotations = [kron(RX(n, "theta" + f"_{layer}{n}") for n in range(n_qubits))]
+
+    right_cnots = reversed(*left_cnots)
+
+    right_rotations = [
+        kron(
+            gate(
+                n,  # type: ignore [arg-type]
+                "beta" + f"_{layer}{n + n_qubits*(len(ops)-i-1)}",  # type: ignore [arg-type]
+            )
+            for n in range(n_qubits)
+        )
+        for i, gate in enumerate(reversed(ops))
+    ]
+
+    krons = [
+        *left_rotations,
+        *left_cnots,
+        *centre_rotations,
+        *right_cnots,
+        *right_rotations,
+    ]
+
+    return tag(chain(*krons), tag=f"BPMA-{layer}")
+
+
+class IdentityQNN(QuantumModel):
+    def __init__(
+        self,
+        feature_map: AbstractBlock,
+        depth: int = 1,
+        observable: list[AbstractBlock] | AbstractBlock | None = None,
+        transform: Callable[[Tensor], Tensor] = None,
+        backend: BackendName = BackendName.PYQTORCH,
+        diff_mode: DiffMode = DiffMode.AD,
+        protocol: Measurements | None = None,
+        configuration: BackendConfiguration | dict | None = None,
+    ):
+        self.circ = self._create_circuit(feature_map, depth)
+
+        super().__init__(
+            circuit=self.circ,
+            observable=observable,
+            backend=backend,
+            diff_mode=diff_mode,
+            protocol=protocol,
+            configuration=configuration,
+        )
+
+        self._initialize_circuit(depth)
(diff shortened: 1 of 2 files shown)


qadence/models/identity_qnn.py
```diff
@@ -0,0 +1,158 @@
...
+from qadence.backends.pytorch_wrapper import DiffMode
+from qadence.blocks import AbstractBlock, chain, kron, tag
+from qadence.blocks.utils import parameters
+from qadence.circuit import QuantumCircuit
+from qadence.measurements import Measurements
+from qadence.models.quantum_model import QuantumModel
+from qadence.operations import CNOT, RX, RY
+
+
+def identity_block(
```
gvelikova, line 19: I'd suggest moving this to qadence.blocks under the name identity.py

nmheim (changes requested):
Thanks a lot for this! Very nice to have I think. I would suggest a little refactoring though. It is less error prone to construct blocks with specific parameters via
```python
# variational parameter
RX(0, Parameter("theta", value=some_value))

# feature parameter
RX(0, Parameter("theta", value=some_value, trainable=False))
```
instead of using `set_parameter`. what do you think?

n-toscano:
> Thanks a lot for this! Very nice to have I think. I would suggest a little refactoring though. It is less error prone to construct blocks with specific parameters via
> 
> ```python
> # variational parameter
> RX(0, Parameter("theta", value=some_value))
> 
> # feature parameter
> RX(0, Parameter("theta", value=some_value, trainable=False))
> ```
> 
> instead of using `set_parameter`. what do you think?

Thanks Niklas. Just to make sure, do you mean that all the parameter initialization can be done within the circuit and before injecting it into the `QuantumModel`?
If so, do we still need an `IdentityQNN` or a specific layer/circuit can do the work?

n-toscano:
Regarding documentation/examples, I was thinking about adding an `identity_qnn.py` script in the `/examples/models` folder, as well as a section in the `custom_models.md` file in `/docs/advanced_tutorials`. Do you agree?

qadence/models/identity_qnn.py
jpmoutinho: Actually `qadence.blocks` is mostly about the infrastructure of the block system :). I would suggest `qadence.constructors` for functions creating specific block compositions that can be generally useful for different algorithms. There is already an `ansatze.py` file there, but it's pretty dedicated to the hardware-efficient ansatz (and thus might need a rename), so this one might be better off in a separate file.

gvelikova: Thanks @jpmoutinho, I guess it might make sense to be in qadence.constructors then.

nmheim:
> Thanks @nmheim. Just to make sure, do you mean that all the parameter initialization can be done within the circuit and before injecting it into the `QuantumModel`? If so, do we still need an `IdentityQNN` or a specific layer/circuit can do the work?

yes, exactly, the `identity_block` function you have now should do the trick. I can see you did it already:) another nitpick of mine: could we rename the `identity_block` function to something that reflects that its an ansatz which (only initially) is an identity? otherwise it could be a bit confusing, because we already have an identity block (`I`).

n-toscano:
> > Thanks @nmheim. Just to make sure, do you mean that all the parameter initialization can be done within the circuit and before injecting it into the `QuantumModel`? If so, do we still need an `IdentityQNN` or a specific layer/circuit can do the work?
> 
> yes, exactly, the `identity_block` function you have now should do the trick. I can see you did it already:) another nitpick of mine: could we rename the `identity_block` function to something that reflects that its an ansatz which (only initially) is an identity? otherwise it could be a bit confusing, because we already have an identity block (`I`). Thanks a lot for all the work here!

Sure, thanks @nmheim! Do you think `identity_initialized_block` would do it? Maybe it's a little too verbose. The tags that I'm using are `BPMA` as for barren-plateau-mitigation-ansatz, so maybe `bpm_`_something_ can be an alternative?

qadence/models/identity_qnn.py
n-toscano: Thanks @jpmoutinho, will move it there!

gvelikova:
@n-toscano wouldn't identity_initialization work too? Having _block or _ansatz in the name would be more precise but as you say, verbose.

nmheim:
> @n-toscano wouldn't identity_initialization work too? Having _block or _ansatz in the name would be more precise but as you say, verbose.

I kind of like `identity_initialized_ansatz`. verbose, but clear :P

jpmoutinho:
> One more thing: I know that qadence's `hea` has multiple options for the entangling technique, to make it more _hardware-efficient_. Is this something that we should try to include in this PR or makes sense to expand this in the future?

I think whatever interface decisions you can make based on making it similar to the `hea` function, the better, but for sure no need to implement all the analog options in this MR.

qadence/constructors/id_init_ansatz.py
```diff
@@ -0,0 +1,120 @@
...
+            2-qubit entangling operation. Supports CNOT, CZ, CRX, CRY, CRZ, CPHASE.
+            Controlled rotations will have variational parameters on the rotation angles.
+        periodic (bool): if the qubits should be linked periodically.
+    """
+    initialized_layers = []
+    for layer in range(depth):
+        alpha = 2 * torch.pi * torch.rand(n_qubits * len(rotations))
+        gamma = torch.zeros(n_qubits)
+        beta = -alpha
+        left_rotations = [
```
jpmoutinho, line 58: My only nitpick is if you could somehow write the code in this function a bit more structured and in shorter lines, with less chaining of operations. It's very annoying I know 😅 but because we have a limit on line length, the pre-commit hooks will format it like it is now, which makes it harder to read.

jpmoutinho: So e.g. instead of `kron(stuff for i in range(n_qubits))` putting the `for` in a separate line above. You can use list appends for example and then at the end do `kron(*list)`.

n-toscano: Sounds good, @jpmoutinho. I reorganized the way left and right rotations are created, let me know if this version accomplishes what you pointed out.

jpmoutinho: Thanks! It looks better :)

jpmoutinho (approved):

nmheim (changes requested):
Thanks again for this! The code looks good to me. Before merging, can we add a test that verifies that the ansatz is doing what it is supposed to, i.e. something like:
```python
state = rand_state(2)
assert torch.allclose(state, run(identity_initialized_ansatz(2,2), state=state))
```

n-toscano:
> Thanks again for this! The code looks good to me. Before merging, can we add a test that verifies that the ansatz is doing what it is supposed to, i.e. something like:
> 
> ```python
> state = rand_state(2)
> assert torch.allclose(state, run(identity_initialized_ansatz(2,2), state=state))
> ```

Makes sense! Solved with the latest commit

gvelikova (approved):

nmheim (approved):
LGTM! 🎉

PR #158, opened 2023-11-03
[Feature] Added Rydberg HEA

madagra:
Constructor for a hardware-efficient ansatz compatible with Rydberg atom arrays with semi-local addressability.


docs/digital_analog_qc/daqc-basics.md
```diff
@@ -6,18 +6,27 @@ paradigm[^1], based on two primary computations:
...
+Furthermore, the essence of digital-analog computation is the ability to represent any
+analog operation, _i.e._ any arbitrary Hamiltonian, using an
+auxiliary device-amenable Hamiltonian, such as the ubiquitous Ising model[^2].
+This is at the core of the DAQC implementation in Qadence.
+
+## Execution on Rydberg atom arrays with restriced addressability
+
+Finally, Qadence offers some convenience constructors and interfaces to execute
+programs compatible with neutral atom quantum computers with only restricted access
+to individual qubit addressability which is instead required by DAQC.
```
vincentelfving, line 29: this part may be read a bit ambiguously. In my opinion, we should still call restricted-addressability + always-on interactions, as a DAQC protocol. e.g. DAQC would not necessarily mean full-addressing. It is delicate, because any QC paradigm should ideally be defined in a way to have proven universality. We haven't proven it for restricted-addressing, but my suspicion is that it can be done.

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
+Qadence offers simplifies the execution of digital-analog
```
vincentelfving, line 1: offers or simplifies

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
...
+    \mathcal{H}_{\textrm{global}} = \frac{\Omega}{2}\sum_{i}^N \left(
+        \textrm{cos}(\phi)\sigma^x_i - \textrm{sin}(\phi)\sigma^y_i \right) -
+        \delta \sum_{i}^N \hat{n}_i \\
+    \mathcal{H}_{\textrm{int}} = \sum_{i<j} \frac{C_6}{|R_i - R_j|^6} \hat{n}_i \hat{n}_j
+$$
+where $\Omega$ is the Rabi frequency, $\phi$ the global phase, $\delta$ the
+detuning which can all be time-dependent (here omitted for simplicity). The operator
+$\hat{n}_i = \frac{1+\sigma_i^z}{2}$ is the occupation operator for the Rydberg state.
+
+The local addressability term read instead as follows:
```
vincentelfving, line 26: term reads, or terms read

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
...
+and read as follows:
+
+$$
+    \mathcal{H}_{\textrm{global}} = \frac{\Omega}{2}\sum_{i}^N \left(
+        \textrm{cos}(\phi)\sigma^x_i - \textrm{sin}(\phi)\sigma^y_i \right) -
+        \delta \sum_{i}^N \hat{n}_i \\
+    \mathcal{H}_{\textrm{int}} = \sum_{i<j} \frac{C_6}{|R_i - R_j|^6} \hat{n}_i \hat{n}_j
+$$
+where $\Omega$ is the Rabi frequency, $\phi$ the global phase, $\delta$ the
+detuning which can all be time-dependent (here omitted for simplicity). The operator
```
vincentelfving, line 23: maybe it's nicer to re-introduce the time-dependence? is it too complicated?

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
...
+$\hat{n}_i = \frac{1+\sigma_i^z}{2}$ is the occupation operator for the Rydberg state.
```
vincentelfving, line 24: also mention what is R_i

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
...
+
+Qadence implements the Hamiltonian above in two different flavors of increasing complexity described
+below.
+
+## Circuit constructor
+
+The [`rydberg_hea`][qadence.constructors.rydberg_hea.rydberg_hea] allows to
+build a circuit instance implementing a basic version of the Hamiltonian
+evolution described above where both $\Delta$ and $\tilde{\Omega}$ coefficients
+are considered constants and **no global drive and detuning** are allowed. Furthermore,
```
vincentelfving, line 48: this may be confusing to the reader; we could mention that, although the term is not included explicitly, as we have no limittions on the weigths normalization, effectively the same exact hamiltonian can be obtained, e.g. the global drive or detuning could be obtained by increasing all individual components by an equal amount.

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
...
+
+```python exec="on" source="material-block" result="json" session="rydb-hea"
+import qadence as qd
+from qadence import rydberg_hea
+
+n_qubits = 4
+n_layers = 2
+
+# determine the interaction
+register = qd.Register.line(n_qubits)
```
vincentelfving, line 69: afaik, this line does not directly determine the interaction, but rather the atom layout. The interaction cutoffs (NN, NNN etc) can be chosen independently from this

docs/digital_analog_qc/rydberg-hea.md
```diff
@@ -0,0 +1,106 @@
...
+    register,
+    tevo_det,
+    tevo_drive,
+    tevo_int,
+    detunings=weights_det,
+    drives=weights_drive,
+)
+```
+
+## Usage with digital-analog emulation
```
vincentelfving, line 103: maybe we also mention something about the above layers mostly being _intended_ to be run on differentiable/PyQ simulators rather than Pulser, although in principle it could be done? The reason we don't encourage Pulser with this is because in any case it is not a HW-realistic sequence, for that we would use the full realistic addressing MR.

tests/constructors/test_rydberg_hea.py
```diff
@@ -0,0 +1,64 @@
...
+    assert ndet_params == 1 if not detunings else n_qubits + 1
+    assert len(wait_layer.parameters.names()) == 1
+    if not phase:
+        # the +2 comes from the time evolution parameter and the scaling factor
+        assert ndrive_params == 1 if not drives else n_qubits + 2
+    else:
+        assert ndrive_params == 4 if not drives else n_qubits + 4
+
+
+def test_rydberg_hea_differentiation() -> None:
```
vincentelfving, line 43: good!

vincentelfving (approved):
Looks good to me, only text suggestions basically. Thanks for improving and cleaning up the suggested functions from the original.

docs/digital_analog_qc/rydberg-hea.md
madagra: Yes, good point. I wanted to say something like: "determine the interaction strength" but I could not find the right wording. I will change it.

docs/digital_analog_qc/daqc-basics.md
madagra: I agree. However, the DAQC as it has been introduced in the paper we cite, requires fully local addressability. That is why I mention it in this way. I will rephrase a bit.

docs/digital_analog_qc/rydberg-hea.md
madagra: It is just that in the code it is not actually implemented, so I did not want to introduce it here and then make it confusing for the user.

PR #159, opened 2023-11-03
[Fix] Pass diffmode in execution

dominikandreasseitz:


awennersteen (approved):
lgtm

qadence/backends/pytorch_wrapper.py
```diff
@@ -315,13 +315,16 @@ def convert(
         observable: list[AbstractBlock] | AbstractBlock | None = None,
     ) -> Converted:
         if self.diff_mode != DiffMode.AD and observable is not None:
+            msg = (
+                f"Differentiation mode '{self.diff_mode}' does not support parametric observables."
```
awennersteen, line 319:
```suggestion
                f"Differentiation mode {self.diff_mode} does not support parametric observables."
```

awennersteen: with f strings that works, right?

dominikandreasseitz: i prefer "ValueError: Differentiation mode 'adjoint' does not support parametric observables." (so the quote, if you dont mind)

PR #162, opened 2023-11-06
[Fix] Braket CPHASE always having fixed parameters

dominikandreasseitz:
Previously, the cphase gate in braket was ALWAYS initialized with a fixed parameter instead of checking if the underlying expression contains trainable symbols, now its possible

### DIFF
--- a/qadence/backends/braket/convert_ops.py
+++ b/qadence/backends/braket/convert_ops.py
@@ -81,13 +81,19 @@ def BraketOperation(block: PrimitiveBlock) -> Instruction:
         return two_qubit[operation](block.qubit_support[0], block.qubit_support[1])
 
     elif operation in two_qubit_parametrized:
-        (expr,) = block.parameters.expressions()  # type: ignore [attr-defined]
-        angle_value = evaluate(expr)
-        return two_qubit_parametrized[operation](
-            control=block.qubit_support[0],
-            target=block.qubit_support[1],
-            angle=angle_value,
-        )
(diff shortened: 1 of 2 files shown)


jpmoutinho (approved):

Roland-djee (approved):
Thanks !

nmheim:
Could we add a brief description what was the bug here?:)

qadence/backends/braket/convert_ops.py
```diff
@@ -81,13 +81,19 @@ def BraketOperation(block: PrimitiveBlock) -> Instruction:
...
+                control=block.qubit_support[0],
+                target=block.qubit_support[1],
+                angle=evaluate(expr),
+            )
+        else:
+            return two_qubit_parametrized[operation](
+                control=block.qubit_support[0],
+                target=block.qubit_support[1],
+                angle=FreeParameter(uuid),
+            )
```
madagra, line 96: I do not really understand what is happening here. If the expression is an actual `sympy.Expr` and you do not evaluate it, you'll have a bug, I think.

madagra (commented):
I do not fully understand the fix here.

qadence/backends/braket/convert_ops.py
dominikandreasseitz: look at how the single qubit parametric gate is done, we check if the expr is fixed (i.e., is_number). if not, its a a expression with trainable/non-trainable (but not fixed) parameters

madagra (approved):

PR #164, opened 2023-11-07
[Docs] Add simplified train loop example to docs

nmheim:


docs/qml/ml_tools.md
```diff
@@ -209,3 +209,76 @@ for i in range(n_epochs):
...
+    optimizer: Optimizer,
+    config: TrainConfig,
+    loss_fn: Callable,
+    device: str = "cpu",
+    optimize_step: Callable = optimize_step,
+    write_tensorboard: Callable = write_tensorboard,
+) -> tuple[Module, Optimizer]:
+
+    # Move model to device before optimizer is loaded
+    model = model.to(device)
```
awennersteen, line 250: would it be an idea in general to have `device = None` so that if the user happen to have already put everything where they want it we don't make any changes to that? 

Or would that just be more complicated and error prone for 99.99% of users?

nmheim: the idea of our training loop was to cover something like 80% of use cases and everything that goes beyond can be implemented by users themselves so that we don't end up with a `train` loop beast in qadence. this is why I want to add this example of a simplified train loop to the docs. If people want to take care of device things before `train`, the can remove the two related lines in this template:)

awennersteen (approved):

docs/qml/ml_tools.md
awennersteen: I do agree, but at the same time I wonder if it could potentially fix a bug in the future.
On the other hand it could introduce bugs too... so hard to say.

LGTM

Issue #138, opened 2023-10-26
Better documentation on how to work with interactions


madagra:
Documentation on the `add_interaction()` function and on the emulated analog interface in general is currently quite confusing.

Issue #132, opened 2023-10-24
`add_interaction` explanation in the documentation


CdeTerra:
In the [section explaining `add_interaction`](https://pasqal-io.github.io/qadence/v1.0.1/qadence/transpile/#qadence.transpile.emulate.add_interaction), we can read the following:

> You can also use `add_interaction` directly on a block, but you have to provide either the `Register` or define a non-global qubit support.

Maybe an note could clarify what it means, since the following code works without any *explicit* register:

```
circ = kron(chain(H(0),CNOT(0, 1)), chain(H(2),CNOT(2,3)))
add_interaction(circ)
```

Issue #108, opened 2023-10-17
[Refactoring, DA, Documentation] `AnalogRX`'s blockade radius


CdeTerra:
In the [documentation](https://pasqal-io.github.io/qadence/v1.0.1/) on the "digital-analog example" section, one can find the following code:

```python
from torch import pi
from qadence import Register, AnalogRX, sample

# Global analog RX block.
block = AnalogRX(pi)

# Almost non-interacting qubits as too far apart.
register = Register.from_coordinates([(0,0), (0,15)])  # Dimensionless.
samples = sample(register, block)

# Interacting qubits as close together.
register = Register.from_coordinates([(0,0), (0,5)])
samples = sample(register, AnalogRX(pi))
```

We can't see any explanation why 5 should make atoms close and 15 too far apart to interact. These values are arbitrary and in the implementation of `_analog_rot`, the default blockade radius corresponds to omega defaulted to π. This raises the following points:

- Why is omega defaulted to π?
- Regardless of what it's defaulted to, shouldn't the user have access to the corresponding blockade radius to know what is close and what is too far? They could for example do something like:
  ```python
  Register.from_coordinates([(0,0), (0, qadence.default_blockade / 2)]) # close
  Register.from_coordiantes([(0,0), (0, qadence.default_blockade * 2)]) # far
  ```
  This would allow them to make sure they get a specific behavior. It's true they could use `AnalogRot` instead, but they would loose the syntactic sugar of doing X/Y/Z rotations. Besides, functions like `AnalogRX` which behavior is unpredictable may be pointless.

Roland-djee:
Hey @nmheim would you be able to take that one on whenever possible ?

PR #166, opened 2023-11-07
[Docs] Update emulated analog docs

jpmoutinho:
- In the new docs we don't explain the `add_interaction` function, but instead hide it, with the goal being that it will not be user facing in the next refactors. If we agree, this closes https://github.com/pasqal-io/qadence/issues/132 .
- Regarding https://github.com/pasqal-io/qadence/issues/108, we now make it explicit what each Analog function is doing, and it should be clear that `AnalogRot` should be used to have full control over the parameters. The goal is to improve this in the next refactors. For now we can either close this issue or leave it open until the refactors are done.

Closes https://github.com/pasqal-io/qadence/issues/138.


docs/digital_analog_qc/analog-qcl.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Feature map with a few global analog rotations
+fm = chain(
+    AnalogRX(x),
+    AnalogRY(2*x),
+    AnalogRZ(3*x),
+)
+```
+
+Next, we define the ansatz with parameterized rotations and
```
nmheim, line 26: I think there is something missing after the "and"

docs/digital_analog_qc/analog-basics.md
```diff
@@ -1,202 +1,362 @@
...
 
-# Or use the shortcut.
-block = AnalogRX(torch.pi)
-print(f"AnalogRX = {block}") # markdown-exec: hide
+print(wf)
 ```
 
-!!! note "Automatic emulation in the PyQTorch backend"
+<details>
+    <summary>Under the hood of AnalogRot</summary>
```
nmheim, line 110: amazing<3

docs/digital_analog_qc/analog-basics.md
```diff
@@ -1,202 +1,362 @@
...
+# Some random initial state
+init_state = random_state(n_qubits)
+
+# Compare the final state using the full digital and the AnalogRX
+wf_analog_pyq = run(
+    reg,
+    rot_analog,
+    state = init_state,
+    backend = BackendName.PYQTORCH
+    )
```
nmheim, line 240: nitpick :P
```suggestion
)
```

nmheim (approved):
Big fan of what is happening here. Thanks a lot! Just a few very nitpicky comments:)

docs/digital_analog_qc/analog-qcl.md
```diff
@@ -0,0 +1,128 @@
...
+they can be optimized. The next snippet examplifies the creation of an analog and parameterized ansatz
+to fit a simple function. First, define a register and feature map block:
+
+```python exec="on" source="material-block" session="qcl"
+from qadence import Register, FeatureParameter, chain
+from qadence import AnalogRX, AnalogRY, AnalogRZ, wait
+
+# Line register
+n_qubits = 2
+dx = 8.0
```
Roland-djee, line 11:
```suggestion
dx = 8.0. # Spacing (units ?).
```

docs/digital_analog_qc/analog-qcl.md
```diff
@@ -0,0 +1,128 @@
...
+coordinates = [(i*dx, 0) for i in range(n_qubits)]
+register = Register.from_coordinates(coordinates)
+
+# The input feature parameter
+x = FeatureParameter("x")
```
Roland-djee, line 16: Would it make sense to rename it something that resembles a rotation angle ? Ie. `theta` ? So as not to confuse with the coordinates.

docs/tutorials/register.md
```diff
@@ -90,14 +90,16 @@ print(docsutils.fig_to_html(fig)) # markdown-exec: hide
 ```
 
 !!! warning "Units for qubit coordinates"
-    Qubits coordinates in Qadence are *dimensionless* but converted to the required unit when executed on a backend.
-	For instance, [Pulser](https://github.com/pasqal-io/Pulser) uses $\mu \textrm{m}$.
+    In general, Qadence makes no assumption about the units given to qubit coordinates.
+    However, if used in the context of a Hamiltonian coefficient, the quantity $Ht$
```
Roland-djee, line 94:
```suggestion
    However, if used in the context of a Hamiltonian coefficient, the quantity $H.t$
```

docs/tutorials/register.md
```diff
@@ -90,14 +90,16 @@ print(docsutils.fig_to_html(fig)) # markdown-exec: hide
...
+    must be **dimensionless** for exponentiation in the PyQTorch backend, where it is assumed that $\hbar = 1$.
```
Roland-djee, line 95:
```suggestion
    must be **dimensionless** for exponentiation in the PyQTorch backend, where it is assumed that $\hbar = 1$ (also called atomic units).
```

Roland-djee (approved):
Minor comments from me. Otherwise looking good. Thanks

Roland-djee (approved):
Minor comments from me. Otherwise looking good. Thanks

docs/digital_analog_qc/analog-qcl.md
jpmoutinho: I think I would rather keep it as "x" in the sense of "learning f(x)". It does get encoded into an angle, but the meaning of that feature is not necessarily an angle.

docs/tutorials/register.md
jpmoutinho: I think I would rather not call them atomic units because we are not dealing with any other atomic related constants, just setting hbar to 1. Natural unit systems can be very general and arbitrary anyway, and Hartree atomic units is just one possible system. I'm sure there are others that also set hbar to 1.

docs/digital_analog_qc/analog-qcl.md
Roland-djee: Right OK. I would still try to disambiguate the role of that variable in the comment if possible.

docs/tutorials/register.md
Roland-djee: OK. Related to the above, H.t has [pi rad] as phase unit right ?

jpmoutinho: If `hbar = 1` then `H.t` needs to be dimensionless so it can indeed represent an "angle" in the exponential, which you could write as "rads", but radians are not physical units.

jpmoutinho: I've kept `rad` in the units of the docs for now because they do it in Pulser but I would be in favour of eventually removing that.

PR #173, opened 2023-11-09
[FIX] Enable Python 3.11 by fixing dataclass defaults

awennersteen:
Changes required to enable python 3.11

I think it should be refactored a bit, and also check if frozen can be re-enabled.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,12 +19,6 @@ repos:
     hooks:
       - id: black
 
--   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.5.1
-    hooks:
-      - id: mypy
-        exclude: examples|docs
-
 -   repo: https://github.com/DanielNoord/pydocstringformatter
     rev: v0.7.3
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,4 +29,3 @@ repos:
     rev: v0.7.3
     hooks:
       - id: pydocstringformatter
-
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.5.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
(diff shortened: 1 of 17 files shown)


qadence/backends/braket/backend.py
```diff
@@ -41,21 +41,38 @@ def promote_parameters(parameters: dict[str, Tensor | float]) -> dict[str, float
...
+        self.is_remote = is_remote
+        self.with_measurements = with_measurements
+        self.with_noise = with_noise
+        self.native_endianness = native_endianness
+        if config is None:
+            self.config = Configuration()
+        elif isinstance(config, dict):
+            self.config = Configuration(**config)
+        elif isinstance(config, Configuration):
+            self.config = config
```
awennersteen, line 71: I guess we could do this in the parent class, call super, and keep this. 
This fixes the default_dict issue in python 3.11, it requried removing frozen (can investigate), and then in the past it was possible to pass config as a dict, hence the isinstance dict

tests/backends/pulser_basic/test_configuration.py
```diff
@@ -70,5 +71,5 @@ def test_configuration_as_dict() -> None:
     assert not np.isclose(div, 0.0, rtol=1e-2, atol=1e-2)
 
     wrong_conf = {"wrong": "value"}
-    with pytest.raises(ValueError):
+    with pytest.raises(TypeError):
```
awennersteen, line 74: different type here

pyproject.toml
```diff
@@ -38,7 +39,7 @@ dependencies = [
     "deepdiff",
     "jsonschema",
     "nevergrad",
-    "scipy<1.11",
+    "scipy",
```
nmheim, line 42: are we ok with this?

tests/backends/pulser_basic/test_configuration.py
nmheim: hmm, I had to change this back to `ValueError` to make the test pass, or did you change this because we should convert that thrown `ValueError` to a `TypeError`?

qadence/backends/braket/backend.py
nmheim: Instead of this I added `field(default_config=Configuration)` everywhere. are we ok with that?

.pre-commit-config.yaml
```diff
@@ -19,12 +19,6 @@ repos:
     hooks:
       - id: black
 
--   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.5.1
-    hooks:
-      - id: mypy
-        exclude: examples|docs
-
```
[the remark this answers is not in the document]
awennersteen: my bad.
I had a temporarily failing mypy but wanted to run the pipeline since I didn't have 3.11 locally.
then I forgot to remove

tests/backends/pulser_basic/test_configuration.py
awennersteen: interesting, on my end suddenly it started raising `TypeError` rather than ``ValueError`

tests/backends/pulser_basic/test_configuration.py
```diff
@@ -44,6 +44,7 @@ def test_configuration() -> None:
     assert not np.isclose(div, 0.0, rtol=1e-2, atol=1e-2)
 
 
+# @pytest.mark.xfail(reason="TODO, FIXME")
```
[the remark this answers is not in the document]
nmheim: ahh, yes, thank you!

madagra (approved):
LGTM.

PR #184, opened 2023-11-15
[Feature] Add semi-local addressing

vytautas-a:
Implement semi-local addressing using amplitude/detuning patterns.


qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+                warn("Maximum absolute value of amplitude is exceeded")
+        if not isinstance(self.max_det, Parameter):
+            if self.max_det > DEFAULT_MAX_DETUNING:
+                warn("Maximum absolute value of detuning is exceeded")
+
+        # augment weight dicts if needed
+        self.weights_amp = {
+            i: Parameter(0.0) if i not in self.weights_amp else self.weights_amp[i]
+            for i in range(self.n_qubits)
+        }
```
madagra, line 113: What if the other parameters are not variational? You would have some fixed weights and some variational weights mixed up, I guess, right? I would just use the same approach for all the weights.

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # validate number of qubits in mask
+        if max(list(self.weights_amp.keys())) >= self.n_qubits:
+            raise ValueError("Amplitude weight specified for non-existing qubit")
+        if max(list(self.weights_det.keys())) >= self.n_qubits:
+            raise ValueError("Detuning weight specified for non-existing qubit")
+
+    def evaluate(self, weights: dict, values: dict) -> dict:
+        # evaluate weight expressions with actual values
+        return {k: evaluate(v, values, as_torch=True).flatten() for k, v in weights.items()}  # type: ignore [union-attr]
```
madagra, line 138: Cool!

qadence/analog/interaction.py
```diff
@@ -139,56 +143,94 @@ def _(
...
+    return _add_interaction(kron(b), register, interaction, pattern)
 
 
 @_add_interaction.register
-def _(block: AnalogKron, register: Register, interaction: Callable) -> AbstractBlock:
+def _(
+    block: AnalogKron,
+    register: Register,
+    interaction: Callable,
+    pattern: Union[AddressingPattern, None],
```
madagra, line 228: Why not make `pattern` as a keyworded argument?

qadence/analog/utils.py
```diff
@@ -130,3 +132,21 @@ def rot_generator(block: ConstantAnalogRotation) -> AbstractBlock:
...
+        weights_amp = pattern.weights_amp
+        weights_det = pattern.weights_det
+    else:
+        max_amp = 0.0
+        max_det = 0.0
+        weights_amp = {i: 0.0 for i in support}
+        weights_det = {i: 0.0 for i in support}
+
+    p_drive_terms = (1 / 2) * max_amp * add(X(i) * weights_amp[i] for i in support)
+    p_detuning_terms = -max_det * add(0.5 * (I(i) - Z(i)) * weights_det[i] for i in support)
```
madagra, line 151: Can you use `N(i)` instead? It should be equivalent and more clear.

qadence/backends/pulser/config.py
```diff
@@ -81,6 +82,9 @@ class Configuration(BackendConfiguration):
     # configuration for cloud simulations
     cloud_configuration: Optional[CloudConfiguration] = None
 
+    addressing_pattern: AddressingPattern | None = None
+    """Semi-local addressing pattern."""
+
```
madagra, line 87: How do you specify an addressing pattern for the PyQ backend?

tests/analog/test_patterns.py
```diff
@@ -0,0 +1,125 @@
...
+    # define pulser backend
+    pulser_backend = PulserBackend(config=conf)  # type: ignore[arg-type]
+    conv = pulser_backend.convert(circ, obs)
+    pulser_circ, pulser_obs, embedding_fn, params = conv
+    diff_backend = DifferentiableBackend(pulser_backend, diff_mode=DiffMode.GPSR)
+    expval_pulser = diff_backend.expectation(pulser_circ, pulser_obs, embedding_fn(params, values))
+
+    # define pyq backend
+    int_circ = add_interaction(circ, spacing=spacing, pattern=p)
+    pyq_backend = PyqBackend()  # type: ignore[arg-type]
```
madagra, line 65: This kind of breaks the interface for the `Analog` blocks, right? AFAIK as of today the interaction is added automatically for the analog blocks in a transpilation pass for the PyQ backend. I guess this is temporary while @jpmoutinho is revising the whole structure for the emulated analog interface, is that right?

tests/analog/test_patterns.py
```diff
@@ -0,0 +1,125 @@
...
+    "max_amp,max_det",
+    [(0.0, 10.0), (15.0, 0.0), (15.0, 9.0)],
+)
+@pytest.mark.parametrize(
+    "spacing",
+    [8.0, 30.0],
+)
+def test_pulser_pyq_addressing(max_amp: float, max_det: float, spacing: float) -> None:
+    n_qubits = 3
+    block = AnalogRX("x")
```
madagra, line 34: Can you please use a more complicated circuit with at least all the analog rotations?

madagra (commented):
Looks almost there to me, even though the interface is currently a bit confusing, I fear. This will very likely be fixed after #116 and the overhaul of the emulated analog interface. If you did not agree this with @jpmoutinho, let's discuss a bit further.

tests/analog/test_patterns.py
jpmoutinho: Indeed the usage of `add_interaction` here is temporary. You can check the `add_background_hamiltonian` function in https://github.com/pasqal-io/qadence/pull/176/ (parse_analog.py) which is where @vytautas-a and me envisioned the addressing pattern will be handled. That function replaces `add_interaction` and was renamed given that it will no longer be just the interaction. Then it is still handled automatically by the transpilation pass in the PyQ backend which will get the addressing pattern from the configuration the user makes in the device.

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _restrict_weights(self) -> None:
+        self.weights_amp = {
+            k: v * (sigmoid(v, 20, 0.0) - sigmoid(v, 20.0, -1.0))
+            for k, v in self.weights_amp.items()
+        }
+        self.weights_det = {
+            k: v * (sigmoid(v, 20.0, 0.0) - sigmoid(v, 20.0, -1.0))
+            for k, v in self.weights_det.items()
+        }
```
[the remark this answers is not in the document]
vytautas-a: yes

qadence/analog/utils.py
vytautas-a: It seems that no - since `N` operation doesn't have a generator or generator eigenvalues defined. Because of this when calculating derivatives an exception is thrown.

qadence/backends/pulser/config.py
vytautas-a: Currently it's passed as an argument to `add_interaction` function. This of course will change with rework of analog interface.

tests/analog/test_patterns.py
vytautas-a: It's added in transpilation pass, yes, but you can call this function again externally if you want to set spacing or, in this case, pass the pattern.

tests/analog/test_patterns.py
vytautas-a: sure

qadence/analog/interaction.py
vytautas-a: This will be changed anyway in @jpmoutinho's PR.

qadence/analog/addressing.py
vytautas-a: Yes, some mixing can happen that might be confusing. I can write some checks that enforce either all weights to be constant or all - variational. If user passes a one or more constant values as weights that are between 0 and 1 - everything is ok and all other weights are set constant and equal to 0. If one or more weights are passed as variational parameters - the rest of parameters are also set constant, thus not trainable. Thus all weights will be trainable only if the user explicitly passes weights for each qubit.

qadence/analog/utils.py
vytautas-a: The real reason for using `I` and `Z` instead of `N` is that when we have a prefactor in front of `N` it is treated as a `ScaleBlock` and gets assigned a uuid. This in turn means that `uuid_to_eigen` function tries to get the eigenvalues of `N` that it doesn't have.

jpmoutinho: Hmm strange, I thought I added the eigenvalues to N

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+            ):
+                raise ValueError("Detuning addressing pattern weights must sum to 1.0")
+
+        # validate detuning value
+        if not isinstance(self.max_amp, Parameter):
+            if self.max_amp > DEFAULT_MAX_AMPLITUDE:
+                warn("Maximum absolute value of amplitude is exceeded")
+        if not isinstance(self.max_det, Parameter):
+            if self.max_det > DEFAULT_MAX_DETUNING:
+                warn("Maximum absolute value of detuning is exceeded")
```
a-corni, line 107: Here I suggest to compare max* max(weights) with DEFAULT_MAX, and add a test max* sum(weights) with DEFAULT_GLOBAL_MAX (value of 2000)

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+    weights_amp: dict[int, float | torch.Tensor | Parameter]
+
+    # list of weights for fixed detuning pattern that cannot be changed during the execution
+    weights_det: dict[int, float | torch.Tensor | Parameter]
+
+    # maximum amplitude can also be chosen as a variational parameter if needed
+    max_amp: float | torch.Tensor | Parameter = DEFAULT_MAX_AMPLITUDE
+
+    # maximum detuning can also be chosen as a variational parameter if needed
+    max_det: float | torch.Tensor | Parameter = DEFAULT_MAX_DETUNING
```
a-corni, line 46: There might be something I am missing, but won't it be easier to maintain if you only had `weights` and `max_value` instead of `weights_amp`/`weights_det` and `max_amp`/`max_det` ? It seems to me that you are doing the same operations on `weights_amp` and `weights_det`, and on `max_amp` and `max_det`. You would either have to generate two instances of AddressingPattern (one for amp and one for det) or create a class that takes weights_amp,weights_det and max_amp, max_det and uses the modified AddressingPattern to update values ?

qadence/analog/addressing.py
a-corni: I don't know much here but perhaps you can convert it directly in this passage ?
```suggestion
        has_parameter = np.any([isinstance(Parameter, self.weights_amp[i] for i in range(self.n_qubits)])
        for i in range(self.n_qubits):
            if i not in self.weights_amp:
                self.weights_amp[i] = (Parameter(0.0) if has_parameter else 0.0) 
            elif has_parameter and not isinstance(Parameter, self.weights_amp[i]):
                self.weights_amp[i] = Parameter(self.weights_amp[i])
```

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+        }
+        sum_weights_amp = sum(list(self.weights_amp.values()))
+        self.weights_amp = {k: v / sum_weights_amp for k, v in self.weights_amp.items()}
+
+        self.weights_det = {
+            k: Parameter(v) if not isinstance(v, Parameter) else abs(v)
+            for k, v in self.weights_det.items()
+        }
+        sum_weights_det = sum(list(self.weights_det.values()))
+        self.weights_det = {k: v / sum_weights_det for k, v in self.weights_det.items()}
```
a-corni, line 64: I would delete this function and add a condition in _restrict_weights

qadence/analog/addressing.py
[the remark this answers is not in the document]
a-corni: In pulser, I have restricted the weights to [0, min(1, MAX_GLOBAL/sum(weights))] I don't know how that would apply here unfortunately :)

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+            for k, v in self.weights_det.items()
+        }
+
+    def _restrict_max_vals(self) -> None:
+        self.max_amp = self.max_amp * (
+            sympy.Heaviside(self.max_amp) - sympy.Heaviside(self.max_amp - DEFAULT_MAX_AMPLITUDE)
+        )
+        self.max_det = self.max_det * (
+            sympy.Heaviside(self.max_det) - sympy.Heaviside(self.max_det - DEFAULT_MAX_DETUNING)
+        )
```
a-corni, line 82: In pulser, I have restricted the values to [0, min(DEFAULT_MAX_LOCAL, DEFAULT_MAX_GLOBAL/sum(weights))] I guess I would replace DEFAULT_MAX by min(MAX_LOCAL, MAX_GLOBAL/sum(weights)) :)

a-corni (commented):
Few comments on qadence/analog.addressing.py, mainly on the last deletion of the normalization, but also joining Mario in his comment that we are doing things twice in this class.

qadence/analog/addressing.py
vytautas-a: I see now your point. From code perspective the applied transformations for amplitude/detuning weights are similar, however how they integrate in later stages of `qadence` execution pipeline is different. So I'm currently refactoring the duplicated code into functions but keeping separate `weights_amp`, `weights_det`, etc.

qadence/analog/addressing.py
vytautas-a: Deleted this function.

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+    weights_det: dict[int, float | torch.Tensor | Parameter]
+
+    # maximum amplitude can also be chosen as a variational parameter if needed
+    max_amp: float | torch.Tensor | Parameter = DEFAULT_MAX_AMPLITUDE
+
+    # maximum detuning can also be chosen as a variational parameter if needed
+    max_det: float | torch.Tensor | Parameter = DEFAULT_MAX_DETUNING
+
+    # weight constraint
+    weight_constraint: WeightConstraint = WeightConstraint.NORMALIZE
```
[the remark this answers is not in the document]
vytautas-a: deleted

qadence/analog/addressing.py
```diff
@@ -0,0 +1,138 @@
...
+
+DEFAULT_MAX_AMPLITUDE = 2 * pi * 3
+DEFAULT_MAX_DETUNING = 2 * pi * 20
+
+
+class WeightConstraint(StrEnum):
+    """Supported types of constraints for addressing weights."""
+
+    NORMALIZE = "normalize"
+    """Normalize weights so that they sum up to 1."""
```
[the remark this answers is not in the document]
vytautas-a: Removed this.

qadence/analog/addressing.py
vytautas-a: I added these two new constraints.

qadence/backends/pulser/pulses.py
```diff
@@ -42,6 +42,54 @@
...
+) -> None:
+    total_duration = sequence.get_duration()
+    n_qubits = len(sequence.register.qubits)
+
+    support = tuple(range(n_qubits))
+    if config.addressing_pattern is not None:
+        amp = config.addressing_pattern.amp
+        det = config.addressing_pattern.det
+        weights_amp = config.addressing_pattern.weights_amp
+        weights_det = config.addressing_pattern.weights_det
```
madagra, line 57: Is this assuming that the pattern weights are given as numbers for the Pulser backend?

madagra (approved):
@vytautas-a it looks good to me now. I pre-approve but please add some documentation even if drafty. It is **very** important to mention the current limitations of the addressing patterns.

qadence/backends/pulser/pulses.py
vytautas-a: Currently, weights for the Pulser backend can only be fixed. I'll open a separate issue for that since it will require more changes.

madagra: Great, makes sense. Could you please add a comment on that? It's quite important. Then go ahead with merging this one.

qadence/analog/addressing.py
```diff
@@ -0,0 +1,163 @@
...
+            i: Parameter(0.0)
+            if i not in weights
+            else (Parameter(weights[i]) if not isinstance(weights[i], Parameter) else weights[i])
+            for i in range(self.n_qubits)
+        }
+
+        # restrict weights to [0, 1] range
+        weights = {
+            k: abs(v * (sigmoid(v, 20, 1.0) - sigmoid(v, 20.0, -1.0))) for k, v in weights.items()
+        }
```
jpmoutinho, line 74: If I recall correctly this makes it so that any weight higher than 1.0 or lower than -1.0 gets effectively counted as 0 right? Can you add a line to the comment saying what it does?

qadence/analog/addressing.py
```diff
@@ -0,0 +1,163 @@
...
+    """Restrict weight values to interval [0, 1]."""
+
+
+def sigmoid(x: torch.Tensor, a: float, b: float) -> sympy.Expr:
+    return 1.0 / (1.0 + sympy.exp(-a * (x + b)))
+
+
+@dataclass
+class AddressingPattern:
+    # number of qubits
```
jpmoutinho, line 35: Maybe use docstrings as in the str enum above for cleaner look?

qadence/analog/addressing.py
```diff
@@ -0,0 +1,163 @@
...
+        weights = {
+            k: abs(v * (sigmoid(v, 20, 1.0) - sigmoid(v, 20.0, -1.0))) for k, v in weights.items()
+        }
+
+        return weights
+
+    def _constrain_max_vals(self) -> None:
+        # enforce constraints:
+        # 0 <= amp <= GLOBAL_MAX_AMPLITUDE
+        # 0 <= abs(det) <= GLOBAL_MAX_DETUNING
```
jpmoutinho, line 81: Also docstring

qadence/analog/addressing.py
```diff
@@ -0,0 +1,163 @@
...
+            * (
+                sympy.Heaviside(self.det + GLOBAL_MAX_DETUNING)
+                - sympy.Heaviside(self.det - GLOBAL_MAX_DETUNING)
+            )
+        )
+
+    def _create_local_constraint(self, val: sympy.Expr, weights: dict, max_val: float) -> dict:
+        # enforce local constraints:
+        # amp * w_amp_i < LOCAL_MAX_AMPLITUDE or
+        # abs(det) * w_det_i < LOCAL_MAX_DETUNING
```
jpmoutinho, line 100: Also docstring

qadence/analog/addressing.py
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        return local_constr
+
+    def _create_global_constraint(
+        self, val: sympy.Expr, weights: dict, max_val: float
+    ) -> sympy.Expr:
+        # enforce global constraints:
+        # amp * sum(w_amp_0, w_amp_1, ...) < GLOBAL_MAX_AMPLITUDE or
+        # abs(det) * sum(w_det_0, w_det_1, ...) < GLOBAL_MAX_DETUNING
```
jpmoutinho, line 113: Also docstring

tests/analog/test_patterns.py
```diff
@@ -0,0 +1,134 @@
...
+    expval_pulser = model.expectation(values=values)
+
+    # define pyq backend
+    int_circ = add_interaction(circ, pattern=p)
+    model = QuantumModel(
+        circuit=int_circ, observable=obs, backend=BackendName.PYQTORCH, diff_mode=DiffMode.AD
+    )
+    expval_pyq = model.expectation(values=values)
+
+    torch.allclose(expval_pulser, expval_pyq, atol=ATOL_DICT[BackendName.PULSER])
```
jpmoutinho, line 75: Could you add an extra test with `model.run` and comparing `equivalent_state`?

tests/analog/test_patterns.py
```diff
@@ -0,0 +1,134 @@
...
+
+@pytest.mark.flaky(max_runs=10)
+def test_addressing_training() -> None:
+    n_qubits = 3
+    reg = Register(support=n_qubits, spacing=8)
+    f_value = torch.rand(1)
+
+    # define training parameters
+    w_amp = {i: Parameter(f"w_amp{i}", trainable=True) for i in range(n_qubits)}
+    w_det = {i: Parameter(f"w_det{i}", trainable=True) for i in range(n_qubits)}
```
jpmoutinho, line 86: Maybe we should allow direct string input here and have those converted to `VariationalParameter` automatically to be consistent with other parameter inputs in Qadence.

jpmoutinho (approved):
Thanks @vytautas-a, looks really nice, a few extra comments from me. Like we already discussed feel free to merge this one when ready even if before the Qadence device one. I already checked what I will need to adapt and it should be straightforward.

qadence/analog/addressing.py
vytautas-a: expanded the comment

qadence/analog/addressing.py
vytautas-a: added docstrings under parameter definitions

tests/analog/test_patterns.py
vytautas-a: added wavefunction comparison

tests/analog/test_patterns.py
vytautas-a: Added this possibility

PR #186, opened 2023-11-15
[Feature] Add coords scaling and distance calculation directly to Register

jpmoutinho:
- Moves the register `spacing` as an argument to the `Register` itself.
- Changes the `spacing` functionality: Instead of multiplying whatever coordinates are in the register with the value of the spacing, it rescales the coordinates so that the distance between the two closest qubits is equal to the spacing (consistent with the spacing argument in the Pulser register).
- Rescaling coordinates from the register can still be done with the method `reg.rescale_coords(scaling: float)` which initializes a new register where all coordinates are multiplied by the scaling value.

BREAKING:
- Some of the arguments from the `Register` methods have been changed so that they are all now `spacing`.
- The default scaling of coordinates in the Pulser backend has been removed so that now a Qadence register passed to either the PyQ backend or Pulser backend gives the same results without needing to tune any other config.
- The `spacing` is no longer a valid argument in the Pulser backend configuration.

### DIFF
--- a/qadence/backends/pulser/backend.py
+++ b/qadence/backends/pulser/backend.py
@@ -36,10 +36,6 @@ from .pulses import add_pulses
 
 logger = get_logger(__file__)
 
-WEAK_COUPLING_CONST = 1.2
-
-DEFAULT_SPACING = 8.0  # µm (standard value)
-
 
 def _convert_init_state(state: Tensor) -> np.ndarray:
     """Flip and squeeze initial state consistent with Pulser convention."""
@@ -48,12 +44,11 @@ def _convert_init_state(state: Tensor) -> np.ndarray:
     return np.flip(state.cpu().squeeze().numpy())
 
 
-def create_register(register: Register, spacing: float = DEFAULT_SPACING) -> PulserRegister:
+def create_register(register: Register) -> PulserRegister:
     """Create Pulser register instance.
 
     Args:
         register (Register): graph representing a register with accompanying coordinate data
-        spacing (float): distance between qubits in micrometers
 
     Returns:
         Register: Pulser register
@@ -61,7 +56,7 @@ def create_register(register: Register, spacing: float = DEFAULT_SPACING) -> Pul
 
     # create register from coordinates
     coords = np.array(list(register.coords.values()))
-    return PulserRegister.from_coordinates(coords * spacing)
+    return PulserRegister.from_coordinates(coords)
 
 
 def make_sequence(circ: QuantumCircuit, config: Configuration) -> Sequence:
@@ -72,27 +67,14 @@ def make_sequence(circ: QuantumCircuit, config: Configuration) -> Sequence:
     else:
         raise ValueError("Specified device is not supported.")
 
-    max_amp = device.channels["rydberg_global"].max_amp
-
-    if config.spacing is not None:
-        spacing = config.spacing
-    elif max_amp is not None:
-        # TODO: Fix this more consistently so both pulser and
-        # pyqtorch get the spacing from the same place
-        # Ideal spacing for entanglement gate
-        # since Pulser's QutipEmulator doesn't allow simulation of sequences
-        # with total duration < 4ns
-        spacing = WEAK_COUPLING_CONST * device.rydberg_blockade_radius(max_amp)  # type: ignore
-    else:
-        spacing = DEFAULT_SPACING
-
-    pulser_register = create_register(circ.register, spacing)
+    pulser_register = create_register(circ.register)
 
     sequence = Sequence(pulser_register, device)
+
     sequence.declare_channel(GLOBAL_CHANNEL, "rydberg_global")
     sequence.declare_channel(LOCAL_CHANNEL, "rydberg_local", initial_target=0)
 
-    add_pulses(sequence, circ.block, config, circ.register, spacing)
+    add_pulses(sequence, circ.block, config, circ.register)
     sequence.measure()
 
     return sequence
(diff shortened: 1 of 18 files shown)


jpmoutinho:
> backend any

Yes exactly, the results would be the same.

jpmoutinho:
> If the spacing is not part of the backend anymore but of the `Register` will it still be possible to run two circuits
> 
> ```python
> run(reg, circ, backend="pyqtorch")
> run(reg, circ, backend="pulser")
> ```
> 
> and get the same result? or would one have to change the register?

Yes exactly, the results would be the same

nmheim:
> > If the spacing is not part of the backend anymore but of the `Register` will it still be possible to run two circuits
> > ```python
> > run(reg, circ, backend="pyqtorch")
> > run(reg, circ, backend="pulser")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and get the same result? or would one have to change the register?
> 
> Yes exactly, the results would be the same

Coooool, so how is this handled?:) does the pyq backend ignore the spacing? sorry for the ignorant questions, just trying to figure out how its implemented:)

qadence/register.py
```diff
@@ -50,22 +52,37 @@ def __init__(self, support: nx.Graph | int):
...
 
+        if scale != 1.0:
+            _scale_node_positions(self.graph, scale)
+
+        # Auxiliary complete graph
+        support = self.graph.nodes
+        all_edges = list(filter(lambda x: x[0] < x[1], product(support, support)))
+        self.complete_graph = nx.Graph()
+        self.complete_graph.add_nodes_from(support)
+        self.complete_graph.add_edges_from(all_edges)
```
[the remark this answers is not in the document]
jpmoutinho: I use it in this MR: https://github.com/pasqal-io/qadence/pull/176, but I figured I would put the changes here since they relate to the register. But actually there's something else I need to add so that `hamiltonian_factory` properly works with it.

Basically `hamiltonian_factory(register, ...)` will create a hamiltonian based on the edges of the register. But if I want for example a hamiltonian dependent on the distance between the qubits, independently of the edges in the graph, that's what this `complete_graph` is used for.

jpmoutinho:
> > > If the spacing is not part of the backend anymore but of the `Register` will it still be possible to run two circuits
> > > ```python
> > > run(reg, circ, backend="pyqtorch")
> > > run(reg, circ, backend="pulser")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and get the same result? or would one have to change the register?
> > 
> > 
> > Yes exactly, the results would be the same
> 
> Coooool, so how is this handled?:) does the pyq backend ignore the spacing? sorry for the ignorant questions, just trying to figure out how its implemented:)

It's all just dependent on the coordinates of the qubits. By passing the scaling in the register the coordinates are automatically re-scaled, and then the pyq backend computes distances between the qubits dependent on the coordinates. When passing to the pulser backend the pulser register is created from the coordinates of the qadence register, so everything is consistent.

qadence/register.py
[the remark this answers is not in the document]
nmheim: awesome that sounds great 👍

qadence/backends/pulser/backend.py
```diff
@@ -72,27 +67,14 @@ def make_sequence(circ: QuantumCircuit, config: Configuration) -> Sequence:
...
-        spacing = config.spacing
-    elif max_amp is not None:
-        # TODO: Fix this more consistently so both pulser and
-        # pyqtorch get the spacing from the same place
-        # Ideal spacing for entanglement gate
-        # since Pulser's QutipEmulator doesn't allow simulation of sequences
-        # with total duration < 4ns
-        spacing = WEAK_COUPLING_CONST * device.rydberg_blockade_radius(max_amp)  # type: ignore
-    else:
-        spacing = DEFAULT_SPACING
```
nmheim, line 89: awesome

qadence/backends/pulser/backend.py
```diff
@@ -48,20 +44,19 @@ def _convert_init_state(state: Tensor) -> np.ndarray:
     return np.flip(state.cpu().squeeze().numpy())
 
 
-def create_register(register: Register, spacing: float = DEFAULT_SPACING) -> PulserRegister:
+def create_register(register: Register) -> PulserRegister:
```
nmheim, line 49: I think this function is used only once, maybe we can even get rid of it now?

nmheim (approved):
LGTM! Last question: what are the units of the coordinates now in pulser/pyq?

jpmoutinho:
> LGTM! Last question: what are the units of the coordinates now in pulser/pyq?

For pulser in general and analog blocks in pyq: micrometers, same as it was. But if a user just does some arbitrary HamEvo to run in PyQ that depends on some register coordinates, they are responsible for setting the units correctly. There is a warning about this in the register docs.

qadence/backends/pulser/backend.py
jpmoutinho: I think it's ok here just for the sake of organization (and potential reuse) but I simplified the docstring

qadence/backends/pulser/backend.py
jpmoutinho: I would like to ask @kaosmicadei about it but he is on vacations 😅

jpmoutinho: Either way I've sent him a message about it to discuss when he's back.

PR #189, opened 2023-11-16
[Refactoring] Improve mitigation protocol

### LINKED ISSUE #187: Improve `Mitigations` protocol
Related to #183 adn #181. Pasting comments to be addressed from #105 here for cleanliness:

- [x] [madagra](https://github.com/madagra) [last week](https://github.com/pasqal-io/qadence/pull/105#discussion_r1383232224)
If you properly implement the run_noisy --> run_dm method this should not be necessary.
- [x] [madagra](https://github.com/madagra) [last week](https://github.com/pasqal-io/qadence/pull/105#discussion_r1383226640)
I honestly don't understand why we have this. The noise model should be something that the user supplies externally via the backend configuration and not in the mitigation routine. Here you can inspect the sim_config object of the input backend to check the noise model but otherwise, it does not really make sense to add the noise directly here. What if you want to extrapolate for real QPU executions, for example?
- [x] [madagra](https://github.com/madagra) [last week](https://github.com/pasqal-io/qadence/pull/105#discussion_r1383228915)
I suggested to remove this run_noisy, now I see how you use it. To make it cleaner, I would define a run_dm function in the backend which returns a density matrix instead of the wavefunction. Only the Pulser backend would implement it, thus making this function fail if the Pulser backend is not used + it makes it also a bit more flexible for future updates (e.g. at some point also PyQ will implement a run_dm method).
- [x] [madagra](https://github.com/madagra) [last week](https://github.com/pasqal-io/qadence/pull/105#discussion_r1383230464)
Are batched dimensions supported as well, right?
- [x]  Harmonise `NoiseTypes` with `Noise`. Currently, analog ZNE uses `NoiseTypes` but readout mitigation uses `Noise`
- [ ] Implement ZNE for pulse stretching

Roland-djee:
Closes #187.


qadence/mitigations/readout.py
```diff
@@ -0,0 +1,94 @@
...
+    """Minimize a correction matrix subjected to stochasticity constraints.
+
+    See Equation (5) in https://arxiv.org/pdf/2001.09980.pdf.
+    """
+    corrected_counters: list[Counter] = []
+    for sample in samples:
+        n_qubits = len(list(sample.keys())[0])
+        n_shots = sum(sample.values())
+        bitstring_length = 2**n_qubits
+        noise_matrices = noise.options.get("noise_matrix", noise.options["confusion_matrices"])
```
gvelikova, line 52: I'd  suggest as a minor restructuring comment to move 52-54 lines above the for loop as they are not sample specific.

qadence/mitigations/readout.py
```diff
@@ -0,0 +1,94 @@
...
+        # Build the whole T matrix.
+        T_matrix = reduce(torch.kron, noise_matrices).detach().numpy()
+        # List of bitstrings in lexicographical order.
+        ordered_bitstrings = [f"{i:0{n_qubits}b}" for i in range(bitstring_length)]
+        # Tensor of raw probabilites.
+        p_raw = torch.tensor([sample[bs] for bs in ordered_bitstrings]) / n_shots
```
gvelikova, line 58: Can p_raw directly be  np.array([sample[bs] for bs in ordered_bitstrings]) / n_shots or the tensor is needed for some precision aspect?

qadence/mitigations/readout.py
```diff
@@ -0,0 +1,94 @@
...
+        positivity_constraint = LinearConstraint(
+            np.eye(bitstring_length).astype(int), lb=0.0, ub=1.0
+        )
+        constraints = [normality_constraint, positivity_constraint]
+        # Minimize the corrected probabilities.
+        res = minimize(corrected_probas, p_corr0, args=(T_matrix, p_raw), constraints=constraints)
+        # breakpoint()
+        corrected_counts = np.rint(res.x * n_shots).astype(int)
+        # Renormalize if total counts differs from n_shots.
+        corrected_counts = renormalize_counts(corrected_counts=corrected_counts, n_shots=n_shots)
```
gvelikova, line 75: Should this line 75 move under the if line 78?

qadence/mitigations/readout.py
gvelikova: Also, these square shape tests for the noise matrix should be added somewhere for the readout mitigation.

gvelikova (approved):

qadence/mitigations/readout.py
Roland-djee: What square shape tests exactly ?

qadence/mitigations/readout.py
```diff
@@ -0,0 +1,94 @@
...
+def mitigation_minimization(
+    noise: Noise, mitigation: Mitigations, samples: list[Counter]
+) -> list[Counter]:
+    """Minimize a correction matrix subjected to stochasticity constraints.
+
+    See Equation (5) in https://arxiv.org/pdf/2001.09980.pdf.
+    """
+    corrected_counters: list[Counter] = []
+    for sample in samples:
+        n_qubits = len(list(sample.keys())[0])
```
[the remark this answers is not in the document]
Roland-djee: Probably not. Will move outside the loop.

qadence/mitigations/readout.py
Roland-djee: Well, the logic is that first renormalise using count redistribution. Then if the count is still off by at most 2, increase the max count by the difference.

PR #206, opened 2023-11-21
[Bugfix] Consistent variable ordering in QNNs

nmheim:
Before this MR we relied on the order of `self.inputs` defined in the `QuantumModel`
```
self.inputs = [p for p in circuit.unique_parameters if not p.trainable and not p.is_number]
```
when casting QNN input tensors to dicts.

### DIFF
--- a/qadence/constructors/hamiltonians.py
+++ b/qadence/constructors/hamiltonians.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import warnings
 from typing import List, Tuple, Type, Union
 
 import numpy as np
@@ -236,23 +235,11 @@ def _update_interaction_strength(
     return register
 
 
-# FIXME: Previous hamiltonian / observable functions, now refactored, to be deprecated:
-
-DEPRECATION_MESSAGE = "This function will be removed in the future. "
-
-
 def single_z(qubit: int = 0, z_coefficient: float = 1.0) -> AbstractBlock:
-    message = DEPRECATION_MESSAGE + "Please use `z_coefficient * Z(qubit)` directly."
-    warnings.warn(message, FutureWarning)
     return Z(qubit) * z_coefficient
 
 
 def total_magnetization(n_qubits: int, z_terms: np.ndarray | list | None = None) -> AbstractBlock:
-    message = (
-        DEPRECATION_MESSAGE
-        + "Please use `hamiltonian_factory(n_qubits, detuning=Z, node_coeff=z_terms)`."
-    )
-    warnings.warn(message, FutureWarning)
     return hamiltonian_factory(n_qubits, detuning=Z, detuning_strength=z_terms)
 
 
@@ -261,20 +248,11 @@ def zz_hamiltonian(
     z_terms: np.ndarray | None = None,
     zz_terms: np.ndarray | None = None,
 ) -> AbstractBlock:
-    message = (
-        DEPRECATION_MESSAGE
-        + """
-Please use `hamiltonian_factory(n_qubits, Interaction.ZZ, Z, interaction_strength, z_terms)`. \
-Note that the argument `zz_terms` in this function is a 2D array of size `(n_qubits, n_qubits)`, \
-while `interaction_strength` is expected as a 1D array of size `0.5 * n_qubits * (n_qubits - 1)`."""
-    )
-    warnings.warn(message, FutureWarning)
     if zz_terms is not None:
         register = Register(n_qubits)
         interaction_strength = [zz_terms[edge[0], edge[1]] for edge in register.edges]
     else:
         interaction_strength = None
-
     return hamiltonian_factory(n_qubits, Interaction.ZZ, Z, interaction_strength, z_terms)
 
 
@@ -284,13 +262,6 @@ def ising_hamiltonian(
     z_terms: np.ndarray | None = None,
     zz_terms: np.ndarray | None = None,
 ) -> AbstractBlock:
-    message = (
-        DEPRECATION_MESSAGE
-        + """
-You can build a general transverse field ising model with the `hamiltonian_factory` function. \
-Check the hamiltonian construction tutorial in the documentation for more information."""
-    )
-    warnings.warn(message, FutureWarning)
     zz_ham = zz_hamiltonian(n_qubits, z_terms=z_terms, zz_terms=zz_terms)
     x_ham = hamiltonian_factory(n_qubits, detuning=X, detuning_strength=x_terms)
     return zz_ham + x_ham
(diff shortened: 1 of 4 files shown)


qadence/constructors/hamiltonians.py
```diff
@@ -236,23 +235,11 @@ def _update_interaction_strength(
...
 
-# FIXME: Previous hamiltonian / observable functions, now refactored, to be deprecated:
-
-DEPRECATION_MESSAGE = "This function will be removed in the future. "
-
-
 def single_z(qubit: int = 0, z_coefficient: float = 1.0) -> AbstractBlock:
-    message = DEPRECATION_MESSAGE + "Please use `z_coefficient * Z(qubit)` directly."
-    warnings.warn(message, FutureWarning)
     return Z(qubit) * z_coefficient
```
Doomsk, line 240: does it mean those functions won't be deprecated anymore?

qadence/models/qnn.py
```diff
@@ -43,6 +44,7 @@ def __init__(
         self,
         circuit: QuantumCircuit,
         observable: list[AbstractBlock] | AbstractBlock,
+        inputs: tuple[sympy.Basic | str, ...] | None = None,
         transform: Callable[[Tensor], Tensor] = None,  # transform output of the QNN
         backend: BackendName = BackendName.PYQTORCH,
```
Doomsk, line 49: can the `inputs` rely on `Parameter` (`FeatureParamater`, for instance)? I'm thinking in terms of custom feature map circuits

Doomsk (commented):
seems good to me! Some comments on the usability of the new inputs and the deprecation functions and then it's good to go

qadence/constructors/hamiltonians.py
nmheim: yes

qadence/models/qnn.py
nmheim: `Parameter` is subclassing `sympy.Basic`, so yes:)

qadence/models/qnn.py
```diff
@@ -87,15 +87,17 @@ def __init__(
...
                 `xs = torch.rand(batch_size, input_size:=3)` to the QNN `svars = ("t", "x", "y")`
-                will make the QNN use `t = xs[:,0]`, etc. You can also pass a tuple of sympy
+                will make the QNN use `t = xs[:,0]`, etc. You can also pass a list of sympy
                 symbols.
-            """)
+            """
+            )
         elif len(self.inputs) > 1:
-            self.inputs = tuple(sympy.symbols(x) if isinstance(x, str) else x for x in inputs)
+            self.inputs = [sympy.symbols(x) if isinstance(x, str) else x for x in inputs]  # type: ignore[union-attr]
```
Doomsk, line 100: got strings split? 🥲

qadence/finitediff.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
+
+from typing import Callable
+
+import torch
+from torch import Tensor
+
+
+def finitediff(
```
dominikandreasseitz, line 9: very cool! thank you!

qadence/models/quantum_model.py
```diff
@@ -77,7 +77,9 @@ def __init__(
                 f"The circuit should be of type '<class QuantumCircuit>'. Got {type(circuit)}."
             )
 
-        self.inputs = [p for p in circuit.unique_parameters if not p.trainable and not p.is_number]
+        self.domain_vars = [
+            p for p in circuit.unique_parameters if not p.trainable and not p.is_number
+        ]
```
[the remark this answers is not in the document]
dominikandreasseitz: i would vote against renaming

qadence/finitediff.py
```diff
@@ -0,0 +1,47 @@
...
+        f: Function to differentiate
+        x: Input of shape `(batch_size, input_size)`
+        derivative_indices: which *input* to differentiate (i.e. which variable x[:,i])
+        eps: finite difference spacing
+    """
+
+    if eps is None:
+        order = len(derivative_indices)
+        # FIXME: this is the float64 machine epsilon, should probably be fixed for float32
+        eps = 2.220446049250313e-16 ** (1 / (2 + order))
```
madagra, line 27: @nmheim Isn't there a constant available somewhere in the Python standard library for machine precision?

qadence/finitediff.py
```diff
@@ -0,0 +1,47 @@
...
+        di = derivative_indices[1:]
+        return (finitediff(f, x + ev, di) - finitediff(f, x - ev, di)) * _eps / 2
+    elif len(derivative_indices) == 3:
+        return (f(x + 2 * ev) - 2 * f(x + ev) + 2 * f(x - ev) - f(x - 2 * ev)) * _eps**3 / 2
+    elif len(derivative_indices) == 2:
+        return (f(x + ev) + f(x - ev) - 2 * f(x)) * _eps**2
+    elif len(derivative_indices) == 1:
+        return (f(x + ev) - f(x - ev)) * _eps / 2
+    else:
+        raise
```
madagra, line 47: @nmheim You should do a better exception here.

qadence/models/qnn.py
```diff
@@ -81,9 +89,22 @@ def __init__(
...
+                symbols.
+            """
+            )
+        elif len(self.domain_vars) > 1:
+            self.domain_vars = [sympy.symbols(x) if isinstance(x, str) else x for x in domain_vars]  # type: ignore[union-attr]
+
     def forward(
         self,
-        values: dict[str, Tensor] | Tensor = None,
+        values: Tensor,
```
madagra, line 107: Is the type right? You are still keeping backward compatibility below.

qadence/models/quantum_model.py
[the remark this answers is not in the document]
madagra: Same here.

madagra (commented):
Looks overall good to me, I would just avoid the renaming and let the libraries downstream handle it.

dominikandreasseitz:
@nmheim thanks for this, nice idea, however breaking. can we somehow retain backwards compatibility ? @madagra wdyt?

nmheim:
> @nmheim thanks for this, nice idea, however breaking. can we somehow retain backwards compatibility ? @madagra wdyt?

True, the additional argument to the QNN is breaking, but without specifying it we might end up with incorrectly mapped variables in QNNs. So it's a quite critical fix IMO.

dominikandreasseitz (approved):

Roland-djee:
@nmheim Is this ready to be merged ? Or do you need more review/time ?

PR #208, opened 2023-11-21
[Feature] Projector blocks

### LINKED ISSUE #193: Implement a general Projector(PrimitiveBlock) i.e. |1><1|
I would like to request a feature to implement a projector-block such that a user can create a matrix-block like |1><1| or |01><10|. These are in general non-unitary, but can be handled by backends such as PyQ in-principle, and occur often in QML codes
How it would look like in usage:

```
number_operator = Projector([1], [1]) # measures the occupation of the 1 state on qubit 0
swap_but_throw_away_equals = Projector([0,1], [1,0], support = [2,3]) # swaps the population of the 01 state and the 10 state on two qubits on indices 2 and 3
```

suggested class/interface, open for suggestions:

```python
class Projector(PrimitiveBlock):
   '''
      this block makes a state project from `bra` to `ket`, and can be initialized by a tuple/list of 0's and 1's, or even by fully custom torch vectors/matrices.
   '''
    def __init__(self, ket: Union[Iterable[int], torch.Tensor], bra: Union[Iterable[int], torch.Tensor], support: Iterable[int] = None)
        
        n_qubits = None
        for wf in [ket, bra]
            if isinstance(wf, Iterable):
                assert all([(item==0 or item==1) for item in wf]), f"Make sure all qubits in {wf} are either in the 0 or 1 state"
                if support is None:
                     if n_qubits is None:
                          n_qubits = len(wf)
                     else:
                          assert n_qubits==len(wf)
                          self.support = support
                else:
                     assert len(wf)=len(support), f"make sure the support and {wf} are the same in number of qubits"
                     n_qubits = len(support)
                     self.support = support
            if isinstance(ket, torch.Tensor):
                if support is not None:
                     assert np.log2(len(wf))==len(support), f"make sure the support and {wf} are the same in number of qubits"
                     n_qubits = len(support)
                     self.support = support
                else:
                     n_qubits = np.log2(len(wf))
                     self.support = tuple(range(n_qubits))
        self.ket = ket
        self.bra = bra
```

and we can then make helper constructors like this:

```
function n(index: int):
     return Projector([1], [1], support=[index]
```

which is actually the same operator as `(I-Z)/2` but much cleaner.

These projectors can be handled natively on all backends in the observable case, and in some cases also \_inside \_the circuit as non-unitary operators (like in PyQ).

Please note that in-principle these projectors can be handled like (dense) matrices, such as the suggested `MatrixBlock`, but I think that is not smart both from performance perspective (no need to instantiate a full matrix, the representation is very sparse in Projectors) and from code maintenance point of view (but I let you judge).

I would also like this to work with control-unitary approach, e.g. apply this block conditional on, controlled by, another qubit, as usual.

Thanks beforehand for your help!

Roland-djee:
Closes #193.

Few points to consider:

- [x]  Preconditions checks: bra,ket and support must be of the same length.
- [x] qubit_support as a positional argument.
- [x] Projector should be block_to_tensorable.
- [x] Express `N` as a projector
- [x] Integrate `N` in controlled blocks, i.e. Toffoli.
- [x] Integration with PyQ


qadence/blocks/primitive.py
```diff
@@ -454,3 +454,34 @@ def _block_title(self) -> str:
...
+
+class ProjectorBlock(PrimitiveBlock):
+    """The abstract ProjectorBlock."""
+
+    name = "ProjectorBlock"
+
+    def __init__(
+        self,
+        bra: str,
+        ket: str,
```
jpmoutinho, line 467: Just a nitpick but I would reorder it as `ket, bra` so that `ProjectorBlock("00", "11")` means |00><11|.

qadence/operations.py
```diff
@@ -169,6 +170,28 @@ def dagger(self) -> Z:
...
 
+class Projector(ProjectorBlock):
+    """The projector operator."""
+
+    name = OpName.PROJ
+
+    def __init__(
+        self,
+        bra: str,
+        ket: str,
```
jpmoutinho, line 181: Same as above

jpmoutinho (commented):
Thanks @Roland-djee! Looks good so far, not much to say :)

qadence/blocks/primitive.py
Roland-djee: Done.

qadence/operations.py
Roland-djee: Done.

qadence/blocks/primitive.py
```diff
@@ -454,3 +454,34 @@ def _block_title(self) -> str:
...
+        bra: str,
+        qubit_support: int | tuple[int, ...],
+    ) -> None:
+        if isinstance(qubit_support, int):
+            qubit_support = (qubit_support,)
+        if len(bra) != len(ket):
+            raise ValueError(
+                "Bra and ket must be bitstrings of same length in the 'Projector' definition."
+            )
+        elif len(bra) != len(qubit_support):
```
Roland-djee, line 476: @jpmoutinho I am unsure about that check. Is it correct ? It doesn't feel like it.

qadence/operations.py
```diff
@@ -878,9 +899,9 @@ def __init__(self, control: int | tuple[int, ...], target1: int, target2: int) -
         if isinstance(control, tuple):
             control = control[0]
 
-        a00m = 0.5 * (Z(control) - I(control))
+        a00m = -N(control)
```
Roland-djee, line 902: @jpmoutinho Can we do better here by introducing more number operators ? It feels so but I haven't found a godd way so far.

qadence/blocks/primitive.py
jpmoutinho: It looks correct to me, what makes you say that? Since the `bra` is a bitstring corresponding to a basis state it must be of the same length as the number of qubits in the support to which the projector is applied, and the previous check already made sure it is of the same length as `ket`, so everything should be fine.

qadence/operations.py
jpmoutinho: You mean to simplify all the lines here? The `Z+I` is the `|0><0|` projector, but not sure it's worth adding more operators just to simplify that if they won't be used in other places. I like keeping the Pauli's explicit.

jpmoutinho (commented):
Thanks @Roland-djee. One thing to note is that @vytautas-a mentioned he had some issues with GPSR in his addressing pattern MR due to `N` not having a defined `generator` and `generator_eigenvalues`, whereas `(I-Z)/2` was fine because each of those separately do have them. Maybe @vytautas-a you can take a look here to see if you think any of these changes might lead to more of those issues?

qadence/operations.py
Roland-djee: What I meant was that since I integrated `N` in places, it feels here that it is halfway backed. I would replace with `N` wherever possible (say `a00m` and `a22` here) for consistency and also because it makes it cleaner I think ?

qadence/blocks/primitive.py
```diff
@@ -454,3 +454,34 @@ def _block_title(self) -> str:
...
+
+class ProjectorBlock(PrimitiveBlock):
+    """The abstract ProjectorBlock."""
+
+    name = "ProjectorBlock"
+
+    def __init__(
+        self,
+        ket: str,
+        bra: str,
```
madagra, line 467: Is this supposed to be given as a bitstring: e.g. "0010"?

qadence/circuit.py
```diff
@@ -7,8 +7,9 @@
 
 from sympy import Array, Basic
 
-from qadence.blocks import AbstractBlock, AnalogBlock, CompositeBlock, chain
+from qadence.blocks import AbstractBlock, AnalogBlock, CompositeBlock
 from qadence.blocks.utils import parameters, primitive_blocks
+from qadence.operations import chain
```
madagra, line 12: This import does not look correct. Before it was fine.

qadence/operations.py
```diff
@@ -169,13 +170,35 @@ def dagger(self) -> Z:
...
+
+class N(Projector):
     """The N = (1/2)(I-Z) operator."""
 
     name = OpName.N
 
-    def __init__(self, target: int):
-        super().__init__((target,))
+    def __init__(self, target: int, state: str = "1"):
+        super().__init__(ket=state, bra=state, qubit_support=(target,))
```
madagra, line 201: Wasn't more efficient to keep the previous implementation and leave the `ProjectorBlock` just for generic projectors?

qadence/states.py
```diff
@@ -8,10 +8,10 @@
 from torch import Tensor, concat
 from torch.distributions import Categorical, Distribution
 
-from qadence.blocks import ChainBlock, KronBlock, PrimitiveBlock, chain, kron
+from qadence.blocks import ChainBlock, KronBlock, PrimitiveBlock
 from qadence.circuit import QuantumCircuit
 from qadence.execution import run
-from qadence.operations import CNOT, RX, RY, RZ, H, I, X
+from qadence.operations import CNOT, RX, RY, RZ, H, I, X, chain, kron
```
madagra, line 14: Same as above, the import looked more correct before.

tests/qadence/test_operators.py
```diff
@@ -90,6 +94,405 @@ def cphase_eigenvals(p: float, n_qubits: int = 2) -> torch.Tensor:
...
+                            1.0 + 0.0j,
+                            0.0 + 0.0j,
+                        ],
+                    ]
+                ]
+            ),
+        ),
+    ],
+)
+def test_projector(projector: Projector, exp_projector_mat: Tensor) -> None:
```
madagra, line 490: Typing looks not correct, here.

madagra (commented):
A few comments but looks overall good to me.

qadence/blocks/primitive.py
Roland-djee: Yes, indeed. I can add a description in the docstring.

tests/qadence/test_operators.py
Roland-djee: Indeed, thanks.

qadence/blocks/primitive.py
Roland-djee: Done.

qadence/circuit.py
Roland-djee: Fixed.

qadence/states.py
Roland-djee: Fixed.

qadence/operations.py
Roland-djee: Well, it felt natural to express all possible operators using the projectors whenever possible ? I was thinking along Vincent's comment about helper functions for the number operator. I also tried to use `N` everywhere possible which makes for a clean interface I think ?

madagra: Yeah, actually thinking about it again, it makes sense. As long as it does not break something else. However, now it is not anymore the `N` operator, technically.

madagra (approved):
Looks good to me. Just check it does not generate issues with GPSR as mentioned.

Roland-djee:
> i guess this MR is dependent on [pasqal-io/pyqtorch#131](https://github.com/pasqal-io/pyqtorch/pull/131) @Roland-djee, should we wrap up that one first

Indeed. I was waiting for the PyQ one to be merged and a new release before moving forward this one.

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -127,7 +128,11 @@ def convert_block(
             # which would be wrong.
             return [pyq.QuantumCircuit(n_qubits, ops)]
     elif isinstance(block, tuple(non_unitary_gateset)):
-        return [getattr(pyq, block.name)(qubit_support[0])]
+        if isinstance(block, ProjectorBlock):
+            projector = getattr(pyq, block.name)
+            return [projector(ket=block.ket, bra=block.bra, target=qubit_support)]
```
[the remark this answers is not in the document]
Roland-djee: I changed that in PyQTorch. The argument is now `qubit_support` for consistency.

qadence/blocks/block_to_tensor.py
```diff
@@ -463,6 +466,14 @@ def _block_to_tensor_embedded(
         # add missing identities on unused qubits
         mat = _fill_identities(block_mat, block.qubit_support, qubit_support, endianness=endianness)
 
+    elif isinstance(block, ProjectorBlock):
```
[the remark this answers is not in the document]
Roland-djee: Is [this](https://github.com/pasqal-io/qadence/blob/f51bd0710486bdc9ef1302963f232539f76cea67/tests/qadence/test_operators.py#L494)  what you meant @dominikandreasseitz ? Or do you want me to explicitly add it to `test_matrices` ?

PR #212, opened 2023-11-22
[Fix] Validate state and parameter batch sizes

dominikandreasseitz:
This MR adds a validation step for user-provided batches of parameter values and states:
```
                "Batching of parameter values and states is only valid for the cases:\
                            (1) batch_size_values == batch_size_state\
                            (2) batch_size_values == 1 and batch_size_state > 1\
                            (3) batch_size_values > 1 and batch_size_state == 1."
```


nmheim (approved):
LGTM!

PR #218, opened 2023-11-25
[DRAFT] revamp control blocks

### LINKED ISSUE #215: Dagger/adjoint operation of qd.MCRX/qd.MCRY/qd.MCRZ functions
Hi Qadence team,

display(qd.MCRX([1,2,3],0,pi/2))

display(qd.MCRX([1,2,3],0,pi/2).dagger())
TypeError: MCRX.__init__() takes 4 positional arguments but 6 were given

Please also check
display(qd.MCRY([1,2,3],0,pi/2).dagger())
display(qd.MCRZ([1,2,3],0,pi/2).dagger())

vincentelfving:
While attempting to fix #215 , I came across several generalization opportunities, but also inconsistencies/troubles with the control and parametric control blocks.
I tried to solve this in the way presented in this PR, but I am not 100% confident this is the right direction to go. I am getting stuck in stuff like `from_dict`, and tests are not passing yet.

In any case, clearly a lot of multi-controlled and controlled-parametric blocks are not covered by the unit tests. Also, some gates are missing. So I am convinced some kind of revamp would be nice.

Another direction than the solution presented here, would be to frame controlled-unitary as `CompositeBlocks` from the start, rather than `PrimitiveBlocks`. Then we can handle any kind of control; however, that has cons too.

Thoughts/ideas?


dominikandreasseitz:
interesting idea @vincentelfving, i will have a look this week

PR #222, opened 2023-11-27
Run pipeline on windows and mac

awennersteen:
Let's add tests on windows and mac on staurdays?


dominikandreasseitz:
cool thanks!

.github/workflows/test_all.yml
```diff
@@ -1,35 +1,33 @@
...
-  test_qadence_ubuntu:
-    name: Test Qadence (ubuntu)
-    runs-on: ubuntu-latest
+  test_qadence_windows_mac:
+    name: Test Qadence (Windows, MacOS)
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        os: [macos-latest, windows-latest]
```
dominikandreasseitz, line 16: lets re-add ubuntu-latest and merge?

dominikandreasseitz (approved):

.github/workflows/test_all.yml
awennersteen: yeah. Notice I'm running the full tests on ubuntu but only the fast tests on mac and windows. But I think that should be ok

PR #234, opened 2023-11-30
[Feature] Analog feature maps

madagra:
Implemented the following:

- [x] analog feature map with a single global analog rotation
- [x] Rydberg feature map with addressing pattern
- [x] Rydberg tower-feature map with constant addressing pattern
- [x] revised the documentation of the semi-local addressing implementation to make only one file
- [x] refactor `feature_maps.py` to extract some reusable functions


qadence/constructors/rydberg_feature_maps.py
```diff
@@ -0,0 +1,90 @@
...
+    duration = 1000 * param / tower_detuning
+    return kron(
+        AnalogRot(
+            duration=duration,
+            delta=-tower_detuning * tower_coeffs[i],
+            phase=0.0,
+            qubit_support=(i,),
+        )
+        for i in range(n_qubits)
+    )
```
jpmoutinho, line 59: Note that currently on main this will implement all the pulses for these rotations in sequence for pulser and in parallel for PyQ. Once #176 is merged both backends do it in sequence.

I guess this could be written with the `AddressingPattern` but it would fix the pattern for the whole circuit, not just the feature map.

jpmoutinho (commented):
Hey @madagra I see it's just a draft, let me know when you want a more thorough review :)

qadence/constructors/feature_maps.py
```diff
@@ -163,8 +113,82 @@ def feature_map(
...
+    if fm_type == "tower":
+        logger.warning("Forcing reupload scaling strategy to TOWER")
+        reupload_scaling = ReuploadScaling.TOWER
+
+    basis_tag = fm_type.value if isinstance(fm_type, BasisSet) else str(fm_type)
+    rs_func, rs_tag = fm_reupload_scaling_fn(reupload_scaling)
+
     # Set overall multiplier
-    multiplier = 1 if multiplier is None else multiplier
+    multiplier = 1.0 if multiplier is None else Parameter(multiplier)
```
jpmoutinho, line 191: IIRC the `1` here was on purpose because `1.0` gets explicitly drawn in visualization and `1` doesn't.

qadence/constructors/rydberg_feature_maps.py
```diff
@@ -0,0 +1,100 @@
...
+    param = FeatureParameter(param)
+    duration = 1000 * param / tower_detuning
+    return kron(
+        AnalogRot(
+            duration=duration,
+            delta=-tower_detuning * tower_coeffs[i],
+            phase=0.0,
+            qubit_support=(i,),
+        )
+        for i in range(n_qubits)
```
jpmoutinho, line 59: Maybe add a `support` argument to the input?

qadence/constructors/rydberg_feature_maps.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def analog_feature_map(
+    param: str = "phi",
+    op: Callable[[Parameter | Basic], AnalogBlock] = AnalogRX,
+    fm_type: BasisSet | type[Function] | str = BasisSet.FOURIER,
+    feature_range: tuple[float, float] | None = None,
+    target_range: tuple[float, float] | None = None,
+    multiplier: Parameter | TParameter | None = None,
+) -> AnalogBlock:
```
jpmoutinho, line 79: One thing I did in the QCL example in the documentation that I seem to recall helped with training was to reupload data by applying a few different analog rotations in sequence with a different reupload scaling. This could be built into this function.

But anyway, this can be expanded on in a separate MR.

qadence/constructors/feature_maps.py
madagra: Got it, I will revert it.

qadence/constructors/rydberg_feature_maps.py
madagra: Good point.

madagra: Actually, adding `support` would break the implementation, I think. You need to add the analog (with interaction) rotation to each qubit separately given that each one will have its own local addressability weight.

PR #241, opened 2023-12-05
[Feature] Deserialize external models

jpmoutinho:


qadence/serialization.py
```diff
@@ -353,4 +355,4 @@ def load(file_path: str | Path, map_location: str = "cpu") -> SUPPORTED_TYPES:
         logger.debug(f"Successfully loaded {d} from {file_path}.")
     except Exception as e:
         logger.error(f"Unable to load Object from {file_path} due to {e}")
-    return deserialize(d)
+    return deserialize_fn(d)
```
dominikandreasseitz, line 358: why rename?

jpmoutinho: it's not a rename, it's a new argument in `load`

PR #243, opened 2023-12-05
[BugFix] Fix parameters arithmetic.

### LINKED ISSUE #242: Fix parameters arithmetic
While trying to parametrise an `AnalogRot` that error is raised:

```python
tests/qadence/test_mitigation.py:134: in <module>
    chain(AnalogRot(duration="t", omega="omega", delta="delta"), AnalogRZ(pi)),
qadence/operations.py:1196: in AnalogRot
    alpha = duration * sympy.sqrt(omega**2 + delta**2) / 1000  # type: ignore [operator]
E   TypeError: unsupported operand type(s) for ** or pow(): 'str' and 'int'
```

Roland-djee:
Closes #242

### DIFF
--- a/qadence/operations.py
+++ b/qadence/operations.py
@@ -1152,10 +1152,11 @@ def AnalogRot(
         ConstantAnalogRotation
     """
     q = _cast(QubitSupport, qubit_support)
-    if isinstance(duration, str):
-        duration = Parameter(duration)
-    alpha = duration * sympy.sqrt(Parameter(omega) ** 2 + Parameter(delta) ** 2) / 1000  # type: ignore [operator]
-
+    duration = Parameter(duration)
+    omega = Parameter(omega)
+    delta = Parameter(delta)
+    phase = Parameter(phase)
+    alpha = duration * sympy.sqrt(omega**2 + delta**2) / 1000
     ps = ParamMap(alpha=alpha, duration=duration, omega=omega, delta=delta, phase=phase)
     return ConstantAnalogRotation(parameters=ps, qubit_support=q)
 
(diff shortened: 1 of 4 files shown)


dominikandreasseitz (approved):
can we add a test for this atleast? i guess it went undiscovered because this line was never executed?

Roland-djee:
> can we add a test for this atleast? i guess it went undiscovered because this line was never executed?

Yes, it feels very much like untested functionality 😞

qadence/operations.py
```diff
@@ -1154,7 +1154,7 @@ def AnalogRot(
     q = _cast(QubitSupport, qubit_support)
     if isinstance(duration, str):
         duration = Parameter(duration)
-    alpha = duration * sympy.sqrt(omega**2 + delta**2) / 1000  # type: ignore [operator]
+    alpha = duration * sympy.sqrt(Parameter(omega) ** 2 + Parameter(delta) ** 2) / 1000  # type: ignore [operator]
```
jpmoutinho, line 1157: What happens here if `omega` and `delta` are already passed as `Parameter` ? Is calling `Parameter(Parameter(...))` ok?

I hadn't seen this MR earlier and also added this fix to [#244](https://github.com/pasqal-io/qadence/pull/244) but there I used the `if isinstance(..., str):` logic as done for the `duration` here.

dominikandreasseitz: @jpmoutinho its fine

jpmoutinho: Ok nice

nmheim (approved):

qadence/operations.py
Roland-djee: Nevertheless @jpmoutinho there is what we discussed in the way to test this.

jpmoutinho (approved):

PR #244, opened 2023-12-06
[Feature] Allow turning off the semi-local addressing pattern in pyqtorch backend

jpmoutinho:
Adds the possibility to turn off the addressing pattern for specific analog operations in the pyqtorch backend.

Currently just throws a warning in the pulser backend. The implementation there is a bit more tricky.

### DIFF
--- a/qadence/backends/pulser/pulses.py
+++ b/qadence/backends/pulser/pulses.py
@@ -20,12 +20,15 @@ from qadence.blocks.analog import (
     Interaction,
     WaitBlock,
 )
+from qadence.logger import get_logger
 from qadence.operations import RX, RY, RZ, AnalogEntanglement, OpName
 from qadence.parameters import evaluate
 
 from .channels import GLOBAL_CHANNEL, LOCAL_CHANNEL
 from .config import Configuration
 from .waveforms import SquareWaveform
+
+logger = get_logger(__file__)
 
 TVar = Union[Variable, VariableItem]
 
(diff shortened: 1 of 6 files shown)


qadence/backends/pulser/pulses.py
```diff
@@ -118,6 +121,11 @@ def add_pulses(
 
     # TODO: lets move those to `@singledipatch`ed functions
     if isinstance(block, WaitBlock):
+        if not block.add_pattern:
+            logger.warning(
+                "Found block with `add_pattern = False`. This is not yet supported in the Pulser "
+                "backend. If an addressing pattern is specified, it will be added to all blocks."
+            )
```
madagra, line 128: I thought the implementation would have not been so tricky given that the patterns are in a completely separate channel (the DMM). What is the main hurdle here @vytautas-a?

qadence/operations.py
```diff
@@ -1154,16 +1156,23 @@ def AnalogRot(
     q = _cast(QubitSupport, qubit_support)
     if isinstance(duration, str):
         duration = Parameter(duration)
-    alpha = duration * sympy.sqrt(omega**2 + delta**2) / 1000  # type: ignore [operator]
+    if isinstance(omega, str):
+        omega = Parameter(omega)
+    if isinstance(delta, str):
+        delta = Parameter(delta)
+    if isinstance(phase, str):
+        phase = Parameter(phase)
```
madagra, line 1164: Thanks, this was a bug?

madagra (approved):
Thanks, @jpmoutinho. It looks good to me.

qadence/backends/pulser/pulses.py
jpmoutinho: If I understood the code correctly, currently the addressing pattern is a separate function that acts on the fully built sequence and adds the pattern pulses for the full duration of the sequence. To implement this there needs to be some refactoring so that the pattern is added when the actual pulse for each block is built, such that we can check if `block.add_pattern = True` and then add the pattern pulse for the duration of that block.

qadence/operations.py
madagra: Great, thanks.

qadence/backends/pulser/pulses.py
madagra: I see. I will check the code a bit more in detail.

PR #251, opened 2023-12-11
[Fix] Add mitigation tolerance

dominikandreasseitz:


Roland-djee (approved):
Thanks.

dominikandreasseitz:
merging this since its breaking tests, @Roland-djee  @rajaiitp we can do another MR with proper tolerances after you investigated the paper

rajaiitp (commented):
The paper does not make claims about producing optimality for some reason. Did a couple of test for norm error (row 1), js_error(row 2), runtime(row 3) in the case of mle(col 1) and optimization (col 2) over 1000 iterations for 6 qubits. Following are the results,
                           MLE                             OPT 
2 norm  error   60.1152783062038     29.70477756745552
js error             39.4015272580539     6.603815432167311
runtime           0.023070573806762695      25.450140476226807

The runtime for mle is 1000 times faster with an average error of ~ 0.03 for js and 2-norm. With a middle acceptance of 0.05 over 5 runs should help with preventing the test cases from failing

PR #260, opened 2023-12-12
[Bugfix] Fix cron syntax for nightly tests.

Roland-djee:


dominikandreasseitz:
how do we test if is actually does what we want?

dominikandreasseitz:
thanks for that!

Roland-djee:
> how do we test if is actually does what we want?

We'll see tomorrow I guess. I also think that there is redundant dependabot files. The one in `.github` should be deleted I think.

.github/workflows/test_all.yml
```diff
@@ -2,8 +2,8 @@ name: Run all Qadence tests on schedule.
 
 on:
   schedule:
-    # 03:00 every Saturday morning
-    - cron: '0 3 * * 6'
+  # 03:00 nightly
+  - cron: "0 3 * * *"
```
nmheim, line 6: haha, damn, was it the `'`-ticks really?

Roland-djee:
Right, as per this [example](https://www.tinybird.co/docs/guides/scheduling-with-github-actions-and-cron.html#:~:text=Cron%20syntax%20explanation&text=These%20would%20be%20typical%20cron,%5C*%20%5C*%201%20%5C*%20%5C*),  I think there was a `scheduled` key missing to make it all work.

dominikandreasseitz (approved):

PR #269, opened 2023-12-13
[Docs] Add projector docs

Roland-djee:
Add missing doc section for the projector blocks.

### DIFF
--- a/docs/advanced_tutorials/projectors.md
+++ b/docs/advanced_tutorials/projectors.md
@@ -13,9 +13,10 @@ projector_matrix = block_to_tensor(projector_block)
 print(f"projector matrix = {projector_matrix}") # markdown-exec: hide
 ```
 
-Other standard operations are expressed as projectors in Qadence. For instance, the number operator $N=\dfrac{1}{2}(I-Z)=|1\rangle\langle 1|$ is used to build $\textrm{CNOT}(i,j)=N(i)\otimes(X(j)-I(j))$. In fact, `CNOT` can also be defined as a projector controlled-unitary operation: $\textrm{CNOT}(i,j)=|0\rangle\langle 0|\otimes \mathbb{I}+|1\rangle\langle 1|\otimes X$ and we can check their matrix representations are identical:
+Other standard operations are expressed as projectors in Qadence. For instance, the number operator $N=\dfrac{1}{2}(I-Z)=|1\rangle\langle 1|$ is used to build $\textrm{CNOT}(i,j)=N(i)\otimes(X(j)-I(j))$. In fact, `CNOT` can also be defined as a projector controlled-unitary operation: $\textrm{CNOT}(i,j)=|0\rangle\langle 0|_i\otimes \mathbb{I}(j)+|1\rangle\langle 1|_i\otimes X(j)$ and we can check their matrix representations are identical:
 
 ```python exec="on" source="material-block" session="projector" result="json"
+from qadence.blocks import block_to_tensor
 from qadence import kron, I, X, CNOT
 
 # Define a projector for |0> onto the qubit labelled 0.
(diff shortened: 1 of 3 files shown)


docs/advanced_tutorials/projectors.md
```diff
@@ -0,0 +1,20 @@
+This section introduces the `ProjectorBlock` as an implementation for the the quantum mechanical projection operation. It evaluates to the outer product of a ket and a bra expressed as bitstrings.
```
madagra, line 1: Can you add the mathematical formula and a little bit more explanation of the interface?

Roland-djee: Done.

docs/advanced_tutorials/projectors.md
```diff
@@ -0,0 +1,34 @@
...
+# Define a projector for |1> onto the qubit labelled 0.
+projector_block = Projector(ket="1", bra="1", qubit_support=0)
+
+# As any block, the matrix representation can be retrieved.
+projector_matrix = block_to_tensor(projector_block)
+
+print(f"projector matrix = {projector_matrix}") # markdown-exec: hide
+```
+
+Other standard operations are expressed as projectors in Qadence. For instance, the number operator $N=\dfrac{1}{2}(I-Z)=|1\rangle\langle 1|$ is used to build projector controlled-unitary gates: $\textrm{CNOT}(i,j)=N(i)\otimes(X(j)-I(j))$.
```
jpmoutinho, line 16: Hm do you mean to write here about the generator of controlled-operations? I think that's the only thing you can write with the projector, but you would still need the exponential to make the unitary gate [(e.g. as done here)](https://pasqal-io.github.io/qadence/latest/digital_analog_qc/daqc-cnot/).

jpmoutinho: But I wouldn't necessary include this since we don't really express the controlled unitaries we have as the exponential of the projectors, we mostly just use their direct matrix definitions.

docs/advanced_tutorials/projectors.md
```diff
@@ -0,0 +1,34 @@
...
+
+# Define a projector for |1> onto the qubit labelled 1.
+projector_block = Projector(ket="1", bra="1", qubit_support=1)
+
+# Some non unitary computation.
+non_unitary_block = chain(H(0), CNOT(0,1), projector_block)
+
+non_unitary_wf = run(non_unitary_block)  # Run on PyQTorch.
+
+print(f"non_unitary_wf = {non_unitary_wf}") # markdown-exec: hide
```
jpmoutinho, line 33: I would suggest `projected_wf` since `unitary` most often refers to the operators (although it does indeed become a "non-unit" vector).

jpmoutinho (commented):
Thanks @Roland-djee. I think I would still include one more example for a multi-qubit projector, maybe something like defining a `SWAP(i, j)` gate as `Projector("01", "10", qubit_support = (i, j)) + Projector("10", "01", qubit_support = (i, j)`

Hmm I guess we could later add an option to the projectors to simplify the above by automatically adding the hermitian conjugate, mimicking the typical $+\text{ h.c.}$ notation found in Physics papers. Maybe `Projector(..., add_hc = True)`. Wdyt?

docs/advanced_tutorials/projectors.md
Roland-djee: Actually, what I really meant is this:

Screenshot 2023-12-18 at 16 03 07

But this isn't what we really do to define `CNOT`. Maybe a good example is to construct the `CNOT` using projectors and compare the unitaries with our own implementation ?

jpmoutinho: Ah right! Yes sounds good.

docs/advanced_tutorials/projectors.md
Roland-djee: Done.

docs/advanced_tutorials/projectors.md
Roland-djee: Done.

docs/advanced_tutorials/projectors.md
```diff
@@ -0,0 +1,85 @@
...
+# Define a projector for |1> onto the qubit labelled 0.
+projector_block = Projector(ket="1", bra="1", qubit_support=0)
+
+# As any block, the matrix representation can be retrieved.
+projector_matrix = block_to_tensor(projector_block)
+
+print(f"projector matrix = {projector_matrix}") # markdown-exec: hide
+```
+
+Other standard operations are expressed as projectors in Qadence. For instance, the number operator $N=\dfrac{1}{2}(I-Z)=|1\rangle\langle 1|$ is used to build $\textrm{CNOT}(i,j)=N(i)\otimes(X(j)-I(j))$. In fact, `CNOT` can also be defined as a projector controlled-unitary operation: $\textrm{CNOT}(i,j)=|0\rangle\langle 0|_i\otimes \mathbb{I}(j)+|1\rangle\langle 1|_i\otimes X(j)$ and we can check their matrix representations are identical:
```
jpmoutinho, line 16: I think this equality is not correct: $\textrm{CNOT}(i,j)=N(i)\otimes(X(j)-I(j))$.

```suggestion
Other standard operations are expressed as projectors in Qadence. For instance, the number operator is the projector onto the 1-subspace, $N=|1\rangle\langle 1|$.

In fact, projectors can be used to compose arbitrary operators. For example, the `CNOT` can also be defined as $\textrm{CNOT}(i,j)=|0\rangle\langle 0|_i\otimes \mathbb{I}(j)+|1\rangle\langle 1|_i\otimes X(j)$ and we can check their matrix representations are identical:
```

docs/advanced_tutorials/projectors.md
```diff
@@ -0,0 +1,85 @@
...
+# Define a projector for |1> onto the qubit labelled 0.
+projector1 = Projector(ket="1", bra="1", qubit_support=0)
+
+# Construct the projector controlled CNOT.
+projector_cnot = kron(projector0, I(1)) + kron(projector1, X(1))
+
+# Get the underlying unitary.
+projector_cnot_matrix = block_to_tensor(projector_cnot)
+
+# Qadence CNOT unitary defined as N @ (X-I).
```
jpmoutinho, line 34:
```suggestion
# Qadence CNOT unitary.
```

docs/advanced_tutorials/projectors.md
```diff
@@ -0,0 +1,85 @@
...
+qadence_cnot_matrix = block_to_tensor(CNOT(0,1))
+
+print(f"projector cnot matrix = {projector_cnot_matrix}") # markdown-exec: hide
+print(f"qadence cnot matrix = {qadence_cnot_matrix}") # markdown-exec: hide
+```
+
+Another example is the canonical SWAP gate that can be defined as $SWAP(i,j)=|00\rangle\langle 00|+|01\rangle\langle 10|+|10\rangle\langle 01|+|11\rangle\langle 11|$. Indeed, it can be shown that their matricial representations are again identical:
```
jpmoutinho, line 41: Just to contextualize: my previous suggested change to the CNOT was to emphasize that you really can write anything you want in terms of projectors. In the CNOT it was by composing with paulis, here in the SWAP it is by composing projectors alone. 

In the limit you can always write some arbitrary matrix $M = [[a, b], [c, d]]$ as $M = a|0><0| + b|0><1|+ c|1><0|+ d|1><1|$ (and the same for higher dimension).

jpmoutinho (approved):
Thanks @Roland-djee, a few final suggestions, but approving. The only thing that really needs changing I think is the $\textrm{CNOT}(i,j)=N(i)\otimes(X(j)-I(j))$.

docs/advanced_tutorials/projectors.md
Roland-djee: Sure. But I couldn't get a better formulation for the SWAP unitary unfortunately.

Issue #272, opened 2023-12-18
[Feature] Add `param_prefix` to Identity Initialized Ansatz to optionally specify a prefix for parameter names.


smitchaudhary:
I think there is a need to add `param_prefix` to the [Identity Initialized Ansatz](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/iia.py#L64). Same as in [hea here](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/ansatze.py#L16), I would add it to all the other names that we already have, namely `alpha`, `beta`, `gamma`, (for the 3 rotation blocks) and `theta_ent` for parametric entanglers. For the following reasons:

1. Currently there is no way to provide this prefix. Someone could easily name their Parameter alpha and end up with two different parameters having same name, and thus, in essence being the same parameter (which is not the desired behaviour).
2. If someone wants to have two IIA blocks and calls them separately, they will end up being the same parameter, even if we want them to be different.

Currently, there is no way to prepend to/change the name of these parameters.

Proposal:

Just add an optional `param_prefix` that users can set? Potentially defaults to an empty string? Thus maintaining the existing desired structure with `alpha`, `beta`, and `gamma`, without breaking anything?

Also open to suggestions about some other nomenclature if it is deemed more appropriate.

PR #273, opened 2023-12-18
[Feature] Added param_prefix to IIA

smitchaudhary:
Relates to #272 Added `param_prefix` to Identity initialized ansatz. It defaults to an empty string.

### DIFF
--- a/qadence/constructors/iia.py
+++ b/qadence/constructors/iia.py
@@ -64,6 +64,7 @@ def _rotations(
 def identity_initialized_ansatz(
     n_qubits: int,
     depth: int = 1,
+    param_prefix: str = "",
     rotations: Any = [RX, RY],
     entangler: Any = CNOT,
     periodic: bool = False,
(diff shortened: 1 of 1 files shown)


smitchaudhary:
One consequence of having an empty string as the prefix is that in cases where it defaults to the empty string, the parameter names would start with an underscore `_`.

On the other hand, keeping some other prefix, like `"iia"` or `"phi"` is that the names would become unnecessarily too long? Such as `phi_alpha_01` or `iia_theta_ent_01`.

Which of the two is the better alternative?

jpmoutinho (commented):
Thanks @smitchaudhary!

Personally I would prefer that the global `param_prefix` defaults to something because I don't like the default starting with "_". You can replace the other param prefixes by the `α`, `β`, `γ`, `θ`, it should work fine.

smitchaudhary:
Thanks @jpmoutinho 

I added `"iia"` as the default string, and taking your suggestion, included the respective Greek characters to shorten the name.

jpmoutinho (approved):
Nice, thanks :)

PR #276, opened 2023-12-19
[Bugfix] Fix projector matrices

Roland-djee:
Projector matrices were computed without filling with identities. Tested with CNOT and SWAP as projector compositions.


tests/qadence/test_matrices.py
```diff
@@ -470,6 +470,36 @@ def test_projector_tensor(projector: AbstractBlock, exp_projector_mat: Tensor) -
     assert torch.allclose(projector_mat, exp_projector_mat, atol=1.0e-4)
```
jpmoutinho, line 470: I guess this `atol` could be as low as `ATOL_64` no?

tests/qadence/test_matrices.py
```diff
@@ -470,6 +470,36 @@ def test_projector_tensor(projector: AbstractBlock, exp_projector_mat: Tensor) -
...
+            cnot,
+            CNOT(0, 1),
+        ),
+        (swap, SWAP(0, 1)),
+    ],
+)
+def test_projector_composition_unitaries(
+    projector: AbstractBlock, exp_projector: AbstractBlock
+) -> None:
+    assert torch.allclose(block_to_tensor(projector), block_to_tensor(exp_projector))
```
jpmoutinho, line 500: I would also specify the `atol` here.

jpmoutinho (approved):

tests/qadence/test_matrices.py
Roland-djee: Here, it is possible. Done.

tests/qadence/test_matrices.py
Roland-djee: Nope, this is the best I could achieve for this test.

PR #277, opened 2023-12-19
[Docs] Add docs measurements

### LINKED ISSUE #146: Better document usage of shot-based protocols
Currently, there is almost no documentation on how to use tomography and classical shadows protocols.

Roland-djee:
Closes #146.


dominikandreasseitz (approved):
thanks for that!

PR #284, opened 2024-01-02
[Refact] Replace alias wait with AnalogInteraction

jpmoutinho:
I suggest to replace the `wait` alias operation with `AnalogInteraction` to be a bit more formal and explicit about what it does (it evolves the interaction term in the Hamiltonian for a register of interacting qubits).

### DIFF
--- a/docs/digital_analog_qc/analog-basics.md
+++ b/docs/digital_analog_qc/analog-basics.md
@@ -268,16 +268,16 @@ print("States equivalent: ", bool_equiv)
 
 Finally, besides applying specific qubit rotations, we can also choose to evolve only the interaction term
 $\mathcal{H}^\text{int}$, equivalent to setting $\Omega = \delta = \phi = 0$. To do so, Qadence provides the
-function `wait` which does exactly this.
+function `AnalogInteraction` which does exactly this.
 
 ```python exec="on" source="material-block" result="json" session="int"
-from qadence import Register, BackendName, random_state, equivalent_state, wait, run
+from qadence import Register, BackendName, random_state, equivalent_state, AnalogInteraction, run
 
 n_qubits = 3
 reg = Register.line(n_qubits, spacing=8.0)
 
 duration = 1000.
-op = wait(duration = duration)
+op = AnalogInteraction(duration = duration)
 
 init_state = random_state(n_qubits)
 
@@ -376,7 +376,7 @@ of the emulated analog interface, and in the next tutorial on Quantum Circuit Le
 in a simple QML example. Here we specify some extra details of this interface.
 
 In the block system, all the Analog rotation operators initialize a [`ConstantAnalogRotation`][qadence.blocks.analog.ConstantAnalogRotation]
-block, while the `wait` operation initializes a [`WaitBlock`][qadence.blocks.analog.WaitBlock]. As we have shown, by default,
+block, while the `AnalogInteraction` operation initializes a [`WaitBlock`][qadence.blocks.analog.WaitBlock]. As we have shown, by default,
 these blocks use a global qubit support, which can be passed explicitly by setting `qubit_support = "global"`. However, the blocks do support
 local qubit supports, with some constraints. The main constraint is that using `kron` on operators with different durations is not allowed.
 
--- a/docs/digital_analog_qc/analog-basics.md
+++ b/docs/digital_analog_qc/analog-basics.md
@@ -371,10 +371,10 @@ In the previous section we have exemplified the main ingredients of the current 
 of the emulated analog interface, and in the next tutorial on Quantum Circuit Learning we will exmplify its usage
 in a simple QML example. Here we specify some extra details of this interface.
 
-In the block system, all the Analog rotation operators initialize a [`ConstantAnalogRotation`][qadence.blocks.analog.ConstantAnalogRotation]
-block, while the `AnalogInteraction` operation initializes a [`InteractionBlock`][qadence.blocks.analog.InteractionBlock]. As we have shown, by default,
-these blocks use a global qubit support, which can be passed explicitly by setting `qubit_support = "global"`. However, the blocks do support
-local qubit supports, with some constraints. The main constraint is that using `kron` on operators with different durations is not allowed.
+In the block system, all analog rotation operators initialize a [`ConstantAnalogRotation`][qadence.blocks.analog.ConstantAnalogRotation]
+block, while the `AnalogInteraction` operation initializes an [`InteractionBlock`][qadence.blocks.analog.InteractionBlock]. As we have shown, by default,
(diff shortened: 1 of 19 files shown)


vytautas-a:
Looks good to me - definitely more transparent.

vytautas-a (approved):

docs/digital_analog_qc/analog-basics.md
```diff
@@ -266,16 +266,16 @@ print("States equivalent: ", bool_equiv)
 
 Finally, besides applying specific qubit rotations, we can also choose to evolve only the interaction term
 $\mathcal{H}^\text{int}$, equivalent to setting $\Omega = \delta = \phi = 0$. To do so, Qadence provides the
-function `wait` which does exactly this.
+function `AnalogInteraction` which does exactly this.
```
awennersteen, line 269: `wait` was a funciton but `AnalogInteraction` is not.

qadence/operations.py
```diff
@@ -1124,6 +1112,30 @@ def wait(
     return WaitBlock(parameters=ps, qubit_support=q, add_pattern=add_pattern)
```
awennersteen, line 1112: to me it's strange to maintain the `WaitBlock`

qadence/operations.py
```diff
@@ -1085,34 +1086,21 @@ def eigenvalues(self) -> Tensor:
...
-    def duration(self) -> Basic:
-        return self.parameters.duration
-
-
 def _cast(T: Any, val: Any) -> Any:
     return val if isinstance(val, T) else T(val)
 
 
-def wait(
+def AnalogInteraction(
```
awennersteen, line 1093: Usually `CamelCase` means that we're dealing with a class. If you want to keep this a function I'd go with `analog_interaction`

qadence/operations.py
awennersteen, line 1112: It's strange to me that we have `WaitBlock` here, but otherwise renamed it

jpmoutinho: Yes there are quite a few things still to be worked on in this analog part of the block system imo, here I just wanted to deprecate the usage of `wait` in favor of something more explicit and deal with the rest later.

awennersteen: But could we not at least also rename that object?
With this change we'll have
```
>>> print(AnalogInteraction(10))
WaitBlock(t=10.0, support=(<QubitSupportType.GLOBAL: 'global'>,))
```
Moreover, if you print the diagram of the circutit it will also be named "Wait"

awennersteen: If theres a decent reason not to, feel free to merge, but let's not postpone much finalizing it?

jpmoutinho: Yes indeed I think it can be renamed

docs/digital_analog_qc/analog-basics.md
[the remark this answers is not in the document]
jpmoutinho: Yes I understand the convention, but the other aliases like `AnalogRX` (which are also functions) were already camel case, which is why I also did that here. Although they are functions, they are used like operation blocks, which for everything else in Qadence we have them as classes. I also had in mind to look into simplifying this analog blocks at some point, in which case I would probably just define them as classes directly, and thus keeping the camel case right now would make that transition easier.

Regarding the `evolve` function, that is already the interface when doing `HamEvo(Hint, t)`, so I don't think we need a separate function for that.

docs/digital_analog_qc/pulser-basic.md
```diff
@@ -53,10 +53,10 @@ Currently, the Pulser backend supports the following operations:
 
 | gate        | description                                                                                      | trainable parameter |
 |-------------|--------------------------------------------------------------------------------------------------|---------------------|
-| `RX`, `RY`     | Single qubit rotations. Notice that the interaction is on and this affects the resulting gate fidelity.                                                                        | rotation angle      |
-| `AnalogRX`, `AnalogRY`, `AnalogRZ` | Span a single qubit rotation among the entire register.                                          | rotation angle      |
+| `RX`, `RY`     | Single qubit rotations. Notice that the interaction is on and this affects the resulting gate fidelity.          | rotation angle      |
+| `AnalogRX`, `AnalogRY`, `AnalogRZ` | Span a single qubit rotation among the entire register.                                      | rotation angle      |
 | `entangle`  | Fully entangle the register.                                                                     | interaction time    |
```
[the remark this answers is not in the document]
jpmoutinho: Yes, the `entangle` still needs to be disentangled in another MR 😅 I created an issue here: https://github.com/pasqal-io/qadence/issues/290

qadence/operations.py
```diff
@@ -1085,43 +1086,54 @@ def eigenvalues(self) -> Tensor:
...
-    return WaitBlock(parameters=ps, qubit_support=q, add_pattern=add_pattern)
+    return InteractionBlock(parameters=ps, qubit_support=q, add_pattern=add_pattern)
+
+
+def wait(
+    duration: TNumber | sympy.Basic,
+    qubit_support: str | QubitSupport | tuple = "global",
+    add_pattern: bool = True,
+) -> InteractionBlock:
+    logger.warning("The alias `wait` is deprecated, please use `AnalogInteraction`")
```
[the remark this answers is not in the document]
jpmoutinho: Yes, should it be 1.3 or is that too soon for this type of thing?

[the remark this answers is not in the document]
Roland-djee: Major version releases are a good time to remove deprecated methods. Minor releases should typically not contain breaking changes I think.

docs/digital_analog_qc/analog-basics.md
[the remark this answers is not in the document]
Roland-djee: Alright. Did you open an issue for the analog interface simplification ? I have few suggestions.

qadence/blocks/analog.py
```diff
@@ -99,9 +99,9 @@ def dagger(self) -> AbstractBlock:
 
 
 @dataclass(eq=False, repr=False)
-class WaitBlock(AnalogBlock):
+class InteractionBlock(AnalogBlock):
     """
-    Waits.
+    Free-evolution for the interaction term of a register of qubits.
```
Roland-djee, line 104:
```suggestion
    Free-evolution for the Hamiltonian interaction term of a register of qubits.
```

Roland-djee (approved):
LGTM @jpmoutinho. However, as you sugegsted, we should consider a simplification of the analog interface. It feels like there are some abstraction layers that could be more directly used and reused.

docs/digital_analog_qc/analog-basics.md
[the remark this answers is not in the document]
jpmoutinho: There was one in a bit of a different direction, so I opened this one: https://github.com/pasqal-io/qadence/issues/291

PR #285, opened 2024-01-05
[Feature] Hybrid Adjoint/AD higher order

dominikandreasseitz:


dominikandreasseitz:
can be reopened once proper theory has been derived

PR #287, opened 2024-01-09
[BugFix] DA-QFT for larger number of qubits in the analog block

jpmoutinho:


dominikandreasseitz (approved):
dont have context for this so i cant judge what it fixes exactly

jpmoutinho:
> dont have context for this so i cant judge what it fixes exactly

Sorry, it's an edge case of using a QFT of a smaller size inside a circuit with larger number of qubits. It was only tested for both having same size. Thanks!

Roland-djee:
> > dont have context for this so i cant judge what it fixes exactly
> 
> Sorry, it's an edge case of using a QFT of a smaller size inside a circuit with larger number of qubits. It was only tested for both having same size. Thanks!

Yes, I guess it makes it easier to review when a minimal description is provided. If that's real minor, maybe no need to open an full-fledge issue (though I'm inclined to think this is the way to go) but the least should be to state what it is supposed to fix.

qadence/constructors/qft.py
```diff
@@ -175,7 +176,7 @@ def _qft_layer_sDAQC(
     support: tuple[int, ...],
     layer: int,
     inverse: bool,
-    gen_build: AbstractBlock,
+    gen_build: AbstractBlock | None,
 ) -> AbstractBlock:
     """
     QFT Layer using the sDAQC technique following the paper:
```
Roland-djee, line 182:
```suggestion
    QFT Layer using the sDAQC technique. 
    
    Following the paper:
```

Roland-djee (approved):

PR #289, opened 2024-01-09
[Feature] Added SDAQC strategy for IIA

### LINKED ISSUE #288: [Feature] Add sDAQC strategy to Identity Initialised Ansatz
The [Hardware Efficient Ansatz](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/ansatze.py#L16) already supports a fully digital as well as [sDAQC](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/ansatze.py#L249) strategy. The [identity initialized ansatz](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/iia.py#L65) is automatically digital. We can also have an IIA with sDAQC. It just requires implementing an analog entangler (similar to the hea case), and initialize the trainable time parameter to zero to keep it identity initialized.

smitchaudhary:
Closes #288
Added SDAQC strategy to IIA just like how it is for HEA.

Introduced changes:

- Added an argument `strategy` in the function to get identity initialized ansatz. It defaults to fully digital, thus not breaking any existing code.
- Changed the default value of the `entangler` to `None` from `CNOT` to take into account the digital-analog paradigm. For digital, it still defaults to `CNOT` thus not breaking any existing code. Defaults to NN interaction for digital-analog.
- Added CPHASE to the list of possible 2 qubit digital entangling gates.
- Some minor docstring changes.
- Added tests for the IIA with SDAQC strategy.

### DIFF
--- a/qadence/constructors/iia.py
+++ b/qadence/constructors/iia.py
@@ -4,12 +4,13 @@ from typing import Any, Type, Union
 
 import torch
 
-from qadence.blocks import AbstractBlock, KronBlock, chain, kron, tag
-from qadence.operations import CNOT, CRX, CRY, CRZ, CZ, RX, RY
+from qadence.blocks import AbstractBlock, KronBlock, block_is_qubit_hamiltonian, chain, kron, tag
+from qadence.constructors.hamiltonians import hamiltonian_factory
+from qadence.operations import CNOT, CPHASE, CRX, CRY, CRZ, CZ, RX, RY, HamEvo
 from qadence.parameters import Parameter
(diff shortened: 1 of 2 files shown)


jpmoutinho (changes requested):
Thanks @smitchaudhary! Looks good in terms of the logic for adding the entangler.

Could you add a small test for it? Replicating the `test_hea_sDAQC` in `tests/constructors/test_ansatz.py` should be sufficient.

jpmoutinho (approved):
Thanks 👍

PR #293, opened 2024-01-12
[Feature] Allow custom interactions in the hamiltonian factory

jpmoutinho:


tests/constructors/test_hamiltonians.py
```diff
@@ -1,28 +1,28 @@
...
 
-from qadence import (
-    Interaction,
-    N,
-    Register,
-    X,
-    Y,
-    Z,
-    hamiltonian_factory,
-)
```
awennersteen, line 15: I guess this was an automatic action?

awennersteen (approved):
Thanks!

tests/constructors/test_hamiltonians.py
jpmoutinho: Yes some re-formatting in the pre-commit hooks

docs/tutorials/hamiltonians.md
```diff
@@ -19,6 +19,19 @@ hamilt = hamiltonian_factory(n_qubits, interaction=Interaction.ZZ)
 print(hamilt) # markdown-exec: hide
 ```
 
+Alternatively, a custom interaction function can also be defined. The input should be two integer indices $i$ and $j$ and it should return a composition of pauli terms representing the interaction between qubits $i$ and $j$:
+
+```python exec="on" source="material-block" result="json" session="hamiltonians"
+def custom_int(i: int, j: int):
```
Roland-djee, line 25: Would it make sense to pass a pair object ? Ie. a couple of ints ? I guess this is what you get out of the register when you extract the complete graph (which sounds to be the most common use case but I might be wrong there).

Roland-djee (approved):
Thanks @jpmoutinho. LGTM.

docs/tutorials/hamiltonians.md
jpmoutinho: The signature here follows the other interaction functions that are pre-defined for use in the `hamiltonian_factory`. To change it here I would also have to change those (which I could do).

Either way, the current signature can be easily used with the information the register provides like so:
```
from qadence import *

reg = Register.line(5)

def custom_int(i: int, j: int):
    return X(i) @ X(j) + Y(i) @ Y(j)

add(custom_int(*p) for p in reg.all_node_pairs)
```

Roland-djee: Also doable indeed.

awennersteen:
Let's merge this?

qadence/constructors/hamiltonians.py
```diff
@@ -119,9 +119,16 @@ def hamiltonian_factory(
 
     # Get interaction function
     if interaction is not None:
-        int_fn = INTERACTION_DICT.get(interaction, None)
-        if int_fn is None:
-            raise KeyError(f"Interaction {interaction} not supported.")
+        if callable(interaction):
+            int_fn = interaction
+            try:
+                interaction(0, 1)
```
Roland-djee, line 125: I am no sure it makes for a solid test because it won't pick up if the function signature is type annotated with floats for instance. I would rather use `inspect` or something like [this](https://stackoverflow.com/questions/73206836/how-to-verify-the-signature-of-python-method-when-using-default-parameters).

qadence/constructors/hamiltonians.py
```diff
@@ -119,9 +119,16 @@ def hamiltonian_factory(
...
+            except TypeError:
+                raise TypeError("Please write your custom interaction with two integer inputs.")
```
Roland-djee, line 127:
```suggestion
                raise TypeError("Please provide a custom interaction function signed with  two integer parameters.")
```

qadence/constructors/hamiltonians.py
Roland-djee: Well, my point being that if you want to _at least_ check that the signature is two integers then it makes sense to use `inspect`. Otherwise, there is no other way to ensure validity of the function imho.

awennersteen: I guess Roland has  valid concern. However, we are writing Python code. And that would not be Pythonic.

Viewing Qadence as a DSL, it would in any case raise a "compile-time" error, since it would fail immediately. 

I'm not sure if it's a good idea to plan to introduce that level of type-safety in Python, since it goes against the way Python is intended to be used... Swimming upstream...

Roland-djee: I understand that this is a PITA and very likely an overkill. My point is, the only meaningful check is signature types. Otherwise, I think we can ignore it.

jpmoutinho: Hey @Roland-djee I did try to use `inspect.signature` but it would amount to parsing a string like `"i: int"` to check if the second part was "int" or something else. It seemed a bit of a weird usage to me and maybe not very meaningful since it doesn't control what the function returns.

Then I tried adding a specific control of the return to check it is either Pauli terms or Projectors, but ended up running into some failing tests for Hamiltonians built on Projectors. Since I don't want to delay this, I just dropped the support for Projectors for now and made it so only Pauli operators are allowed, which either way will be most of the usage for now, and IMO the control now is the most robust.

I opened an issue here for adding the Projectors later: https://github.com/pasqal-io/qadence/issues/300

Roland-djee: OK fair enough. My main concern was primarily from the fact that a user can actually plug in a function and break everything. Nevertheless, I also understand that: 1) it's the user's responsibility to do things sensibly, especially if informed 2) it's Python and we can't do much about safety in Python.

qadence/constructors/hamiltonians.py
```diff
@@ -119,9 +119,19 @@ def hamiltonian_factory(
 
     # Get interaction function
     if interaction is not None:
-        int_fn = INTERACTION_DICT.get(interaction, None)
-        if int_fn is None:
-            raise KeyError(f"Interaction {interaction} not supported.")
+        if callable(interaction):
+            int_fn = interaction
+            try:
```
Roland-djee, line 124: Is the try/catch block still needed ? It feels you can get away with it now.

Roland-djee (approved):
Thanks @jpmoutinho.

qadence/constructors/hamiltonians.py
jpmoutinho: It's still relevant in case the call to `interaction(0,1)` fails due to an incorrect signature.

PR #295, opened 2024-01-17
[Refac Fix] Use backend directly in measurements

dominikandreasseitz:
Fixes https://github.com/pasqal-io/qadence/issues/296
before, using 'AD' in measurements was not possible since protocols default to diff_mode 'GPSR' and hence use gate-lvl params. we are now using only one backend which allows for flexible param-type usage


Roland-djee (approved):
Thanks @dominikandreasseitz. Just one remark, there is a global DIFF_MODE variable in the test file, it is worth parametrising tests with it ? Then, all test cases can be checked in both diff modes (for backends that support them of course).

PR #304, opened 2024-01-25
[Serialization] Log incoming dict

dominikandreasseitz:


qadence/serialization.py
```diff
@@ -193,6 +193,7 @@ def deserialize(d: dict, as_torch: bool = False) -> SUPPORTED_TYPES:
     ```
     """
     obj: Any
+    logger.debug(f'Deserializing object: {d}')
```
awennersteen, line 196: This is cause the string there to always be formed, even if debug logging is not created, with a performance penalty. In the case of deserializing we might not care, but in general you shouldnt use f-strings in debug statements.
```suggestion
    logger.debug("Deserializing object: {}".format(d))
```

see e.g.:
https://stackoverflow.com/a/49884004

awennersteen: Actually, rather than doing this here,
I'd suggest that we can add a debug log statement to the exception handling, so that we only print it when the exception is hit

awennersteen:
this is still open @dominikandreasseitz

PR #305, opened 2024-01-26
[Feature] Adding quantum information tools

inafergra:
This PR adds two functionalities related to quantum information:

- Quantum Fisher Information (QFI)
- Quantum Natural Gradient (QNG): gradient-optimizer that uses the QFI matrix to guide the parameter update. It can either be used with the "exact" QFI or using a SPSA approximation (https://arxiv.org/pdf/2103.09232.pdf)

The code works but it is not very efficient, in particular the QNG training loop. Most of the code was developed for previous versions of qadence so there could also more straightforward ways to calculate the QFI matrix. Any help optimizing the code and making the code more efficient is very welcome!

### DIFF
--- a/examples/qinfo_tools/quantum_natural_gradient.py
+++ b/examples/qinfo_tools/quantum_natural_gradient.py
@@ -1,0 +1,121 @@
+import torch
+from torch.utils.data import random_split
+import qadence as qd
+from qadence.qinfo_tools import QuantumNaturalGradient
+import matplotlib.pyplot as plt
+
+# make sure all tensors are kept on the same device
+# only available from PyTorch 2.0
+device = "cuda" if torch.cuda.is_available() else "cpu"
+torch.set_default_device(device)
+
+n_qubits = 4
+
+
+def qcl_training_data(
+    domain: tuple = (0, 2 * torch.pi), n_points: int = 200
+) -> tuple[torch.Tensor, torch.Tensor]:
+    start, end = domain
+
+    x_rand, _ = torch.sort(torch.DoubleTensor(n_points).uniform_(start, end))
+    y_rand = torch.sin(x_rand)
+
+    return x_rand, y_rand
+
+
+x, y = qcl_training_data()
+
+# random train/test split of the dataset
+train_subset, test_subset = random_split(x, [0.75, 0.25])
+train_ind = sorted(train_subset.indices)
+test_ind = sorted(test_subset.indices)
+
+x_train, y_train = x[train_ind], y[train_ind]
+x_test, y_test = x[test_ind], y[test_ind]
+
+
+# create a simple feature map to encode the input data
+feature_param = qd.FeatureParameter("phi")
+feature_map = qd.kron(qd.RX(i, feature_param) for i in range(n_qubits))
+feature_map = qd.tag(feature_map, "feature_map")
+
+# create a digital-analog variational ansatz using Qadence convenience constructors
+ansatz = qd.hea(n_qubits, depth=n_qubits)
+ansatz = qd.tag(ansatz, "ansatz")
+
+# Observable
+observable = qd.hamiltonian_factory(n_qubits, detuning=qd.Z)
+
+# Create separate circuits for the two optimizers
+circuit_adam = qd.QuantumCircuit(n_qubits, feature_map, ansatz)
+circuit_qng = qd.QuantumCircuit(n_qubits, feature_map, ansatz)
+circuit_qng_spsa = qd.QuantumCircuit(n_qubits, feature_map, ansatz)
+
+model_adam = qd.QNN(circuit_adam, [observable])
+model_qng = qd.QNN(circuit_qng, [observable])
+model_qng_spsa = qd.QNN(circuit_qng_spsa, [observable])
+
+circ_params_qng = [param for param in model_qng.parameters() if param.requires_grad]
+circ_params_qng_spsa = [param for param in model_qng_spsa.parameters() if param.requires_grad]
+
+# Train with ADAM
+n_epochs_adam = 40
+lr_adam = 0.01
+mse_loss = torch.nn.MSELoss()  # standard PyTorch loss function
+optimizer = torch.optim.Adam(model_adam.parameters(), lr=lr_adam)  # standard PyTorch Adam optimizer
+loss_adam = []
+print(f"Initial loss: {mse_loss(model_adam(values=x_train), y_train)}")
+for i in range(n_epochs_adam):
+    optimizer.zero_grad()
+    loss = mse_loss(model_adam(values=x_train).squeeze(), y_train)
+    loss_adam.append(float(loss))
+    loss.backward()
+    optimizer.step()
+    if (i + 1) % 2 == 0:
+        print(f"Epoch {i+1} - Loss: {loss.item()}")
+
+# Train with QNG
+n_epochs_qng = 20
+lr_qng = 0.01
+mse_loss = torch.nn.MSELoss()  # standard PyTorch loss function
+optimizer = QuantumNaturalGradient(circ_params_qng, lr=lr_qng, circuit=circuit_qng)
+loss_qng = []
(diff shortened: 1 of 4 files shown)


qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,489 @@
...
+from qadence import DiffMode
+from qadence.blocks import primitive_blocks, parameters
+
+
+def symsqrt(a, cond=None, return_rank=False):
+    """Computes the symmetric square root of a positive definite matrix.
+    Taken from https://github.com/pytorch/pytorch/issues/25481"""
+
+    s, u = torch.linalg.eigh(a)
+    cond_dict = {
```
gvelikova, line 19: I guess thus can be an Enum instead.

qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,489 @@
...
+from qadence.circuit import QuantumCircuit
+from qadence.backend import BackendName
+from qadence.measurements import Measurements
+from qadence import DiffMode
+from qadence.blocks import primitive_blocks, parameters
+
+
+def symsqrt(a, cond=None, return_rank=False):
+    """Computes the symmetric square root of a positive definite matrix.
+    Taken from https://github.com/pytorch/pytorch/issues/25481"""
```
gvelikova, line 16: It is fair to cite the origin of the code.

qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,489 @@
...
+    if return_rank:
+        return B, len(psigma_diag)
+    else:
+        return B
+
+
+def _overlap_with_inputs(
+    circuit: QuantumCircuit,
+    vparams: tuple | None = None,
+    backend: BackendName = BackendName.PYQTORCH,
```
gvelikova, line 42: Have you tested what is the performance with the other backends and GPSR? I am asking out of curiousity.

qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,489 @@
...
+    overlap_method: OverlapMethod = OverlapMethod.EXACT,
+    diff_mode: DiffMode = DiffMode.AD,
+):
+    """Gives an OverlapModel consisting of the overlap of the circuit with itself. The vparamaters
+    of the underlying QuantumModel are the vparameters of the bra state, so it's differentiable.
+
+    Args:
+        circuit (QuantumCircuit): _description_
```
gvelikova, line 50: I guess the docstring here was WIP by definition.

qadence/qinfo_tools/qng.py
```diff
@@ -0,0 +1,158 @@
...
+
+            else:
+                raise NotImplementedError(
+                    "Non valid approximation, please choose 'exact' or 'spsa'."
+                )
+
+        return loss
+
+
+# class QuantumNaturalGradient(Optimizer):
```
gvelikova, line 102: The commented code can be cleaned out.

qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,489 @@
...
+        qfi_mat_estimator = (1 / (k + 1)) * (k * previous_qfi_estimator + qfi_mat)
+
+    # Get the positive-semidefinite version of the matrix for the update rule in QNG
+    qfi_mat_positive_sd = symsqrt(torch.matmul(qfi_mat_estimator, qfi_mat_estimator))
+    qfi_mat_positive_sd = qfi_mat_positive_sd + beta * torch.eye(num_vparams)
+
+    return qfi_mat_estimator, qfi_mat_positive_sd
+
+
+# def symsqrt(a, cond=None, return_rank=False):
```
gvelikova, line 252: The commented code can be cleaned out.

gvelikova:
@inafergra thank you for adding the QNG and QFI  as qadence features.

qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,261 @@
...
+            + torch.matmul(Delta_2, Delta_1.transpose(0, 1))
+        )
+    )
+    qfi_mat = -2 * fid_hess
+
+    # Calculate the new estimator from the old estimator of qfi_mat
+    if k == 0:
+        qfi_mat_estimator = qfi_mat
+    else:
+        qfi_mat_estimator = (1 / (k + 1)) * (k * previous_qfi_estimator + qfi_mat)  # type: ignore
```
gvelikova, line 255: This previous_qfi_estimator isn't defined anywhere it seems.

qadence/qinfo_tools/qfi.py
```diff
@@ -0,0 +1,261 @@
...
+    Args:
+        circuit (QuantumCircuit): _description_
+        backend (BackendName, optional): _description_. Defaults to BackendName.pyq.
+        overlap_method (OverlapMethod, optional): _description_. Defaults to OverlapMethod.EXACT.
+        diff_mode (DiffMode, optional): _description_. Defaults to DiffMode.ad.
+
+    Returns:
+        torch.Tensor: QFI matrix
+    """
+    num_vparams = len(var_values) if var_values is not None else 0
```
gvelikova, line 186: I added the if/else condition depending on type. When would you have None parameters and should it make sense to add a better checking?

dominikandreasseitz:
thanks a lot @inafergra , had a quick glance and it looks very nice. good job! i'll have a closer look sometime tomorrow or friday

qadence/qinfo_tools/qfi.py
inafergra: Not yet! I have only tested pyqtorch for the moment

qadence/qinfo_tools/qfi.py
inafergra: True! I think the docstrings need to be updated in general, I will check them all and make sure they are updated.

`previous_qfi_estimator` is an input to the `get_quantum_fisher_spsa()`. This function computes the estimator of the QFI matrix corresponding to the iteration k (also an input to the function). You need the estimator from the previous iteration (k-1) to construct the new estimator.

qadence/qinfo_tools/qfi.py
inafergra: `var_values` are the values of the variational parameters where you want to compute your QFI matrix. When they are set to `None`, the QFI is calculated in with the default values for the vparams in the model (for example, if you call it after training your model and you want to compute the QFI with the values of the optimal vparams, you can just set `var_values` to None and it will automatically compute it there). However, if you want to calculate the QFI in another point of your landscape, you can pass those values in the input.

gvelikova: OK, thanks for the explanation, but the original len(None) in these cases wouldn't work or am I missing something?

examples/qinfo_tools/quantum_natural_gradient.py
```diff
@@ -0,0 +1,124 @@
...
+
+# Plot losses
+fig, _ = plt.subplots()
+plt.plot(range(n_epochs_adam), loss_adam, label="Adam optimizer")
+plt.plot(range(n_epochs_qng), loss_qng, label="QNG optimizer")
+plt.plot(range(n_epochs_qng_spsa), loss_qng_spsa, label="QNG-SPSA optimizer")
+plt.legend()
+plt.xlabel("Training epochs")
+plt.ylabel("Loss")
+plt.show()
```
dominikandreasseitz, line 124: lets make a tutorial in markdown format too. we can show plots there. however we shouldnt plot in example files since they are run in the CI

Roland-djee:
@inafergra As you are probably aware, there is now [qadence-libs](https://github.com/pasqal-io/qadence-libs) where this work is best suited. Thanks for the contribution!

inafergra:
> @inafergra As you are probably aware, there is now [qadence-libs](https://github.com/pasqal-io/qadence-libs) where this work is best suited. Thanks for the contribution!

inafergra:
Closing the PR due to move to [qadence-libs](https://github.com/pasqal-io/qadence-libs)

PR #308, opened 2024-01-30
[Fix] Correctly move models to GPU

dominikandreasseitz:
Bugfix depending on https://github.com/pasqal-io/pyqtorch/pull/135

1. Custom 'to' device method for QuantumModel and TransformedModule
2. Fix bug in PyQ diagonal observable where register was not moved to device
3. Add logging to QuantumModel and TransformedModule

### DIFF
--- a/qadence/backends/pyqtorch/convert_ops.py
+++ b/qadence/backends/pyqtorch/convert_ops.py
@@ -243,18 +243,15 @@ class PyQObservable(Module):
                 "diagonal_observable", block_to_diagonal(block, tuple(range(n_qubits)))
             )
 
-            def _forward(
-                pyq_obs: PyQObservable, state: Tensor, values: dict[str, Tensor] = None
-            ) -> Tensor:
-                return pyqify(pyq_obs.diagonal_observable * unpyqify(state), n_qubits=self.n_qubits)
-
-            self._forward = _forward
+            self._forward = lambda self, state, values: pyqify(
+                self.diagonal_observable * unpyqify(state), n_qubits=self.n_qubits
+            )
         else:
             self.operation = pyq.QuantumCircuit(
                 n_qubits,
                 convert_block(block, n_qubits, config),
             )
-            self._forward = lambda _, state, values: self.operation(state, values)  # type: ignore[misc, assignment]
+            self._forward = lambda self, state, values: self.operation(state, values)
 
     def run(self, state: Tensor, values: dict[str, Tensor]) -> Tensor:
         return self._forward(self, state, values)
--- a/qadence/backends/pyqtorch/convert_ops.py
+++ b/qadence/backends/pyqtorch/convert_ops.py
@@ -239,11 +239,10 @@ class PyQObservable(Module):
             config = Configuration()
         self.n_qubits = n_qubits
         if block._is_diag_pauli and not block.is_parametric:
-            diag = block_to_diagonal(block, tuple(range(n_qubits)))
-            self.register_buffer("diag", diag)
+            self.register_buffer("diag", block_to_diagonal(block, tuple(range(n_qubits))))
 
             def sparse_operation(state: Tensor, values: dict[str, Tensor] = None) -> Tensor:
-                return pyqify(diag * unpyqify(state), n_qubits=self.n_qubits)
+                return pyqify(self.diag * unpyqify(state), n_qubits=self.n_qubits)
 
             self.operation = sparse_operation
         else:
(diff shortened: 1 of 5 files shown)


qadence/models/quantum_model.py
```diff
@@ -319,3 +319,21 @@ def assign_parameters(self, values: dict[str, Tensor]) -> Any:
         """Return the final, assigned circuit that is used in e.g. `backend.run`."""
         params = self.embedding_fn(self._params, values)
         return self.backend.assign_parameters(self._circuit, params)
+
+    def to(self, device: torch.DeviceObjType) -> QuantumModel:
+        from copy import deepcopy
```
awennersteen, line 324: why do you do the import in function?

qadence/models/quantum_model.py
```diff
@@ -319,3 +319,21 @@ def assign_parameters(self, values: dict[str, Tensor]) -> Any:
...
+
+        if self._observable is not None:
+            if isinstance(self._observable, ConvertedObservable):
+                self._observable.native = self._observable.native.to(device)
+            elif isinstance(self._observable, list):
+                obs_dev = []
+                for obs in self._observable:
+                    new_obs = deepcopy(obs)
+                    new_obs.native = new_obs.native.to(device)
+                    obs_dev.append(new_obs)
```
awennersteen, line 337: is it necessary to deepcopy?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -239,21 +239,25 @@ def __init__(self, block: AbstractBlock, n_qubits: int, config: Configuration =
...
 
-            def sparse_operation(state: Tensor, values: dict[str, Tensor] = None) -> Tensor:
-                return pyqify(diag * unpyqify(state), n_qubits=self.n_qubits)
+            def _forward(
+                pyq_obs: PyQObservable, state: Tensor, values: dict[str, Tensor] = None
+            ) -> Tensor:
+                return pyqify(pyq_obs.diagonal_observable * unpyqify(state), n_qubits=self.n_qubits)
 
-            self.operation = sparse_operation
+            self._forward = _forward
```
awennersteen, line 251: hmmm? I'm confused

awennersteen: Ok, looking at the whole file I think I understand. It's not very pretty :p

I guess I find it strange to make an actual function rather than a lambda if we first do this.

dominikandreasseitz: yeah its far from pretty

qadence/models/quantum_model.py
dominikandreasseitz: no copying anymore

qadence/models/quantum_model.py
dominikandreasseitz: same as below

awennersteen (approved):

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: improved it now

qadence/models/quantum_model.py
```diff
@@ -319,3 +334,20 @@ def assign_parameters(self, values: dict[str, Tensor]) -> Any:
...
+            if isinstance(self._circuit.native, torch.nn.Module):
+                # Backends which are not torch-based cannot be moved to 'device'
+                self._circuit.native = self._circuit.native.to(device)
+                if self._observable is not None:
+                    if isinstance(self._observable, ConvertedObservable):
+                        self._observable.native = self._observable.native.to(device)
+                    elif isinstance(self._observable, list):
+                        for obs in self._observable:
+                            obs.native = obs.native.to(device)
+            logger.debug(f"Moved {self} to device {device}.")
```
awennersteen, line 350:
```suggestion
            if isinstance(self._circuit.native, torch.nn.Module):
                # Backends which are not torch-based cannot be moved to 'device'
                 self._params = self._params.to(device)
                self._circuit.native = self._circuit.native.to(device)
                if self._observable is not None:
                    if isinstance(self._observable, ConvertedObservable):
                        self._observable.native = self._observable.native.to(device)
                    elif isinstance(self._observable, list):
                        for obs in self._observable:
                            obs.native = obs.native.to(device)
                logger.debug(f"Moved {self} to device {device}.")
```

dominikandreasseitz: good catch

PR #315, opened 2024-02-01
Store the noise model when passed to sample method.

Roland-djee:
Store the noise model as QM class attribute for later retrieval when passed to methods.


dominikandreasseitz (approved):
thanks for that

dominikandreasseitz:
bump please if you wanna release

PR #316, opened 2024-02-02
[Fix] Add inputs to qnn deserialize, add serialization of noise and mitigation in QM

dominikandreasseitz:
Fixes https://github.com/pasqal-io/qadence/issues/324


qadence/models/qnn.py
```diff
@@ -218,3 +222,43 @@ def _format_to_dict(self, values: Tensor) -> dict[str, Tensor]:
...
+
+        qnn: QNN
+        try:
+            qm_dict = d[cls.__name__]
+            qnn = cls(
+                circuit=QuantumCircuit._from_dict(qm_dict["circuit"]),
+                observable=[deserialize(q_obs) for q_obs in qm_dict["observable"]],  # type: ignore[misc]
+                backend=qm_dict["backend"],
+                diff_mode=qm_dict["diff_mode"],
+                measurement=Measurements._from_dict(qm_dict["measurement"]),
```
Roland-djee, line 248: Don't you miss the `noise` parameter ? On that note, if QNN is a QM I guess the `mitigation` parameter should be present.

Roland-djee (commented):
Thanks @dominikandreasseitz. I think there are parameters missing.

gvelikova (approved):

qadence/models/qnn.py
jpmoutinho: It seems noise and mitigation are not there in the serialization of `QuantumModel` as well

dominikandreasseitz: added

PR #318, opened 2024-02-02
[Fix] PyQ HamEvo/ Observable GPU buffer fixes, Add device to block_to_tensor

dominikandreasseitz:
Fixes https://github.com/pasqal-io/qadence/issues/322
Fixes https://github.com/pasqal-io/qadence/issues/323

- [x]  Move HamEvo buffers to device
- [x] Use Pyq.QuantumCircuit 'to' for PyQObservable
- [x] Add block_to_tensor device method
- [x] Use pyq v1.0.5 with device fixes


qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -291,23 +292,25 @@ def __init__(
...
             self._hamiltonian = (
-                lambda values: values[self.param_names[1]].squeeze(3).permute(1, 2, 0)
+                lambda self, values: values[self.param_names[1]].squeeze(3).permute(1, 2, 0)
             )
             # FIXME Why are we squeezing
         else:
 
-            def _hamiltonian(values: dict[str, Tensor]) -> Tensor:
+            def _hamiltonian(self: PyQHamiltonianEvolution, values: dict[str, Tensor]) -> Tensor:
+                _dev = list(values.values())[0].device
```
awennersteen, line 306: Could an empty dict be passed here? For unparameterized Hamiltonian?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -291,23 +292,25 @@ def __init__(
                 use_full_support=False,
             )
             hmat = hmat.permute(1, 2, 0)
-            self._hamiltonian = lambda x: hmat
+            self.register_buffer("hmat", hmat)
+            self._hamiltonian = lambda self, values: self.hmat
```
awennersteen, line 296: So I understand that the `self.register_buffer("hmat", hmat)` also implicitlly makes `self.hmat` defined?

dominikandreasseitz: from the torch docs "Buffers can be accessed as attributes using given names."

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: not really, this should only be executed if you have parametrized generators although its not really clear from this long __init__.  non-parametric generators will go into the first if clause

awennersteen (approved):

pyproject.toml
```diff
@@ -20,7 +20,7 @@ authors = [
 ]
 requires-python = ">=3.9,<3.12"
 license = {text = "Apache 2.0"}
-version = "1.2.7"
```
awennersteen, line 23: 1.3.0?
we've made some interface changes here and in the previous

dominikandreasseitz: good point. changed

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -232,9 +244,9 @@ def _batch_last(m: Tensor) -> Tensor:
         )
 
 
-class PyQObservable(Module):
+class PyQObservable(pyq.QuantumCircuit):
     def __init__(self, block: AbstractBlock, n_qubits: int, config: Configuration = None):
-        super().__init__()
+        super().__init__(n_qubits, [])
```
awennersteen, line 249: I'm a bit scared of this init with `[]`, are we sure it's ok?

mutable defaults (as I think this might be regarded) is scary

awennersteen (changes requested):

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: redone, you can check it tomorrow @awennersteen

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -281,7 +301,8 @@ def __init__(
                 use_full_support=False,
             )
             hmat = hmat.permute(1, 2, 0)
-            self._hamiltonian = lambda x: hmat
+            self.register_buffer("hmat", hmat)
+            self._hamiltonian = lambda self, values: self.hmat
```
Roland-djee, line 305: Not quite sure why this has to be a lambda ?

Roland-djee (approved):
Thanks for this @dominikandreasseitz looks like the right way forward.

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -239,26 +251,33 @@ def __init__(self, block: AbstractBlock, n_qubits: int, config: Configuration =
...
         self.n_qubits = n_qubits
         if block._is_diag_pauli and not block.is_parametric:
-            self.register_buffer(
-                "diagonal_observable", block_to_diagonal(block, tuple(range(n_qubits)))
-            )
-
+            self.register_buffer("operation", block_to_diagonal(block, tuple(range(n_qubits))))
             self._forward = lambda self, state, values: pyqify(
-                self.diagonal_observable * unpyqify(state), n_qubits=self.n_qubits
+                self.operation * unpyqify(state), n_qubits=self.n_qubits
```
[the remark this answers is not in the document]
dominikandreasseitz: because we do a outer product between the diagonal and the state. so the state needs to be flattened

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: yes, we need `self` since the state (for example the device) can change after instantiation.

PR #319, opened 2024-02-02
[BugFix] Wrong tensor for control blocks, inefficient eigenvalues

jpmoutinho:
See https://github.com/pasqal-io/qadence/issues/320 for context on eigenvalues change.

Furthermore a recent change to the control blocks made `block_to_tensor` give incorrect results for non-sequential qubit supports, which was not tested.

### DIFF
--- a/qadence/blocks/primitive.py
+++ b/qadence/blocks/primitive.py
@@ -381,10 +381,6 @@ class ControlBlock(PrimitiveBlock):
     def __ascii__(self, console: Console) -> RenderableType:
         raise NotImplementedError
 
-    @property
-    def n_qubits(self) -> int:
-        return len(self.qubit_support)
-
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, AbstractBlock):
             raise TypeError(f"Cant compare {type(self)} to {type(other)}")
--- a/qadence/utils.py
+++ b/qadence/utils.py
@@ -7,7 +7,6 @@ from typing import Any
 import numpy as np
 import sympy
 import torch
-from scipy.sparse.linalg import eigs
 from torch.linalg import eigvals
 
 from qadence.logger import get_logger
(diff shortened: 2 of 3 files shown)


jpmoutinho:
See #320 for context on eigenvalues change.

qadence/blocks/primitive.py
```diff
@@ -381,10 +381,6 @@ def _block_title(self) -> str:
     def __ascii__(self, console: Console) -> RenderableType:
         raise NotImplementedError
 
-    @property
-    def n_qubits(self) -> int:
```
Roland-djee, line 385: Is this redundant with the base class property ?

qadence/utils.py
```diff
@@ -184,16 +183,23 @@ def isclose(
 def eigenvalues(
     x: torch.Tensor, max_num_evals: int | None = None, max_num_gaps: int | None = None
 ) -> torch.Tensor:
+    # FIXME: Currently doing full decomposition everytime because it is generally faster
+    # than converting to numpy and using the scipy routines for "top-k" eigenvalues.
```
Roland-djee, line 187:
```suggestion
    # than converting to numpy and using the scipy routines to select first k-eigenvalues.
```

qadence/utils.py
```diff
@@ -184,16 +183,23 @@ def isclose(
 def eigenvalues(
     x: torch.Tensor, max_num_evals: int | None = None, max_num_gaps: int | None = None
```
Roland-djee, line 184: I would suggest to:

```python
from torch import Tensor
```

More idiomatic to Qadence.

qadence/utils.py
```diff
@@ -184,16 +183,23 @@ def isclose(
...
 ) -> torch.Tensor:
+    # FIXME: Currently doing full decomposition everytime because it is generally faster
+    # than converting to numpy and using the scipy routines for "top-k" eigenvalues.
+    # Furthermore, we should exploit the matrices being Hermitian.
+
+    # get all eigenvalues of generator
+    eigenvals_full = eigvals(x).squeeze(0)
```
Roland-djee, line 191:
```suggestion
    eigenspectrum = eigvals(x).squeeze(0)
```

Roland-djee (approved):
Thanks @jpmoutinho LGTM. Cosmetic suggestions.

qadence/blocks/primitive.py
jpmoutinho: Not just redundant, it was overriding the `max(qubit_support)+1` which would cause wrong results in `block_to_tensor`.

PR #328, opened 2024-02-06
[Feature] Add protocols extras

### LINKED ISSUE #326: Add protocols and libs as extras
- [x] Add protocols.py file
- [x] Add to pyproject.toml
- [ ] Migrate and link documentation

Roland-djee:
Closes #326.

### DIFF
--- a/qadence/protocols.py
+++ b/qadence/protocols.py
@@ -1,0 +1,19 @@
+from __future__ import annotations
+
+import importlib
+from string import Template
+
+from qadence.backend import Backend
+from qadence.blocks.abstract import TAbstractBlock
+from qadence.logger import get_logger
+from qadence.types import BackendName, DiffMode, Engine
+
+backends_namespace = Template("qadence.backends.$name")
+
+logger = get_logger(__name__)
+
+try:
+    module = importlib.import_module("qadence_protocols.mitig")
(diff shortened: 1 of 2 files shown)


qadence/protocols.py
```diff
@@ -0,0 +1,16 @@
...
+logger = get_logger(__name__)
+
+try:
+    module = importlib.import_module("qadence_protocols.protocols")
+    available_protocols = getattr(module, "available_protocols")
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "The 'qadence_protocols' module is not present."
+        " Please install the 'qadence-protocol' package."
+    )
```
jpmoutinho, line 16: Does this make the stuff in protocols importable from `qadence` or does the user need to import stuff from `qadence-protocols` ?

jpmoutinho (approved):

qadence/protocols.py
Roland-djee: It doesn't add functionality from `qadence_protocols` to the `qadence` namespace if that's what you mean. Once you've `pip install` protocols, you still have to do:

```python
from qadence_protocols.mitigations.protocol import Mitigations
```

for it to work. Happy to revisit this if found impractical.

PR #330, opened 2024-02-07
[Refac] Add 'device' to ml_tools optimize_step

dominikandreasseitz:
To avoid having to let the user move arbitary (custom) data loader types to the device, we reintroduce the 'device' arg to `optimize_step` which uses `data_to_device` method which supports a variety of qadence-compatible data types.

### DIFF
--- a/qadence/ml_tools/optimize_step.py
+++ b/qadence/ml_tools/optimize_step.py
@@ -6,12 +6,15 @@ import torch
 from torch.nn import Module
 from torch.optim import Optimizer
 
+from qadence.ml_tools.data import data_to_device
+
 
 def optimize_step(
     model: Module,
     optimizer: Optimizer,
     loss_fn: Callable,
     xs: dict | list | torch.Tensor | None,
+    device: torch.device = torch.device("cpu"),
 ) -> tuple[torch.Tensor | float, dict | None]:
     """Default Torch optimize step with closure.
 
@@ -31,6 +34,7 @@ def optimize_step(
     """
 
     loss, metrics = None, {}
+    xs_to_device = data_to_device(xs, device)
 
     def closure() -> Any:
         # NOTE: We need the nonlocal as we can't return a metric dict and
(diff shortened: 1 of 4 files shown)


qadence/ml_tools/optimize_step.py
```diff
@@ -6,12 +6,15 @@
...
 
+from qadence.ml_tools.data import data_to_device
+
 
 def optimize_step(
     model: Module,
     optimizer: Optimizer,
     loss_fn: Callable,
     xs: dict | list | torch.Tensor | None,
+    device: torch.device = torch.device("cpu"),
```
awennersteen, line 17: what about making device default to None? this way if everything already is in it's place, it won't be moved. And we only move if explictly requested.

```
>>> t = torch.randn(2,2)
>>> t
tensor([[-0.2409, -0.3257],
        [-0.0846,  1.2733]])
>>> t.device
device(type='cpu')
>>> t.dtype
torch.float32
>>> t=t.to(torch.device("cuda:0"), dtype=torch.float64)
>>> t
tensor([[-0.2409, -0.3257],
        [-0.0846,  1.2733]], device='cuda:0', dtype=torch.float64)
>>> t.to(None)
tensor([[-0.2409, -0.3257],
        [-0.0846,  1.2733]], device='cuda:0', dtype=torch.float64)
```

dominikandreasseitz: good idea, added

qadence/ml_tools/train_grad.py
```diff
@@ -10,7 +10,7 @@
 
 from qadence.logger import get_logger
 from qadence.ml_tools.config import TrainConfig
-from qadence.ml_tools.data import DictDataLoader, data_to_device
+from qadence.ml_tools.data import DictDataLoader
 from qadence.ml_tools.optimize_step import optimize_step
 from qadence.ml_tools.printing import print_metrics, write_tensorboard
 from qadence.ml_tools.saveload import load_checkpoint, write_checkpoint
```
awennersteen, line 16: should we make `device` in `train(...)` also default to `None`?

awennersteen (approved):
One suggestion, otherwise LGTM

awennersteen:
I see you've tagged it breaking, I'm not sure if I see why. But if it's breaking shouldn't we at least bump minor rather than patch?

qadence/ml_tools/train_grad.py
dominikandreasseitz: done

dominikandreasseitz:
@awennersteen not sure about the. breaking change either, i tested it with current flow and it seems to be fine. so patch might be enough

PR #341, opened 2024-02-12
[Draft, Refact] Deprecate constructors

jpmoutinho:
To be merged after https://github.com/pasqal-io/qadence-libs/pull/8

### DIFF
--- a/qadence/constructors/ansatze.py
+++ b/qadence/constructors/ansatze.py
@@ -4,6 +4,7 @@ import itertools
 from typing import Any, Optional, Type, Union
 
 from qadence.blocks import AbstractBlock, block_is_qubit_hamiltonian, chain, kron, tag
+from qadence.logger import get_logger
 from qadence.operations import CNOT, CPHASE, CRX, CRY, CRZ, CZ, RX, RY, HamEvo
 from qadence.types import Interaction, Strategy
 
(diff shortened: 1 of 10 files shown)


dominikandreasseitz (approved):
I guess this warnings will not be seen by qdevs who develop locally. didnt we use something like UserWarning at some point? @awennersteen

jpmoutinho:
> I guess this warnings will not be seen by qdevs who develop locally. didnt we use something like UserWarning at some point? @awennersteen

Why not? They show in notebooks

awennersteen:
They'll show I think as the default log level in Python is warning.
However, I think the best practise is to use the warnings module, see e.g. this stackoverflow:

https://stackoverflow.com/a/72285990

Roland-djee:
LGTM. I think this is fine now. However, for future reference, we should keep in mind using the `warnings` module as suggested @awennersteen.

Roland-djee (approved):

Roland-djee:
It is mergeable now @jpmoutinho ? Since the other is merged.

jpmoutinho:
@Roland-djee I do mean to make the changes to the warnings here, and also wanted to sync with you about the qadence-libs in general before merging.

Roland-djee:
> @Roland-djee I do mean to make the changes to the warnings here, and also wanted to sync with you about the qadence-libs in general before merging.

Sure thing.

PR #342, opened 2024-02-12
[Draft, Refact] Trying to remove global qubit support

jpmoutinho:
This is turning out very messy and super breaking.


Roland-djee:
@jpmoutinho I think this should be prototyped with the block system refactoring.

jpmoutinho:
Closing this, not going to pursue it.

Issue #291, opened 2024-01-10
[Refactoring] Simply the internal block struture for analog blocks


jpmoutinho:
Currently there are a lot of definitions in the internal structure of the analog blocks.

The most user facing are defined as python functions that return an instance of a class:
- `AnalogRX`, `AnalogRY`, `AnalogRZ` for a specific axis rotation
- `AnalogRot` which is the general version of the above with control over all Hamiltonian parameters
- `AnalogInteraction` which free-evolves the intercation term

The functions above return an instance of:
- `ConstantAnalogRotation`
- `InteractionBlock`

And then these classes have an actual implementation in the backends, where for pyqtorch they get converted into an instane of `HamEvo`, and for Pulser they get translated into a pulse sequence.

To simplify, we could simply have the user-facing operations themselves be defined as classes and possibly get rid of the "intermediate" blocks.

Relevant issues to keep in mind:
- https://github.com/pasqal-io/qadence/issues/192 - a different proposal that could potentially get rid of `AnalogRX`, `AnalogRY`, `AnalogRZ` (and "global" qubit support), but would be harder to implement.
- https://github.com/pasqal-io/qadence/issues/239 - eigenvalues are required for GPSR, and currently they are not directly available from the `ConstantAnalogRotation` and `InteractionBlock` (partially because these accept a "global" qubit support).

jpmoutinho:
@Roland-djee I came back to this issue and played around with the related MR. I am unsure if it should be pursued. It's a huge amount of work to do the complete refactoring, and I think we already have the necessary knowledge on how to properly design a similar feature in the new block system that will avoid all the spaghetti in the current analog blocks.

Roland-djee:
> @Roland-djee I came back to this issue and played around with the related MR. I am unsure if it should be pursued. It's a huge amount of work to do the complete refactoring, and I think we already have the necessary knowledge on how to properly design a similar feature in the new block system that will avoid all the spaghetti in the current analog blocks.

Hey, indeed. I would focus on something currently more pressing if this is going to be solved in the new block system.

Issue #239, opened 2023-12-04
[Refactoring, DA] Better handling of analog blocks eigenvalues


jpmoutinho:
Analog block eigenvalues are not defined because currently analog blocks only get converted into an actual operator when converted to `HamEvo` in the PyQ backend. GPSR requires these eigenvalues and currently there is a workaround where the pulser backend will call the PyQ backend analog block converter just to get the `HamEvo` and get the eigenvalues from there.

PR #344, opened 2024-02-14
[Draft, Refact] Analog ops inheriting directly from HamEvo

jpmoutinho:
Closes https://github.com/pasqal-io/qadence/issues/291

Basically the main change is that this requires the register to be passed as an argument to these operations. I don't think there is a way around this, because they are inherently dependent on the register topology. From the user perspective, I think this is fine, because they will likely define their register before defining anything else in their program.

The previous design was "hacking" it by delaying the usage of the register information until the final transpilation pass that went over the circuit. If we simply make that information available from the start we can cut down on a lot of code complexity.

The solution started in this MR has the following advantages:
- It does not use anything in `blocks/analog.py`, everything is just `HamEvo`. We can check the various attributes in the analog blocks and add some to the new classes if needed.
- It does not require a special transpilation pass of each circuit. The code in `qadence/analog/parse_analog.py` can be removed.
- It solves https://github.com/pasqal-io/qadence/issues/291
- It solves (or makes it easily solvable) https://github.com/pasqal-io/qadence/issues/239
- It avoids trying to merge something messy like this: https://github.com/pasqal-io/qadence/pull/342, which still wouldn't fix the issues above.

This will be a breaking change, but I am not seeing a way forward with the analog blocks that is not.

Note: keeping this MR as a test-ground while we define the best way forward with the analog blocks.


Roland-djee:
@jpmoutinho Is this related to the idea of making Hamiltonian generators central objects in Qadence ?

jpmoutinho:
@Roland-djee in part yes, but that was already the case for the Analog blocks. The proposal here is just a refactoring where they actually inherit directly from `HamEvo`.

vytautas-a:
I think the general idea here is correct only I would go even further and make also the digital gates inherit from `HamEvo` and simply setting the appropriate generators. Even now we can in principle build digital circuits by specifying necessary generator matrices to `HamEvo` and have the whole circuit consist only of such blocks. So we would only simplify setting this digital generator in the derived digital gate classes.

jpmoutinho:
> I think the general idea here is correct only I would go even further and make also the digital gates inherit from `HamEvo` and simply setting the appropriate generators. Even now we can in principle build digital circuits by specifying necessary generator matrices to `HamEvo` and have the whole circuit consist only of such blocks. So we would only simplify setting this digital generator in the derived digital gate classes.
> 
> Bottom line - I would argue for such unification already at this stage because it does not cost us much more work.

Thanks @vytautas-a, indeed I think this is a good direction to move towards (and I think also in line with previous discussions with @Roland-djee). But we need to think carefully on it and plan how it will integrate with a revised block system, because we don't want the digital gates to _always_ be handled as `HamEvo` in the backend, where the generator is exponentiated. For the standard ones where we have the unitary matrix available it is more efficient to use that directly.

jpmoutinho:
> > > I think the general idea here is correct only I would go even further and make also the digital gates inherit from `HamEvo` and simply setting the appropriate generators. Even now we can in principle build digital circuits by specifying necessary generator matrices to `HamEvo` and have the whole circuit consist only of such blocks. So we would only simplify setting this digital generator in the derived digital gate classes.
> > > Bottom line - I would argue for such unification already at this stage because it does not cost us much more work.
> > 
> > 
> > Thanks @vytautas-a, indeed I think this is a good direction to move towards (and I think also in line with previous discussions with @Roland-djee). But we need to think carefully on it and plan how it will integrate with a revised block system, because we don't want the digital gates to _always_ be handled as `HamEvo` in the backend, where the generator is exponentiated. For the standard ones where we have the unitary matrix available it is more efficient to use that directly.
> 
> Sounds fair but is there any other way ? I mean, for purely digital computations we may not want to exponentiate the Hamiltonian. But for analog and digital-analog ?

I think it needs to be handled in some smart way by the block system. If a block is a combination of only digital gates it should recognize it only needs to grab the unitaries and perform the computation, and any other combination will likely require combining the generators and exponentiating.

vytautas-a:
Maybe all the gates could be derived from `HamEvo` and have a property `unitary` but only purely digital ones would have this property predefined. This way we could distinguish whether the gate's generator has to be exponentiated or not.

Roland-djee:
> Maybe all the gates could be derived from `HamEvo` and have a property `unitary` but only purely digital ones would have this property predefined. This way we could distinguish whether the gate's generator has to be exponentiated or not.

They can be cached. And we can check if this cache exists. Or something like this.

PR #345, opened 2024-02-15
[Fix] Add atol for jax type conversion

dominikandreasseitz:


tests/metrics.py
```diff
@@ -24,3 +24,4 @@
 LARGE_SPACING = 30.0
 DIGITAL_DECOMP_ACCEPTANCE_HIGH = 1e-2
 DIGITAL_DECOMP_ACCEPTANCE_LOW = 1e-3
+JAX_CONVERSION_ATOL = ATOL_E6
```
jpmoutinho, line 27: Just using `ATOL_32` should also be ok

jpmoutinho (approved):

tests/metrics.py
dominikandreasseitz: yeah lets be safe

PR #349, opened 2024-02-20
[Docs] Improve qaoa tutorial

### LINKED ISSUE #282: [Documentation] Improve QAOA tutorial
The QAOA tutorial to solve Max-Cut is currently not showing the standard QAOA algorithm. In particular:

- The current version uses $N$ beta parameters per layer, where $N$ is the number of qubits. While there are some QAOA variations that do this, the standard QAOA uses only a single beta parameter per layer.
- The layer of Hadamards at the beginning of the circuit is missing.

inafergra:
Closes #282


docs/qml/qaoa.md
```diff
@@ -34,57 +36,111 @@ The goal of the MaxCut algorithm is to maximize the following cost function:
 
 $$\mathcal{C}(p) = \sum_{\alpha}^m \mathcal{C}_{\alpha}(p)$$
 
-where $p$ is the given partition of the graph, $\alpha$ is an index over the edges and $\mathcal{C}_{\alpha}(p)$ is written
+where $p$ is a given cut of the graph, $\alpha$ is an index over the edges and $\mathcal{C}_{\alpha}(p)$ is written
 such that if the nodes connected by the $\alpha$ edge are in the same set, it returns $0$, otherwise it returns $1$.
 
+Since in this tutorial we are only dealing with small graphs, we can find the maximum cut by brute force to make sure QAOA works as intended. Recall that every cut is represented by two bitstrings, e.g. cuts "0011" and "1100" are equivalent as they represent the same cut.
```
jpmoutinho, line 42: What is the meaning of the bitstrings here exactly? Is the number of bits equal to the number of edges?

docs/qml/qaoa.md
```diff
@@ -34,57 +36,111 @@ The goal of the MaxCut algorithm is to maximize the following cost function:
...
+
+all_possible_cuts = [bin(k)[2:].rjust(n_nodes, "0") for k in range(2**n_nodes)]
+adjacency_matrix = nx.adjacency_matrix(graph)
+all_costs = [calculate_cost(cut, adjacency_matrix) for cut in all_possible_cuts]
+maxcost = max(all_costs)
+maxcut = all_possible_cuts[all_costs.index(maxcost)]
+other_maxcut_bs = "".join(str((int(i) + 1) % 2) for i in maxcut)
+print(
+    f"The maximum cut is represented by the bitstrings {maxcut} and {other_maxcut_bs}, with a cost of {maxcost}"
+)
```
jpmoutinho, line 65: This code is a bit hard to read. Could you add some description text on what these lines are doing?

jpmoutinho (commented):
Thanks a lot for working in this @inafergra. I didn't read everything yet but will take another look tomorrow.

docs/qml/qaoa.md
inafergra: The number of bits is equal to the number of nodes. With bitstrings I refer to binary strings representing each possible solution (each possible "cut" of the graph), assigning "0" to bit $i$ if node $i$ is in partition A and "1" if it is in partition "B" (or viceversa). I will elaborate more on this so it's more clear!

docs/qml/qaoa.md
```diff
@@ -5,22 +5,24 @@ Algorithm (QAOA), first introduced by Farhi *et al.* in 2014 [^1].
 Given an arbitrary graph, the MaxCut problem consists in finding a
 graph [*cut*](https://en.wikipedia.org/wiki/Cut_(graph_theory)) which partitions
 the nodes into two disjoint sets, such that the number of edges in the
-cut is maximized. This is a very common combinatorial optimization problem which is
-computationally hard.
+cut is maximized. This is a very common combinatorial optimization problem known to be computationally hard (NP-hard).
 
-The graph used for this tutorial is a randomly generated using the `networkx` library with
+The graph used for this tutorial is an unweighted graph randomly generated using the `networkx` library with
 a $0.5$ probability of having an edge between two arbitrary nodes.
```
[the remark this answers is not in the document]
inafergra: That's a good point! I will change the probability to ensure that  $p > \frac{1}{n_{nodes}}$ so that we're guaranteed to have a giant component with high probability for all graph sizes.

docs/qml/qaoa.md
```diff
@@ -34,57 +36,111 @@ The goal of the MaxCut algorithm is to maximize the following cost function:
...
+
 
+# Generate the cost Hamiltonian
+zz_ops = [kron(Z(edge[0]), Z(edge[1])) for edge in graph.edges()]
 cost_ham = Zero()
 for op in zz_ops:
-    cost_ham += 0.5 * op
-cost_ham = 0.5 * kron(I(i) for i in range(n_qubits)) - cost_ham
+    cost_ham += op
+cost_ham = 0.5 * (n_edges * kron(I(i) for i in range(n_qubits)) - cost_ham)
```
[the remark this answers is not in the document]
inafergra: It can indeed be written in terms of of the number operator $N=\frac{1}{2}(1-Z)$, but then the formulation is a bit less straightforward and it is not the usual way to formulate MaxCut in this setting. I think that with the number operator the Hamiltonian would be:

$$ H = \sum_{\langle i,j \rangle} N_i + N_j - 2N_i N_j$$

where $\langle i,j \rangle$ are the graph edges. Note that the only difference between the two formulations is that the number operator acts on the zero state as $N|0\rangle = 0$ and the Z operator as $Z|0\rangle = -1$, while they both act equally on the $|1\rangle$. I think it might be better to keep it with the Z operators since it's the usual way to formulate it.

docs/qml/qaoa.md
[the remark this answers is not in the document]
inafergra: I think I will just set it to $p=0.8$ so we're above the threshold for every size but not explain why, to not overpopulate the tutorial.

docs/qml/qaoa.md
```diff
@@ -5,23 +5,25 @@ Algorithm (QAOA), first introduced by Farhi *et al.* in 2014 [^1].
...
-computationally hard.
+cut is maximized. This is a very common combinatorial optimization problem known to be computationally hard (NP-hard).
 
-The graph used for this tutorial is a randomly generated using the `networkx` library with
-a $0.5$ probability of having an edge between two arbitrary nodes.
+The graph used for this tutorial is an unweighted graph randomly generated using the `networkx` library with
+a certain probability $p$ of having an edge between two arbitrary nodes (known as [Erdős–Rényi](https://en.wikipedia.org/wiki/Erd%C5%91s%E2%80%93R%C3%A9nyi_model) graph).
 
 ```python exec="on" source="material-block" html="1" session="qaoa"
 import numpy as np
```
jpmoutinho, line 14: I might have missed it but I think numpy is not used.

jpmoutinho (approved):
Thanks a lot @inafergra, looks much better overall. I see that this is focusing more on the digital side, we might want to change that at some point, but it's fine for now.

One thing, if you haven't already, could you run `hatch -e docs run serve` to see locally if everything is being displayed correctly? Thanks!

docs/qml/qaoa.md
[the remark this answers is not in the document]
jpmoutinho: Yes that's fine.

docs/qml/qaoa.md
jpmoutinho: Got it, thanks.

docs/qml/qaoa.md
inafergra: It is just used to set the seed with `np.random.seed(seed)`, I think it might be needed for reproducibility since `networkx` is built on top of numpy

docs/qml/qaoa.md
```diff
@@ -5,23 +5,25 @@ Algorithm (QAOA), first introduced by Farhi *et al.* in 2014 [^1].
...
+seed = 42
 np.random.seed(seed)
+random.seed(seed)
 
-n_nodes = 8
-edge_prob = 0.5
+# Create random graph
+n_nodes = 4
+edge_prob = 0.8
 graph = nx.gnp_random_graph(n_nodes, edge_prob)
```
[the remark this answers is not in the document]
inafergra: Not really, it could be set to 6 with no issue

inafergra:
I addressed the comments, I kept the number of qubits to 4 so docs build faster but feel free to change it. I'm going on vacations tomorrow so I'll leave it here for the moment but if you think it's ready feel free to resolve the comments and merge!

inafergra:
Also I checked locally and the docs build fine

PR #356, opened 2024-02-23
[Refactor] Avoid using eval in de-serialization

### LINKED ISSUE #355: [Fix] Serialization
Current approach for serialization (`qadence/serialization.py`) is restricted with the regards of with quantum models it can use and **is not safe** due to the use of `eval` on a `str` (in the `deserialize` function) that can come from external source, such as a json file.

@dominikandreasseitz and @Doomsk to coordinate on issues.

@Doomsk to coordinate with @awennersteen  and cloud team (topic for an RFC ?).

Doomsk:
Closes #355 .

While moving `ml_tools` to `qadence-libs`, a refactoring of the `qadence/serialization.py` was needed. During the process, it was discovered a major issue on where an Python's `eval` was used to execute `str` that could come not only from internally serialized data but also from json files. While addressing the issues on the migration, this is an important fix while doing a good refactoring.


pyproject.toml
```diff
@@ -17,6 +17,7 @@ authors = [
     { name = "Joao P. Moutinho", email = "joao.moutinho@pasqal.com"},
     { name = "Vytautas Abramavicius", email = "vytautas.abramavicius@pasqal.com" },
     { name = "Gergana Velikova", email = "gergana.velikova@pasqal.com" },
+    { name = "Eduardo Maschio", email = "eduardo.maschio@pasqal.com" },
```
jpmoutinho, line 20: Not sure this is acceptable.

Doomsk: RIP

Doomsk:
docs are failing and according to investigations made independently by @jpmoutinho it's `braket` related. On `braket` version `1.71.0` seems to be working.

Doomsk:
@jpmoutinho @Roland-djee can you take a look whenever it's more appropriate

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
+        parser.parse(code)
+    except NoMatch as err:
+        print(f"no match error: {err}")
+        return False
+    else:
+        return True
+
+
+@dataclass
+class SerialModel:
```
jpmoutinho, line 95: Why is this dataclass and all the child ones needed?
Not implying there would be some easier way to do it, just to understand :)

jpmoutinho (commented):
Thanks @Doomsk, I think @Roland-djee or @dominikandreasseitz can better judge this.

Are you introducing some new features here that would require also extending the tests?

Also, would this make it easier to serialize external models? I.e. models that people write that inherit from `QuantumModel` or `torch.nn.Module`. For example some time ago I had come up with a prototype to do it that you can find in this closed MR: https://github.com/pasqal-io/qadence/pull/241/files 

I think this would be a good feature to have :)

qadence/serialization.py
Doomsk: @jpmoutinho  thanks for the questions!

> Are you introducing some new features here that would require also extending the tests?

Not really, as long as the current serialization process works, it's all good, since the new code is made just for that purpose.

> Also, would this make it easier to serialize external models?

Definitely! The `ModelSerial` (line 132) accounts for anything that fits under the "quantum model" category, i.e. inherit either from `QuantumModel` or `torch.nn.Module`. As you can check, I removed the `QNN` and `TransformedModule` from the file imports (and thus from the `SUPPORTED_TYPES`, line 54) and it still works!
 The only catch, and that's something that would eventually happen in the previous case as well, is that the user *needs* to have imported the module (probably with the same name as the serialized code), otherwise it won't work.

> Why is this dataclass and all the child ones needed?

The idea is the dataclasses are "structs" or data holders, so they behave like a data wrapper. The `SerialModel` is there basically to define the attributes every single child will need and also to be easy to generically do this (line 303) (so mypy doesn't complain):

```python
obj: SerialModel
[...]
return obj.value
```

without that, we could not have this generic approach.
Having those dataclasses, makes the process of deserializing much easie, clean and clear (you can check the `deserialize` code (line 262)), because now each dataclass handles the data according to its expected values internally.

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
+
 
 ALL_BLOCK_NAMES = [
     n for n in dir(qadenceblocks) if not (n.startswith("__") and n.endswith("__"))
 ] + [n for n in dir(operations) if not (n.startswith("__") and n.endswith("__"))]
+SYMPY_EXPRS = [n for n in dir(core) if not (n.startswith("__") and n.endswith("__"))]
+QADENCE_PARAMETERS = [n for n in dir(parameters) if not (n.startswith("__") and n.endswith("__"))]
+
+
+THIS_PATH = os.path.dirname(__file__)
```
Roland-djee, line 73: I guess we should use `pathlib` which is more flexible for path manipulation than `os.path`.

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
+        grammar = f.read()
+    return ParserPEG(grammar, "Program")
+
+
+def parse_expr_fn(code: str) -> bool:
+    parser = _parser_fn()
+    try:
+        parser.parse(code)
+    except NoMatch as err:
+        print(f"no match error: {err}")
```
Roland-djee, line 88: Raise an exception ?

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
     QuantumModel,
-    QNN,
-    TransformedModule,
     Register,
-    Basic,
+    core.Basic,
     torch.nn.Module,
 ]
 
+EXPR_TYPES = TypingUnion[core.Expr, float, int, str, tuple, dict]
```
Roland-djee, line 63: Use pipe to type unionise ?

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
+
+@dataclass
+class SerialModel:
+    d: dict = dataclass_field(default_factory=dict)
+    value: Any = dataclass_field(init=False)
+
+
+@dataclass
+class BlockTypeSerial(SerialModel):
+    value: AbstractBlock = dataclass_field(init=False)
```
Roland-djee, line 102: Is it coherent to initialise to `False` ? I mean from a typing point of view.

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
+@dataclass
+class QuantumCircuitSerial(SerialModel):
+    value: QuantumCircuit = dataclass_field(init=False)
+
+    def __post_init__(self) -> None:
+        self.value = QuantumCircuit._from_dict(self.d) if isinstance(self.d, dict) else self.d
+
+
+@dataclass
+class GraphSerial(SerialModel):
```
Roland-djee, line 124:
```suggestion
class RegisterSerial(SerialModel):
```

qadence/serialization.py
```diff
@@ -94,11 +198,11 @@ def serialize(obj: SUPPORTED_TYPES, save_params: bool = False) -> dict:
     """
     Supported Types:
 
-    AbstractBlock | QuantumCircuit | QuantumModel | TransformedModule | Register | Module
+    AbstractBlock | QuantumCircuit | QuantumModel | torch.nn.Module | Register | Module
     Serializes a qadence object to a dictionary.
 
     Arguments:
-        obj (AbstractBlock | QuantumCircuit | QuantumModel | Register | Module):
+        obj (AbstractBlock | QuantumCircuit | QuantumModel | Register | torch.nn.Module):
```
Roland-djee, line 205: Why replacing `TransformedModule` with torch ones ?

qadence/serialization.py
```diff
@@ -132,21 +236,24 @@ def serialize(obj: SUPPORTED_TYPES, save_params: bool = False) -> dict:
...
             d = {**expr_dict, **symb_dict}
-        elif isinstance(obj, (QuantumModel, QNN, TransformedModule)):
-            d = obj._to_dict(save_params)
-        elif isinstance(obj, torch.nn.Module):
-            d = {type(obj).__name__: obj.state_dict()}
         else:
-            d = obj._to_dict()
+            if hasattr(obj, "_to_dict"):
+                fn: Callable = obj._to_dict
+                d = fn(save_params) if isinstance(obj, torch.nn.Module) else fn()
```
Roland-djee, line 252: Maybe a more meaningful naming ?

Roland-djee (commented):
Thanks @Doomsk for this. Quick question, I don't see some tests for this. Do they exist already or the previous one work straight out of the box ? Minor comments from my side otherwise LGTM. One thing is that @awennersteen rewrites logging in Qadence here #366. Probably worth replacing some of your print statements with it.

awennersteen:
I expect that this might fail to deserialize old quantum models?

Could you verify that?

If so, could we wait to merge this until #366 is merged because I'd need that for the algorithm APIs, and once we change the serialization format it might take significant time to get the new format into production.

qadence/serialization.py
Doomsk: Indeed! Also, feel free to come up with a better name! I'm not particularly good at naming ^^"

qadence/serialization.py
Doomsk: It raises on the `ExpressionSerial`. This function is just for checking whether it parsed or not.

Getting rid of the `print` 😅

qadence/serialization.py
Doomsk: I will get rid of this one since it's not used anymore. The `SUPPORTED_TYPES` uses `TypingUnion`, should we replace for something else?

qadence/serialization.py
Doomsk: Makes sense! Although I used `GraphSerial` because the key it came from was `"graph"` (line 308)

qadence/serialization.py
Doomsk: Because we're getting rid of `QNN` and `TransformedModule` on `qadence`, so they cannot be explicitly referenced. `QuantumModel` and `torch.nn.Module` are the base classes for every new model added to the serialization (it was the plan from moving the models out of qadence).

Doomsk:
> I expect that this might fail to deserialize old quantum models?
> 
> Could you verify that?
> 
> If so, could we wait to merge this until #366 is merged because I'd need that for the algorithm APIs, and once we change the serialization format it might take significant time to get the new format into production.

It's the same serialization process, just rewritten. If something is working now with the current serialization will work with this new one. A next step should be to propose a better serialization logic and structure.

qadence/serialization.py
```diff
@@ -192,51 +300,18 @@ def deserialize(d: dict, as_torch: bool = False) -> SUPPORTED_TYPES:
...
-            obj = getattr(globals(), module_name)
-            obj.load_state_dict(d[module_name])
-        except Exception as e:
-            logger.error(
-                TypeError(
-                    f"{msg}. Unable to deserialize object due to {e}.\
-                    Supported objects are: {SUPPORTED_OBJECTS}"
-                )
-            )
-    return obj
```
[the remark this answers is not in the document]
Doomsk: Since the logic in that part has changed, it will log a different response. Putting that into `ModelSerial` and hope that is still relevant.

qadence/serialization.py
```diff
@@ -43,27 +47,127 @@ def file_extension(file: Path | str) -> str:
...
+
+
+THIS_PATH = os.path.dirname(__file__)
+GRAMMAR_FILE = os.path.join(THIS_PATH, "serial_expr_grammar.peg")
+
+
+def _parser_fn() -> ParserPEG:
+    with open(GRAMMAR_FILE, "r") as f:
+        grammar = f.read()
+    return ParserPEG(grammar, "Program")
```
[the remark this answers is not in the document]
Doomsk: just in case I placed a `lru_cache` and defined in a variable instead.

qadence/serialization.py
Roland-djee: True but since it concretises the graph into a `Register`, I'd be more specific.

pyproject.toml
```diff
@@ -44,6 +45,7 @@ dependencies = [
     "scipy",
     "pyqtorch==1.0.6",
     "matplotlib",
+    "Arpeggio==2.0.2",
```
dominikandreasseitz, line 48: does this repo require some binaries or can it just be pip-installed? would be nice to confirm that on the cluster so we can be sure it can be installed properly in the containers

qadence/serialization.py
Doomsk: Oh, I see! `init=False` is just saying that I don't want this attribute to be accessible as a parameter in the `__init__`
like:
```python
@dataclass
class X1:
    a: str
    b: str = field(init=False)
```
is equivalent to

```python
class X2:
    def __init__(self, a: str):
        self.a = a
        self.b: str
```

qadence/serialization.py
Roland-djee: OK makes sense. In that case, I suggest to import them:

```
from torch import nn.Modules
```

To make the type annotations more readable.

Roland-djee:
Hey @Doomsk do you have time to finalise this ?

Doomsk:
> Hey @Doomsk do you have time to finalise this ?

I'll do it tomorrow!

pyproject.toml
Doomsk: @dominikandreasseitz just confirming: there's not binary dependency; it can be pip-installed directly.

The only feature it contains external dependencies (optional), and that will not be used on cluster,  is the graphviz for the dot files.

qadence/serialization.py
Doomsk: I wanted to make explicit that `Module` is a `torch` object on the docs and type annotations by using `torch.nn.Module`, so we don't get confused by `qadence`'s own objects, but I can change that of course.

Doomsk:
can you take a look @dominikandreasseitz ?

dominikandreasseitz (approved):
thanks! lgtm

awennersteen (approved):
lgmt

Maybe the dataclasses in qadence/serialization.py could have some docstring , but I'm not convinced that it would add anything in particular so up to you.

PR #362, opened 2024-02-28
[Refac] Torch sample as no_grad

dominikandreasseitz:
- Move no_grad call  to backend pyq level
- Remove dead code from engine introduction

### DIFF
--- a/qadence/engines/torch/differentiable_expectation.py
+++ b/qadence/engines/torch/differentiable_expectation.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections import Counter, OrderedDict
+from collections import OrderedDict
 from dataclasses import dataclass
 from functools import partial
 from typing import Any, Callable, Sequence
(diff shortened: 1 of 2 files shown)


qadence/engines/torch/differentiable_expectation.py
```diff
@@ -232,160 +229,3 @@ def construct_rules(
                 # Since they are constants their gradients are 0.
                 param_to_psr[param_id] = lambda x: torch.tensor([0.0], requires_grad=False)
         return param_to_psr
-
-
-class DifferentiableBackend(Module):
-    """A class to abstract the operations done by the autodiff engine.
-
-    Arguments:
-        backend: An instance of the QuantumBackend type perform execution.
```
awennersteen, line 241: This one is now one level up right, so it was dead code?

awennersteen (approved):
lgtm

awennersteen:
Thanks!
I'm happy that I hadn't gone totally crazy trying to track the execution path

qadence/engines/torch/differentiable_expectation.py
dominikandreasseitz: we have a "abstract" DifferentiableBackend which is used by both jax and torch. so i guess some remnants werent properly cleaned up

dominikandreasseitz: resolving so i can merge

PR #377, opened 2024-03-26
[Proto] Implement a State result type.

### LINKED ISSUE #376: [Proto] Implement a State result type
Must allow the handling of:

- [x] State vectors as `Tensor`
- [x] Density matrices as `Tensor` for noisy simulations and functionalities in #367

Roland-djee:
Closes #376

### DIFF
--- a/qadence/types.py
+++ b/qadence/types.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import importlib
+from collections import Counter
+from dataclasses import dataclass
 from enum import Enum
 from typing import Callable, Iterable, Tuple, Union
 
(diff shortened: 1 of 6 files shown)


Roland-djee:
@jpmoutinho @dominikandreasseitz I know lint/test fail but works for PyQ. Just for you to tell me what you think about it.

Roland-djee:
@jpmoutinho @dominikandreasseitz I know lint/test fail but works for PyQ. Just for you to tell me what you think about it.

qadence/types.py
```diff
@@ -48,6 +50,27 @@
...
+class DensityMatrices:
+    density_matrix: ArrayLike
+
+
+@dataclass(frozen=True)
+class Samples:
+    samples: list[Counter]
+
+
+@dataclass(frozen=True)
```
dominikandreasseitz, line 68: what about expectation values? also, this would change the convenient torch-like interface quite a bit since you cant use autograd on outputs of qadence methods anymore

Roland-djee: I guess expectation values can stay a `Tensor` I guess ? I don't know of any user request to post-process expectation values yet.

Sure but that is the price to pay for having an intermediate `State` data structure. I know this is breaking but hopefully not too much.

Roland-djee: A probably less disruptive solution would be to subclass `Tensor`. WDYT @dominikandreasseitz ?

Roland-djee:
Closing as moved elsewhere.

PR #379, opened 2024-03-27
[Feature] Allow setting precision for pyqtorch backend

dominikandreasseitz:
- Moving to pyq v1.1.0 which supports precision setting
- Adding the option to move a pyqtorch-based QuantumModel to a desired precision
- `train_grad`  and `optimize_step` now accept a `dtype` arg 
- `data_to_device` now accepts args and kwargs like torch.to
-  New paragraph in `ml_tools.md` showcasing the passing of a `dtype` to `train_grad`


docs/qml/ml_tools.md
```diff
@@ -111,14 +116,16 @@ from qadence import hamiltonian_factory, hea, feature_map, chain
 from qadence.models import QNN
 from qadence.ml_tools import  TrainConfig, train_with_grad, to_dataloader
 
-n_qubits = 2
+DEVICE = torch.device('cuda') if torch.cuda.is_available() else torch.device('cpu')
```
Roland-djee, line 119: Would it make sense to showcase runtime or precisions comparisons with running the highest precision ?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -186,9 +188,10 @@ def forward(self, state: Tensor, _: dict[str, Tensor] = None) -> Tensor:
     def device(self) -> torch_device:
         return self._device
 
-    def to(self, device: torch_device) -> PyQMatrixBlock:
-        self.mat = self.mat.to(device)
-        self._device = device
+    def to(self, *args: Any, **kwargs: Any) -> PyQMatrixBlock:
```
Roland-djee, line 191: Can't we specialise `args` and `kwargs` type annotations ?

qadence/ml_tools/data.py
```diff
@@ -82,34 +82,34 @@ def to_dataloader(*tensors: Tensor, batch_size: int = 1, infinite: bool = False)
 
 
 @singledispatch
-def data_to_device(xs: Any, device: torch_device) -> Any:
+def data_to_device(xs: Any, *args: Any, **kwargs: Any) -> Any:
     """Utility method to move arbitrary data to 'device'."""
-    raise ValueError(f"Cannot move {type(xs)} to a pytorch device.")
+    raise ValueError(f"Unable to move {type(xs)} to {args}, {kwargs}.")
```
Roland-djee, line 87: It feels like since `Any` type is passed, this message could potentially be non-sensical.

Roland-djee (commented):
Thanks @dominikandreasseitz. I feel like using `args` and `kwargs` opens the door to too many vulnerabilities. I assume that `.to` supports a set of input parameters. We should probably target the relevant ones for the user and not letting any potentially breakable input in.

dominikandreasseitz:
> Thanks @dominikandreasseitz. I feel like using `args` and `kwargs` opens the door to too many vulnerabilities. I assume that `.to` supports a set of input parameters. We should probably target the relevant ones for the user and not letting any potentially breakable input in.

its exactly as the torch equivalent, see here https://pytorch.org/docs/stable/generated/torch.Tensor.to.html#torch.Tensor.to

Roland-djee:
> > Thanks @dominikandreasseitz. I feel like using `args` and `kwargs` opens the door to too many vulnerabilities. I assume that `.to` supports a set of input parameters. We should probably target the relevant ones for the user and not letting any potentially breakable input in.
> 
> @Roland-djee , its exactly as the torch equivalent, see here https://pytorch.org/docs/stable/generated/torch.Tensor.to.html#torch.Tensor.to and the reason for that is that Tensor.to infers from the arg what the user passed. for example, str or torch_device -> a device. This is very hard to properly type. if we want to offer the same flexibility as torch, we have to go for this imo

Makes sense. In a way, this is the user's responsibility to use `torch` correctly.

docs/qml/ml_tools.md
```diff
@@ -98,11 +98,16 @@ Let's see it in action with a simple example.
 
 ### Fitting a funtion with a QNN using `ml_tools`
 
+In Quantum Machine Learning, the general consensus is to use complex128 precision for states/ operators and float64 precision for parameters, which is also the convention which is used in `qadence`.
```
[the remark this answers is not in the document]
dominikandreasseitz: changed

docs/qml/ml_tools.md
```diff
@@ -98,11 +98,16 @@ Let's see it in action with a simple example.
...
+However, for specific usecases, using lower precision can greatly speed up training and reduce memory consumption. When using the `pyqtorch` backend, `qadence` offers the option to move a `QuantumModel` instance to a specific precision using the torch `to` syntax.
```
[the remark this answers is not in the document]
dominikandreasseitz: changed

docs/qml/ml_tools.md
dominikandreasseitz: i was thinking the same but i think its not the right place here, this .md file is for introducing the user to ml_tools and adding benchmarks would just overload them

qadence/backends/pyqtorch/convert_ops.py
dominikandreasseitz: i think its common practice to use these type annotations

qadence/ml_tools/data.py
dominikandreasseitz: but it will check the _actual_ type of `xs` and log that. so str, int etc

docs/qml/ml_tools.md
Roland-djee: OK maybe an example notebook ?

qadence/backends/pyqtorch/convert_ops.py
Roland-djee: Yeah, please ignore. Since we agreed that moving to a `torch` device and letting the possibility choose the precision is up to the user to do it correctly.

Roland-djee (approved):

qadence/ml_tools/data.py
dominikandreasseitz: good idea

pyproject.toml
```diff
@@ -20,7 +20,7 @@ authors = [
 ]
 requires-python = ">=3.9,<3.12"
```
awennersteen, line 21: I suggest we remove upper-bound pinning in qadence, but leave it in PyQ.


If you want to keep it here, then bump to <3.13

qadence/backends/utils.py
```diff
@@ -129,7 +130,7 @@ def is_pyq_shape(state: Tensor, n_qubits: int) -> bool:
 
 def validate_state(state: Tensor, n_qubits: int) -> None:
     """Check if a custom initial state conforms to the qadence or the pyqtorch format."""
-    if state.dtype != complex128:
+    if state.dtype not in [complex128, complex64]:
         raise TypeError(f"Expected type complex128, got {state.dtype}")
```
awennersteen, line 134:
```suggestion
        raise TypeError(f"Expected complex dtype, got {state.dtype}")
```

pyproject.toml
dominikandreasseitz: bumped

qadence/backends/utils.py
dominikandreasseitz: changed

awennersteen (approved):
lgtm

docs/qml/ml_tools.md
dominikandreasseitz: added an example

PR #381, opened 2024-04-03
[Performance] Remove validation

dominikandreasseitz:


qadence/backend.py
```diff
@@ -304,8 +280,7 @@ def run(
             A list of Counter objects where each key represents a bitstring
             and its value the number of times it has been sampled from the given wave function.
         """
-        validate_values_and_state(state, circuit.abstract.n_qubits, param_values)
```
jpmoutinho, line 307: Should we still keep the function in the `utils`? It's not used anywhere else, maybe add some comment there at least to be removed in the future if we end up not using it.

jpmoutinho (approved):

qadence/backend.py
dominikandreasseitz: i highly doubt it is used by anyone, it was never announced and it shouldve been marked as a private method. no need to bloat the utils imo

PR #384, opened 2024-04-09
[Bug] Partially fix GPSR for HamEvo

### LINKED ISSUE #397: [Refact] Reviewing block generators
Related to https://github.com/pasqal-io/qadence/issues/266

All unitary operators have an hermitian generator such that
$$\hat{U}=exp(-i\hat{H})$$
In Qadence we have `block.generator` which is supposed to return $\hat{H}$. However, if we want to use `HamEvo` to then represent the original unitary, we have to specify a time value since it implements $exp(-i\hat{H}t)$. Currently, some blocks seems to have some inconsistent definitions.

This also affects GPSR, since there the paper assumes a generator structure of $exp(-iG/2x)$ for some gate $U(x)$.

Probable requirements:
- Make all gates follow the GPSR format, as suggested in https://github.com/pasqal-io/qadence/issues/266
- Have a specific GPSR function to handle `HamEvo` differentiation, so that any arbitrary `HamEvo(G, x)` is processed in a way that the the $1/2$ factor is "considered" in $G$, and $x$ is an angle.

jpmoutinho:
Originally this MR was related to https://github.com/pasqal-io/qadence/issues/397

Decided to simplify it since properly making all other changes caused other issues.

### DIFF
--- a/qadence/backends/gpsr.py
+++ b/qadence/backends/gpsr.py
@@ -20,7 +20,7 @@ def general_psr(spectrum: Tensor, n_eqs: int | None = None, shift_prefac: float 
     sorted_unique_spectral_gaps = torch.tensor(list(sorted_unique_spectral_gaps))
 
     if n_eqs == 1:
-        return partial(single_gap_psr, spectral_gap=sorted_unique_spectral_gaps.item())
+        return single_gap_psr
     else:
         return partial(
             multi_gap_psr,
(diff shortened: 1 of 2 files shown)


Roland-djee:
Hey @jpmoutinho is this reviewable/mergeable ?

jpmoutinho:
Not yet @Roland-djee. I'm currently looking at the GPSR implementation to see how to properly account for that here.

Issue #382, opened 2024-04-05
[Refactoring] Rework TransformedModule


dominikandreasseitz:
@Roland-djee  @Doomsk @jpmoutinho 

Right now, the TransformedModule expects the user to pass floats/tensors to its constructor which are then held as buffers for input/output scaling and shifting. There are several issues which caused by this approach:

- Since we dont use torch pt for serialization,  we save scalings/shifting values in a very hacky way which makes it very hard to reinstantiate the same transformedmodule and get the same results
- the transformedmodule also uses a error-prone positional encoding of input scalings and shifting without enforcing the user to pass a list of featureparameters like the QNN. this can lead to scaling/shiftings applied to the wrong featureparameters

**Proposal 1:**

Fully remove the buffer-approach and instead require the user to pass values for the scalings and shiftings in the values dict:

```
class TransformedModule:
 ....

def expectation(self, values):
    values = self.transform_input(values)
    return values['output_scaling'] * self.model.expectation(values) + values['output_shifting']
```

This is also in line with the qadence philosophy of the values dict + clear separation of circuit/observable and parameters 

**Proposal 2:**

Fully remove TransformedModule and add utility methods to ml_tools like:

```
def transformed_expectation(model, values):
    values = transform_input(values)
    return values['output_scaling'] * model.expectation(values) + values['output_shifting']
```

The TM class is essentially 300 lines of code to accomplish the code snippet above.

let me know what you think

dominikandreasseitz:
@inafergra @smitchaudhary @gvelikova maybe you can give your input on that too

smitchaudhary:
+1 for the second option.

Not a huge fan of `TransformedModule` in the first place. Much better to nudge users towards using `feature_range` (and less importantly `target_range`) during QNN definition and using `transform` argument of `QNN` for output scaling/shifting.

So on that note, a helper function in `ml_tools`?

```python
def create_linear_transform(scaling: Tensor, shifting: Tensor) -> Callable[[Tensor], Tensor]:
    
    def transform(outputs):
        return scaling*outputs + shifting
    
    return transform
```

For the user to use like:

```python
...
QNN(circ, observable, inputs, transform = create_linear_transform())
```
This is maybe too many steps for a simple thing to do. But can help to move away from a clunky `TransformedModule` and also helps expose the `transform` functionality for more complex transformations.

dominikandreasseitz:
> +1 for the second option.
> 
[...]

PR #385, opened 2024-04-09
[Feature | Refac | Breaking] Rework QNN, Remove TransformedModule, Add Configs

dominikandreasseitz:
Closes https://github.com/pasqal-io/qadence/issues/382
Closes https://github.com/pasqal-io/qadence/issues/396

- [x] @smitchaudhary move https://github.com/pasqal-io/qadence-libs/pull/18/ into this MR
- [x] create a models/constructors.py containing the QNN constructor logic and the configs
- [x] add a classmethod to QNN which instantiates a QNN from configs only
- [x] remove "transform" arg from QNN constructor
- [x] Add a descriptive docstring to the new QNN
- [x] Modify the docs showcasing the new QNN capabilities


smitchaudhary:
Thanks. LGTM.

Treating both the input and output transforms on equal footing is much nicer.

qadence/models/qnn.py
```diff
@@ -15,11 +15,85 @@
...
+        'values' has to either be a Tensor when using torch.nn.Module, or a standard values dict.
+
+        Arguments:
+            values: A torch Tensor or a dict containing values for Featureparameters.
+        Returns:
+            A dict containing scaled and shifted FeatureParameters.
+        """
+        if isinstance(values, Tensor):
+            values = format_to_dict(values)
+        values = {
```
dominikandreasseitz, line 87: @smitchaudhary i changed the input scaling/shifting requirements to ALWAYS be 2D

qadence/models/qnn.py
```diff
@@ -195,34 +277,16 @@ def expectation(
...
-        corresponds to the number of input features of the QNN
-        """
-        # for backwards compat...
-        if isinstance(values, dict):
-            return values
-
-        if len(values.size()) == 1:
-            values = values.reshape(-1, 1)
-        msg = f"Model expects in_features={self.in_features} but got {values.size()[1]}."
-        assert len(values.size()) == 2, msg
```
dominikandreasseitz, line 221: lets remove these asserts which are executed on every forward call @awennersteen

awennersteen: yes, please

tests/models/test_qnn.py
```diff
@@ -240,3 +238,67 @@ def compute_state_manually(xs: torch.Tensor) -> torch.Tensor:
...
+    model = QNN(
+        circuit,
+        observable,
+        backend=BackendName.PYQTORCH,
+        diff_mode=DiffMode.AD,
+        inputs=inputs,
+        transform_input=input_transform,
+        transform_output=output_transform,
+    )
+    return model
```
awennersteen, line 273: conftest fixtures?

smitchaudhary:
Maybe this is extremely nit-pciky but on the usage side, would it make sense to clarify the docstring [here](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/feature_maps.py#L160) in `featue_map`? As since now there are two ways of transforming the inputs, there is a risk of double transform.

```
        feature_range: range of data that the input data provided comes from. If a range is provided, it gets linearly transformed to the domain of the feature map used. See `target_range` below. If None is provided, no transformation happens.
```

Or something similar?

jpmoutinho:
@smitchaudhary I guess a question is if it's worth keeping both options. I guess the QNN option is only relevant if someone uses their own custom feature map right?

smitchaudhary:
> I guess the QNN option is only relevant if someone uses their own custom feature map right?
@jpmoutinho 
I think so. Or if someone for some reason wants some fancier transformation.

smitchaudhary:
@dominikandreasseitz I have changed the observable config and added the functionality to automatically scale the output This happens in a manner similar that is similar to what happens with the feature maps, that is, the user provides the range, and the scaling and the shiftings are calculated automatically (according to what is the natural range of the detuning).

In addition, there is the option to also make these transformations trainable. If that is the case, then the scaling and shifting values are initialized to appropriate values, but then training will take over after that of course.

jpmoutinho:
> @dominikandreasseitz I have changed the observable config and added the functionality to automatically scale the output This happens in a manner similar that is similar to what happens with the feature maps, that is, the user provides the range, and the scaling and the shiftings are calculated automatically (according to what is the natural range of the detuning).
> 
> In addition, there is the option to also make these transformations trainable. If that is the case, then the scaling and shifting values are initialized to appropriate values, but then training will take over after that of course.

I looked at the functions for the output transformation, a few thoughts:
- Are you considering only the detuning because otherwise it's harder to know the natural range of the observable?
- Either way, I think it's still hard if the observable is not diagonal, e.g. `detuning = X` or `Y`.
- The natural range of an observable is bounded by the minimum and maximum eigenvalues, so to make it fully general and work the same way the `feature_map` does it would probably require computing the spectrum each time, but that would be very intensive for larger qubit numbers.
- Given that, one option would be to only allow getting the natural range for diagonal observables (but still allow multi-qubit terms). There is a private method `block._is_diag_pauli` that checks if a block is only composed of `I`, `Z` or `N`.
- Another option would be to _not_ make it work like the `feature_map` and the user can only pass a scaling / shifting constant, or set it to be trainable.

Not sure which option is best :)

smitchaudhary:
@jpmoutinho 

> Another option would be to not make it work like the feature_map and the user can only pass a scaling / shifting constant, or set it to be trainable.

This seems like the most straightforward option to me. I did consider this too but went the way I did to keep the interface similar to the `feature_map`, which is not necessary of course.

> Are you considering only the detuning because otherwise it's harder to know the natural range of the observable?

Indeed that is the reason. I agree it is an unnecessary restriction on the kind of observables that one can have.

> The natural range of an observable is bounded by the minimum and maximum eigenvalues, so to make it fully general and work the same way the feature_map does it would probably require computing the spectrum each time, but that would be very intensive for larger qubit numbers.

Currently, the `hamiltonian_factory` only allows for 4 detunings, namely, `X`, `Y`, `Z` and `N`. And I think we know the spectrum in all 4 cases. Right? It is [-1, 1] for the 3 Pauli types and [0, 1] for `N`. Do I have this wrong? This is why the different scaling and shifting parameters for what kind of detuning is used.

Though overall, I think this is imposing many restrictions which might not be necessary. Can also simply ask the user for the scaling and shifting parameters.

smitchaudhary:
Another side effect of going this way is that now, the scaling is basically applied as the detuning strength of the first qubit, and the rest have the default strength of 1.0

This was necessary because if we were to apply the scaling to all the qubits, the range would be scaled by `scaling^n_qubits`. A larger range is not a problem except in the case where `scaling` < 1.

So even in the case of the user explicitly specifying the shift and the scaling values, do you suggest we keep it the same way as here and apply the scaling to the first qubit (pretty arbitrary) or do something else?

jpmoutinho:
@smitchaudhary

Indeed you are correct if we only keep single-qubit terms then `X` or `Y` is still ok, the range is the same as `Z`.

I don't understand the need to scale only the first qubit though, I think it should be a factor multiplying the whole observable, but we can talk about that today :)

smitchaudhary:
@jpmoutinho 

> I don't understand the need to scale only the first qubit though, I think it should be a factor multiplying the whole observable, but we can talk about that today :)

Yeah that seems right. Was just an artifact of the previous code where this scaling was done through specifying `detuning_strength`. A multiplication factor is a lot more straightforward.

qadence/models/qnn_from_config.py
```diff
@@ -0,0 +1,743 @@
...
+
+
+def _global_identity(register: int | Register) -> KronBlock:
+    """Create a global identity block."""
+    return kron(
+        *[I(i) for i in range(register if isinstance(register, int) else register.n_qubits)]
+    )
+
+
+def _create_observable(
```
smitchaudhary, line 663: @dominikandreasseitz this function here creates the required observable. Has the global identity to help it shift and then the appropriate scaling.

If no range is provided, then no transformation occurs.

Roland-djee:
Hey @dominikandreasseitz @smitchaudhary what's the status on this ?

dominikandreasseitz:
> Hey @dominikandreasseitz @smitchaudhary what's the status on this ?

@Roland-djee  in progress, its actually quite a bit of files to go through. i am creating tests atm

Roland-djee:
> > Hey @dominikandreasseitz @smitchaudhary what's the status on this ?
> 
> @Roland-djee in progress, its actually quite a bit of files to go through. i am creating tests atm

Nice thanks for the heads-up.

jpmoutinho:
@smitchaudhary thanks, and sorry for not commenting on this before.

Generally I agree with what you said, mainly to have it behave like the current defaults for the contructors.

I would suggest changing `num_layers` to `depth` just to be consistent with the previous `hea` syntax.

I never liked the default feature map param being called `phi` 😅. I guess changing that to `x` would technically break backwards compatibility for some codes 🤔. We can do it later... But I think the best for the QNN inputs would be to have no defaults right? If a user defines a multi-dimensional feature map they will have to provide different names anyway, so they might as well pass it into the QNN.

smitchaudhary:
A quick overview of the changes made:

1. The defaults are now set as discussed above.
2. Various properties of the feature map, can now be provided as a dictionary instead of a list where the order of the elements was responsible for applying it to the correct feature. For example, earlier, the workflow would look like:
```python
fm_config = FeatureMapConfig(
    num_features=2,
    feature_range=[ (0., 1.), (-1., 1.) ]
)
```
The same would now look like:
```python
fm_config = FeatureMapConfig(
    num_features=2,
    inputs=('t', 'x')
    feature_range={
        't': (0., 1.),
        'x':(-1., 1.)
    }
)
```

Since now, for more than 1 input features, the user is required to give the `inputs` list anyways, so much less error prone for the user to also specify the range (or any other property such as basis set, reupload scaling etc) explicitly.

Thoughts?

jpmoutinho:
Thanks @smitchaudhary, I like the dict option!

smitchaudhary:
Hello, a quick small change made to the `ObservableConfig`. Now, it no longer takes the number of qubits in the input but uses the `register` variable passed to the function that creates qnn. Why so?

1. Consistent with how the other configs are defined.
2. As we are creating this QNN from the configs, the register is fixed already. No? Why provide it in two different ways?
3. For the observable `hamiltonian_factory` can take a `Register` object, thus being able to handle arbitrary configurations with arbitrary distances (and thus an interaction accordingly), restricting `ObservableConfig` to a single `n_qubits` is taking away this possibility when creating analog circuits.

And also add a small example in docstring.

qadence/ml_tools/models.py
```diff
@@ -342,11 +341,3 @@ def _from_dict(cls, d: dict, as_torch: bool = False) -> QNN:
             logger.warning(f"Unable to deserialize object {d} to {cls.__name__} due to {e}.")
 
         return qnn
-
-    @property
-    def device(self) -> torch.device:
```
dominikandreasseitz, line 315: why remove that?

qadence/ml_tools/config.py
```diff
@@ -60,3 +71,262 @@ def __post_init__(self) -> None:
...
+    """
+
+    inputs: list[Basic | str] | None = None
+    """
+    List that indicates the order of variables of the tensors that are passed.
+
+    Optional if a single feature is being encoded, required otherwise. Given input tensors
+    `xs = torch.rand(batch_size, input_size:=2)` a QNN with `inputs=("t", "x")` will
+    assign `t, x = xs[:,0], xs[:,1]`.
+    """
```
debrevitatevitae, line 185: Shouldn't the type of `inputs` be a tuple, rather than a list? This would agree with the types of most other attributes of this config. Also, the docstring here also hints to define `inputs` as a tuple.

debrevitatevitae (changes requested):

qadence/ml_tools/config.py
smitchaudhary: I would be in favor of correcting the docstrings instead. See [this](https://github.com/pasqal-io/qadence/blob/ds/remove_tm/qadence/model.py#L100) or any other such time in `QNN` or `QuantumModel` where the `inputs` are in a list, so just trying to keep it consistent with that.

debrevitatevitae: I see, and it would be out of the scope of this PR to do a refactoring at that level. Still, QNN inputs and such look like hashable things to me. Maybe it should be considered for the future? @dominikandreasseitz

debrevitatevitae (approved):

qadence/draw/utils.py
```diff
@@ -10,6 +10,7 @@
 
 import sympy
 
+from qadence import QuantumModel
```
jpmoutinho, line 13:
```suggestion
from qadence.model import QuantumModel
```

qadence/ml_tools/config.py
```diff
@@ -70,3 +81,265 @@ def __post_init__(self) -> None:
...
+    "hea" for Hardware Efficient Ansatz.
+    "iia" for Identity intialized Ansatz.
+    """
+
+    ansatz_strategy: str = "digital"
+    """Ansatz strategy.
+
+    "digital" for fully digital ansatz. Required if `ansatz_type` is `iia`.
+    "sdaqc" for analog entangling block.
+    "rydberg" for fully rydberg hea ansatz.
```
jpmoutinho, line 296: Maybe you should just add Rydberg to the `Strategy` `StrEnum` and use that everywhere that you have "digital", "sdaqc", "analog".

tests/models/test_qnn.py
dominikandreasseitz: resolving since outdated

qadence/ml_tools/constructors.py
jpmoutinho, line 1: I did not go through this file in detail now, let me know if I should 😅

qadence/model.py
jpmoutinho, line 1: I guess we'll have to mark this as a breaking change (but I agree with it)

jpmoutinho (approved):
@smitchaudhary @dominikandreasseitz amazing, thanks a lot for this huge work 🙏

qadence/backends/utils.py
```diff
@@ -152,8 +152,45 @@ def infer_batchsize(param_values: ParamDictType = None) -> int:
...
+    eps: float = None,
+) -> Tensor:
+    """
+    Arguments:
+
+        f: Function to differentiate
+        x: Input of shape `(batch_size, input_size)`
+        derivative_indices: which *input* to differentiate (i.e. which variable x[:,i])
+        eps: finite difference spacing (uses `torch.finfo(x.dtype).eps ** (1 / (2 + order))` as a
+            default)
```
Roland-djee, line 168: Return ?

qadence/constructors/hamiltonians.py
```diff
@@ -229,3 +230,40 @@ def ising_hamiltonian(
...
+    """
+    Whether to have a trainable transformation on the output of the observable.
+
+    If None, the scale and shift are numbers.
+    If True, the scale and shift are VariationalParameter.
+    If False, the scale and shift are FeatureParameter.
+    """
+
+    def __post_init__(self) -> None:
+        if type(self.scale) in (int, float, complex, np.int64, np.float64) and type(self.shift) in (
```
Roland-djee, line 259: Maybe create some `is_numerical` predicate ?

qadence/ml_tools/config.py
```diff
@@ -70,3 +81,265 @@ def __post_init__(self) -> None:
...
+    Give a single BasisSet to use the same for all features.
+    Give a dict of (str, BasisSet) where the key is the name of the variable and the
+    value is the BasisSet to use for encoding that feature.
+    BasisSet.FOURIER for Fourier encoding.
+    BasisSet.CHEBYSHEV for Chebyshev encoding.
+    """
+
+    reupload_scaling: ReuploadScaling | dict[str, ReuploadScaling] = ReuploadScaling.CONSTANT
+    """
+    Scaling for encoding the same feature on different qubits in the.
```
Roland-djee, line 105: `pydocstringformatter` strikes again.

qadence/model.py
```diff
@@ -258,9 +258,9 @@ def _to_dict(self, save_params: bool = False) -> dict[str, Any]:
                 "observable": abs_obs,
                 "backend": self._backend_name,
                 "diff_mode": self._diff_mode,
-                "measurement": self._measurement._to_dict()
-                if self._measurement is not None
-                else {},
+                "measurement": (
+                    self._measurement._to_dict() if self._measurement is not None else {}
```
Roland-djee, line 262:
```suggestion
                    self._measurement._to_dict() if self._measurement is not None else dict()
```

qadence/types.py
```diff
@@ -401,3 +401,21 @@ class ReadOutOptimization(StrEnum):
...
+    inputs of UFAs.
+    """
+
+    AD = "ad"
+    """Reverse automatic differentiation."""
+    FD = "fd"
+    """Central finite differencing."""
+
+
+class TObservableTransform:
```
Roland-djee, line 418:
```suggestion
class ObservableTransform:
```

tests/ml_tools/test_qnn.py
```diff
@@ -240,3 +242,166 @@ def compute_state_manually(xs: torch.Tensor) -> torch.Tensor:
...
+        observable,
+        backend=BackendName.PYQTORCH,
+        diff_mode=DiffMode.AD,
+        inputs=inputs,
+    )
+    return model
+
+
+@pytest.mark.parametrize("output_range", [(1.0, 0.0, False), (2.0, 0.0, None), (3.0, 1.0, False)])
+@pytest.mark.parametrize("batch_size", [1])
```
Roland-djee, line 282: Does it make sense to parametrise for a single value ?

tests/ml_tools/test_qnn.py
```diff
@@ -240,3 +242,166 @@ def compute_state_manually(xs: torch.Tensor) -> torch.Tensor:
         inputs=["y", "x"],
     )
     assert not torch.allclose(ys, model.run(xs))
+
+
+def quantum_circuit(n_qubits: int = 2, depth: int = 1) -> QuantumCircuit:
```
Roland-djee, line 247: Maybe add them as fixtures ?

Roland-djee (approved):
Thanks a lot @dominikandreasseitz looks very clean. Minor cosmetic comments.

docs/tutorials/qml/config_qnn.md
```diff
@@ -0,0 +1,111 @@
...
+    ansatz_type="hea",
+    ansatz_strategy="digital",
+)
+
+ansatz = create_ansatz(register=4, config=ansatz_config)
+
+print(html_string(ansatz)) # markdown-exec: hide
+```
+
+We have specified that the ansatz should have a depth of 2, and the ansatz type is "hea" (Hardware Efficient Ansatz). The ansatz strategy is set to "digital", which means digital gates are being used. One could alternatively use "analog" or "rydberg" as the ansatz strategy.
```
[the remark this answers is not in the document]
smitchaudhary: This was to get around the fact that "Rydberg" did not exist as a strategy. But as @jpmoutinho suggested above, I will add that and make the changes accordingly.

It is just that "Rydberg" isn't strictly a "strategy" like sDAQC, bDAQC and so on are currently in the StrEnum. But since there is a use for it, I could just go ahead and add it.

docs/tutorials/qml/config_qnn.md
```diff
@@ -0,0 +1,111 @@
...
+
+Let us say, we want to follow this feature map with 2 layers of hardware efficient ansatz.
+
+```python exec="on" source="material-block" session="config" html="1"
+from qadence.ml_tools.config import AnsatzConfig
+from qadence.ml_tools.constructors import create_ansatz
+
+ansatz_config = AnsatzConfig(
+    depth=2,
+    ansatz_type="hea",
```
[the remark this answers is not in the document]
smitchaudhary: We don't as of now, but will add it now. Much cleaner.

qadence/ml_tools/config.py
smitchaudhary: On it.

docs/tutorials/qml/config_qnn.md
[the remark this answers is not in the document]
jpmoutinho: Yes I thought a bit whether to suggest that or not because like you said it's a bit against what the initial purpose of the strategy enum was. But yeah for now it's useful so let's just add it.

qadence/ml_tools/constructors.py
smitchaudhary: It is a newly added file, so another look could be helpful. If you are satisfied, this PR is ready for merge.

dominikandreasseitz: we are

PR #390, opened 2024-04-15
[Docs] New docs home page

jpmoutinho:


jpmoutinho:
@dominikandreasseitz @Roland-djee I changed the title of the "getting started" tab but didn't really change the contents, just removed some info from the start to put in the Home page. I would focus this MR on the Home page, and let me know what else to add there, and I will do more changes to the getting started tab in a follow-up MR.

Roland-djee (approved):
Thanks @jpmoutinho, not much to say appart looking good :)

Roland-djee:
> @dominikandreasseitz @Roland-djee I changed the title of the "getting started" tab but didn't really change the contents, just removed some info from the start to put in the Home page. I would focus this MR on the Home page, and let me know what else to add there, and I will do more changes to the getting started tab in a follow-up MR.

Sure. Once we're OK with the basic structure, we can modify it at will.

PR #391, opened 2024-04-16
[Docs] File reorganization

jpmoutinho:
Trying not to make changes to the files here, just reorganizing them into different tabs


Roland-djee (approved):
Looking nice :)

PR #393, opened 2024-04-17
[Feature] Time-dependent generator proposal

### LINKED ISSUE #395: [Feature] Time dependent generator
Initial proposal of time-dependent generator block.

Code to test some capabilities of current implementation:

```
from qadence.blocks.time_block import TDGenerator
from qadence.parameters import Parameter
from pprint import pprint

# create parametrized blocks
t = Parameter("t")
block1 = tag(t * X(0), "block1")
block2 = tag((t**2) * Y(1), "block2")

# create time-dependent generator objects
td_gen1 = TDGenerator.from_block(100, block1)
td_gen2 = TDGenerator.from_block(200, block2)

# perform chain operation on time-dependent generator objects
td_gen3 = td_gen1 * td_gen2 * td_gen1

pprint(td_gen3.schedule, sort_dicts=False)

print("actual generator matrix:")
print(td_gen3(torch.tensor(0.15)))
```

As it can be seen from this code snippet, the `TDGenerator` object can be called with a time argument that results in a tensor representing the corresponding generator at some time `t`. Now it is easy to pass this `TDGenerator` object to `dynamiqs` or `krylov` solvers to actually get the time evolution of the system. The main questions/suggestions now are the following:

1. `HamEvo` can be modified to accept `TDGenerator` objects as generator and then in the backend it can be checked whether `pyq` or Schrodinger equation solver must handle the simulation. In this way `HamEvo` act as a placeholder that potentially can store time-dependent blocks.
2. Do we allow TDGenerator objects to be part of the circuit on their own, without being encapsulated in `HamEvo`?
3. Do we create a separate backend for Schrodinger equation solvers or integrate them into `pyq`?
4. What is the place of `TDGenerator` in the hierarchy of blocks, e.g., should it inherit from `AbstractBlock` or be completely standalone?

vytautas-a:
Closes #395


qadence/blocks/time_block.py
```diff
@@ -8,73 +9,107 @@
...
-                "fn": None,
+                "generator": block,
                 "qubit_support": block.qubit_support,
             }
-        ]
+        }
         return cls(schedule)
 
+    def _create_new_schedule(self, op: str, other: TDGenerator) -> dict[int, dict]:
+        if op in ["add", "kron"]:
```
Roland-djee, line 36: Shouldn't that be in the operator overloads ?

qadence/blocks/time_block.py
```diff
@@ -8,73 +9,107 @@
...
+
+                new_slot["generator"] = add(*blocks_to_add) if op == "add" else kron(*blocks_to_add)  # type: ignore [assignment]
+
+                # calculate qubit support of current slot
+                qubit_support = tuple(
+                    sorted(set(new_slot["generator"].qubit_support).union(set(qubit_support)))  # type: ignore [attr-defined]
+                )
+                new_slot["qubit_support"] = qubit_support  # type: ignore [assignment]
+
+        elif op == "chain":
```
Roland-djee, line 71: Same ?

Roland-djee (commented):
Thanks @vytautas-a for this. Not much to say now, merely suggestions.

Roland-djee:
> 1. `HamEvo` can be modified to accept `TDGenerator` objects as generator and then in the backend it can be checked whether `pyq` or Schrodinger equation solver must handle the simulation. In this way `HamEvo` act as a placeholder that potentially can store time-dependent blocks.

Wouldn't it be easier to just consider time-dependence as a parameter ? And we can pass values just like any other parameter currently ? Ie. reuse the `TimeEvolutionBlock` ?

> 2. Do we allow TDGenerator objects to be part of the circuit on their own, without being encapsulated in `HamEvo`?

Not sure. It would make for non-unitary computations right ?

> 3. Do we create a separate backend for Schrodinger equation solvers or integrate them into `pyq`?

I would rather integrate it in `PyQ` because: 1) we've got that elsewhere and 2) it would make a nice addition to PyQ and make it a complete DA backend. But this needs to be thoroughly discussed.

> 4. What is the place of `TDGenerator` in the hierarchy of blocks, e.g., should it inherit from `AbstractBlock` or be completely standalone?

If we follow the `TimeEvolutionBlock` route, it naturally inherits from `AbstractBlock`.

qadence/blocks/time_block.py
vytautas-a: There are overloads - `__mul__`, `__add__` and `__matmul__` are implemented just below. I just wanted to avoid duplication of code in case of `__matmul__` and `__add__` since the only difference is in one line.

vytautas-a:
`TDGenerator` as a class is mainly for encapsulating schedule of generators - essentially, when to use what blocks. But underlying blocks are parametrized blocks already now. 

`TimeEvolutionBlock` if I understand correctly, currently is just some kind of a placeholder class that doesn't have it's own functionality. I could move all the stuff from my `TDGenerator` to `TimeEvolutionBlock`.

jpmoutinho:
Thanks @vytautas-a! This is a good start!

> Do we create a separate backend for Schrodinger equation solvers or integrate them into pyq?

I also think integrating something into PyQ would be the best.

> HamEvo can be modified to accept TDGenerator objects as generator and then in the backend it can be checked whether pyq or Schrodinger equation solver must handle the simulation. In this way HamEvo act as a placeholder that potentially can store time-dependent blocks.

> Do we allow TDGenerator objects to be part of the circuit on their own, without being encapsulated in HamEvo?

> What is the place of TDGenerator in the hierarchy of blocks, e.g., should it inherit from AbstractBlock or be completely standalone?

I have a few thoughts on this but we can discuss it in a call :)

vytautas-a:
One more thought: if we want to support mixing the new time-dependent `HamEvo` block with ordinary blocks in the same circuit, it would make sense to all the blocks derived from `HamEvo` and assign them a constant generator with some default "duration". This way all the blocks would have proper time treatment, however if the `HamEvo` generator is found to be constant the usual `pyq` simulation would take place for these blocks. Thus, we would use ODE solver only where it is needed.

Roland-djee:
> One more thought: if we want to support mixing the new time-dependent `HamEvo` block with ordinary blocks in the same circuit, it would make sense to all the blocks derived from `HamEvo` and assign them a constant generator with some default "duration". This way all the blocks would have proper time treatment, however if the `HamEvo` generator is found to be constant the usual `pyq` simulation would take place for these blocks. Thus, we would use ODE solver only where it is needed.

True, although I think this should be set when instantiating the block right ? Say, `X(0)` would implicitly set the generator time parameter to be the right phase to realise the unitary, which might also be backend dependent. Or am I missing something here ?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -303,6 +307,10 @@ def __init__(
         self.block = block
         self.hmat: Tensor
 
+        print("here")
```
jpmoutinho, line 310: 👋

qadence/parameters.py
```diff
@@ -168,6 +172,14 @@ def _from_dict(cls, d: dict) -> Parameter | None:
             return None
 
 
+# class TimeParameter(Parameter):
```
Roland-djee, line 175: Remove.

tests/backends/pyq/test_time_dependent_generator.py
```diff
@@ -0,0 +1,71 @@
...
+    QuantumCircuit,
+    QuantumModel,
+    Register,
+    TimeParameter,
+    X,
+    Y,
+)
+
+
+@pytest.fixture
```
Roland-djee, line 24: I guess you can put all these fixtures in `conftest.py`.

tests/backends/pyq/test_time_dependent_generator.py
```diff
@@ -0,0 +1,71 @@
...
+    def generator_t(t: float, args: Any) -> qutip.Qobj:
+        return omega * (
+            t * qutip.tensor(qutip.sigmax(), qutip.qeye(2))
+            + feature_param_value * t**2 * qutip.tensor(qutip.qeye(2), qutip.sigmay())
+        )
+
+    return generator_t
+
+
+def test_time_dependent_generator(
```
Roland-djee, line 53: It would be good to expand tests to a randomly generated generator if possible.

Roland-djee (commented):
Hey @vytautas-a thanks. This is already in good shape. More comments from my side.

qadence/backends/pyqtorch/backend.py
```diff
@@ -73,9 +72,9 @@ def observable(self, observable: AbstractBlock, n_qubits: int) -> ConvertedObser
             lambda block: chain_single_qubit_ops(block)
             if self.config.use_single_qubit_composition
             else flatten(block),
-            scale_primitive_blocks_only,
+            # scale_primitive_blocks_only,
```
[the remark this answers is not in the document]
vytautas-a: Yes, this shouldn't be commented - it's left from our little session of brainstorming with Dominik :)

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -8,6 +8,10 @@
 
 import pyqtorch as pyq
 import sympy
+import torch
+
+# from pulser_diff.dq.sesolve import sesolve as sesolve_dq  # import dynamiqs solver
```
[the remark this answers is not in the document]
vytautas-a: I left this for reminding that the dynamiqs solver has to be imported somehow as well

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -417,13 +422,74 @@ def forward(
...
+        if self.block.generator.is_time_dependent:  # type: ignore [union-attr]
+
+            def Ht(t: Tensor | float) -> Tensor:
+                # values dict has to change with new value of t
+                # initial value of a feature parameter inside generator block
+                # has to be inferred here
+                new_vals = {}
+                for str_expr, val in values.items():
+                    expr = sympy.sympify(str_expr)
+                    t_symb = sympy.Symbol("t")
```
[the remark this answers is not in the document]
vytautas-a: Yes, the time parameter in this implementation **must** be named "t". This is again due to the embedding stuff.

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -417,13 +422,74 @@ def forward(
...
-            self.n_qubits,
-            self.batch_size,
-        )
+        if self.block.generator.is_time_dependent:  # type: ignore [union-attr]
+
+            def Ht(t: Tensor | float) -> Tensor:
+                # values dict has to change with new value of t
+                # initial value of a feature parameter inside generator block
+                # has to be inferred here
+                new_vals = {}
```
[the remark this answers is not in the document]
vytautas-a: why?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -417,13 +422,74 @@ def forward(
...
+                for str_expr, val in values.items():
+                    expr = sympy.sympify(str_expr)
+                    t_symb = sympy.Symbol("t")
+                    free_symbols = expr.free_symbols
+                    if t_symb in free_symbols:
+                        # expression has time dependence
+                        if len(free_symbols) == 2:
+                            # expression has a single additional feature parameter
+                            feat_symb = free_symbols.difference(set([t_symb])).pop()
```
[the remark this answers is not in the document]
vytautas-a: Because I can't know the original value of some feature parameter sitting inside my time-dependent generator. Again, due to embedding stuff. So I have to solve for original value of that feature parameter.

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -417,13 +422,74 @@ def forward(
...
+                            feat_val = sympy.nsolve(
+                                expr.subs(t_symb, 1.0) - float(val), feat_symb, 1.0
+                            )
+                            new_vals[str_expr] = torch.tensor(
+                                float(expr.subs([(t_symb, t), (feat_symb, feat_val)]))
+                            )
+                        elif len(free_symbols) == 1:
+                            # expression contains only time parameters
+                            new_vals[str_expr] = torch.tensor(float(expr.subs(t_symb, t)))
```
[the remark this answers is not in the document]
vytautas-a: Actually not really.

tests/backends/pyq/test_time_dependent_generator.py
vytautas-a: Regarding this - it might be not that easy to do that since I'll have to also create some kind of parser that creates a Hamiltonian conforming to `qutip`'s requirements. It's possible but it's a bit time consuming.

qadence/backends/pyqtorch/config.py
```diff
@@ -4,6 +4,8 @@
 from logging import getLogger
 from typing import Callable
 
+from pasqal_solvers.utils import SolverType
```
Roland-djee, line 7:
```suggestion
from pasqal_solvers.utils import Solver
```

qadence/backends/pyqtorch/convert_ops.py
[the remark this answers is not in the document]
Roland-djee: Not sure I follow there.

qadence/backends/pyqtorch/convert_ops.py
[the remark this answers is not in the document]
Roland-djee: Because it is less ambiguous. `{}` could either be a set initialisation.

qadence/backends/pyqtorch/convert_ops.py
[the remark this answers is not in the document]
Roland-djee: But this make it less flexible right ? Or is it for internal use only and is never exposed to the user ?

qadence/blocks/abstract.py
```diff
@@ -300,6 +300,13 @@ def is_parametric(self) -> bool:
         params: list[sympy.Basic] = parameters(self)
         return any(isinstance(p, Parameter) for p in params)
 
+    @property
+    def is_time_dependent(self) -> bool:
+        from qadence.blocks.utils import parameters
+
+        params: list[sympy.Basic] = parameters(self)
+        return any(getattr(p, "is_time", False) for p in params)
```
Roland-djee, line 308:
```suggestion
        return any(p.is_time for p in params)
```

qadence/blocks/abstract.py
```diff
@@ -300,6 +300,13 @@ def is_parametric(self) -> bool:
...
+
     def tensor(self, values: dict[str, TNumber | torch.Tensor] = {}) -> torch.Tensor:
```
Roland-djee, line 310:
```suggestion
    def tensor(self, values: dict[str, TNumber | torch.Tensor] = dict()) -> torch.Tensor:
```

qadence/blocks/embedding.py
```diff
@@ -137,9 +140,12 @@ def embedding_fn(params: ParamDictType, inputs: ParamDictType) -> ParamDictType:
             gate_lvl_params: ParamDictType = {}
```
Roland-djee, line 140:
```suggestion
            gate_lvl_params: ParamDictType = dict()
```

qadence/models/quantum_model.py
```diff
@@ -171,7 +171,9 @@ def run(
     ) -> Tensor:
         if values is None:
             values = {}
```
Roland-djee, line 173:
```suggestion
            values = dict()
```

qadence/operations/ham_evo.py
```diff
@@ -70,6 +71,7 @@ def __init__(
         generator: Union[TGenerator, AbstractBlock],
         parameter: TParameter,
         qubit_support: tuple[int, ...] = None,
+        duration: float | None = None,
     ):
         gen_exprs = {}
```
Roland-djee, line 76:
```suggestion
        gen_exprs = dict()
```

tests/backends/pyq/test_time_dependent_generator.py
Roland-djee: OK got it.

Roland-djee (commented):
Thanks @vytautas-a. Some more comments from me.

qadence/backends/pyqtorch/convert_ops.py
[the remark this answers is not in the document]
vytautas-a: Actually you can't initialize an empty set using `{}` - it always produces a `dict`.

qadence/backends/pyqtorch/convert_ops.py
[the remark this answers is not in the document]
vytautas-a: Fixed that. Now any time parameter name is allowed.

tests/qadence/test_quantum_model.py
```diff
@@ -44,7 +44,9 @@ def test_quantum_model_parameters(parametric_circuit: QuantumCircuit) -> None:
     assert len([i for i in model_psr.parameters()]) == 4
     assert len([i for i in model_ad.parameters()]) == 4
     embedded_params_psr = model_psr.embedding_fn(model_psr._params, {"x": torch.rand(1)})
+    embedded_params_psr.pop("orig_param_values", {})
```
dominikandreasseitz, line 47: can you add a note here why we do this and a TODO to remove it again

qadence/backends/pyqtorch/config.py
```diff
@@ -4,6 +4,8 @@
 from logging import getLogger
 from typing import Callable
 
+from pyqtorch.utils import SolverType
+
```
jpmoutinho, line 8: For a user this should also be importable `from qadence import SolverType` and `from qadence.types import SolverType`. I think you can do that by importing it in the `qadence.types.py` and adding it to the `__all__` defined there.

qadence/operations/ham_evo.py
```diff
@@ -43,6 +43,7 @@ class HamEvo(TimeEvolutionBlock):
         generator: Either a AbstractBlock, torch.Tensor or numpy.ndarray.
         parameter: A scalar or vector of numeric or torch.Tensor type.
         qubit_support: The qubits on which the evolution will be performed on.
+        duration: duration of evolution in ns in case of time-dependent generator
```
jpmoutinho, line 46: This is no longer specifically in ns right?

tests/qadence/test_quantum_model.py
jpmoutinho: Is this still needed with the new pyqtorch version? I thought that fixed this?

qadence/operations/ham_evo.py
vytautas-a: not in ns, correct

qadence/backends/utils.py
```diff
@@ -98,10 +98,11 @@ def to_list_of_dicts(param_values: ParamDictType) -> list[ParamDictType]:
     if not param_values:
         return [param_values]
 
-    max_batch_size = max(p.size()[0] for p in param_values.values())
+    max_batch_size = max(p.size()[0] for p in param_values.values() if isinstance(p, Tensor))
```
dominikandreasseitz, line 101: did you merge current main? i thought i added this recently

qadence/blocks/embedding.py
```diff
@@ -111,18 +111,21 @@ def embedding_fn(params: ParamDictType, inputs: ParamDictType) -> ParamDictType:
...
+                    values[symbol.name] = value
                 else:
-                    msg_trainable = "Trainable" if symbol.trainable else "Non-trainable"
-                    raise KeyError(
-                        f"{msg_trainable} parameter '{symbol.name}' not found in the "
-                        f"inputs list: {list(inputs.keys())} nor the "
-                        f"params list: {list(params.keys())}."
-                    )
-                values[symbol.name] = value
+                    values[symbol.name] = tensor(1.0)
```
dominikandreasseitz, line 128: why is this line there?

qadence/backends/utils.py
vytautas-a: well, the `main` is without this addition

qadence/blocks/embedding.py
vytautas-a: the value can actually be anything - I chose 1.0

qadence/backends/utils.py
dominikandreasseitz: oh good catch, thanks!

qadence/blocks/embedding.py
dominikandreasseitz: cool

dominikandreasseitz (approved):
great work @vytautas-a , lgtm

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -289,18 +298,80 @@ def dagger(self, values: dict[str, Tensor]) -> Tensor:
         """Dagger of the evolved operator given the current parameter values."""
         return _dagger(self.unitary(values))
 
+    def _get_time_parameter(self) -> str:
+        # get unique time parameters
+        unique_time_params = set()
+        for p in parameters(self.block.generator):  # type: ignore [arg-type]
+            if getattr(p, "is_time", False):
```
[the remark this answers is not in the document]
vytautas-a: `p` could be not a parameter and not have this attribute. So I'd leave as is

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -289,18 +298,80 @@ def dagger(self, values: dict[str, Tensor]) -> Tensor:
...
+                unique_time_params.add(str(p))
+
+        if len(unique_time_params) > 1:
+            raise Exception("Only a single time parameter is supported.")
```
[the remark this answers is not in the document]
vytautas-a: What would you suggest? I think it's not that important, honestly: the error message tells the user everything that he needs to know.

PR #406, opened 2024-05-10
[Refac] Use native pytorch higher-order blocks and Adjoint Differentiation

dominikandreasseitz:
pyqtorch recently introduced several higher order operations like `Sequence`, `Add` and `Scale` which can now be used via qadence


qadence/backends/adjoint.py
```diff
@@ -105,27 +106,28 @@ def _apply_adjoint(ctx: Any, op: Module) -> list:
                 ctx.projected_state = apply_operator(
                     ctx.projected_state, op.dagger(values), op.qubit_support
                 )
-            elif isinstance(op, ScalePyQOperation):
-                ctx.out_state = apply_operator(ctx.out_state, op.dagger(values), op.qubit_support)
+            elif isinstance(op, pyq.Scale):
```
Roland-djee, line 109: Couldn't you simply import them ? I guess there is no ambiguity.

Roland-djee (approved):
Thanks @dominikandreasseitz looks nice and clean.

jpmoutinho (approved):
Nice :)

qadence/backends/adjoint.py
dominikandreasseitz: file removed

PR #409, opened 2024-05-13
[Refactor] Remove setting noise model in sample call.

### LINKED ISSUE #407: [Refactor] Removing setting noise model in `sample`
While running the following script:

```python
from qadence import QuantumModel, QuantumCircuit,kron, H, Z, I
from qadence.noise import Noise

n_qubits = 4
block = kron(*[I(i) for i in range(n_qubits)])
circuit = QuantumCircuit(n_qubits, block)
model = QuantumModel(circuit=circuit)
noise = Noise(protocol=Noise.READOUT)
print(model.sample(noise=noise))
print(model.sample())
```

one would expect the last line to sample from a noiseless distribution, but it gives noisy samples. Feels like the previous line sets the the noise of the model even though it is called to sample.

Roland-djee:
Closes #407.

### DIFF
--- a/qadence/models/quantum_model.py
+++ b/qadence/models/quantum_model.py
@@ -185,8 +185,6 @@ class QuantumModel(nn.Module):
         params = self.embedding_fn(self._params, values)
         if noise is None:
             noise = self._noise
-        else:
-            self._noise = noise
         if mitigation is None:
             mitigation = self._mitigation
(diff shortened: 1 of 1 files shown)


dominikandreasseitz (approved):
thanks @Roland-djee

rajaiitp (approved):
Nice, that should fix it

Issue #422, opened 2024-05-16
[Bug] Parameter does not support numpy.int64


FergusHayes-pasqal:
If I run:

```
import numpy as np
from qadence import N
np.array([1])[0]*N(0)
```

I get the following error:

`TypeError: Parameter does not support type <class 'numpy.int64'>
`

PR #425, opened 2024-05-22
[Fix] Allow numpy numeric types in Parameters

dominikandreasseitz:
fixes https://github.com/pasqal-io/qadence/issues/422

### DIFF
--- a/tests/qadence/test_parameters.py
+++ b/tests/qadence/test_parameters.py
@@ -361,6 +361,7 @@ def test_sympy_modules() -> None:
     assert evaluate(expr) == evaluate(loaded_expr)
 
 
-def test_numpy_paramtypes() -> None:
-    Parameter(np.array([1.0]).item())
-    Parameter(np.array([1]).item())
+def test_numeric_paramtypes() -> None:
+    # We test every numeric type allowed in Parameter
+    for _numeric_type in [int, float, complex, np.int64, np.float64]:
+        Parameter(_numeric_type(1.0))
--- a/tests/qadence/test_parameters.py
+++ b/tests/qadence/test_parameters.py
@@ -359,3 +359,8 @@ def test_sympy_modules() -> None:
     loaded_expr = deserialize(d)
     assert loaded_expr == expr
     assert evaluate(expr) == evaluate(loaded_expr)
+
+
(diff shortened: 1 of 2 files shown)


tests/qadence/test_parameters.py
```diff
@@ -359,3 +359,8 @@ def test_sympy_modules() -> None:
     loaded_expr = deserialize(d)
     assert loaded_expr == expr
     assert evaluate(expr) == evaluate(loaded_expr)
+
+
+def test_numpy_paramtypes() -> None:
```
Roland-djee, line 364: What is this testing ?

dominikandreasseitz: That the cases dont fail

Roland-djee: Right, what I meant is that will any error raise be picked up as a test failure ? Another point, if we do test that we should test all other numeric types allowed.

dominikandreasseitz: yeah if the instantiation of the parameter instance fails, the test and hence the test suite fails. i think this case is just odd because python doesnt see `numpy.int64` as a child class of `int` which is the case for`numpy.float64` and `float`

dominikandreasseitz: added tests for all numeric types. lets merge?

tests/qadence/test_parameters.py
```diff
@@ -359,3 +359,9 @@ def test_sympy_modules() -> None:
     loaded_expr = deserialize(d)
     assert loaded_expr == expr
     assert evaluate(expr) == evaluate(loaded_expr)
+
+
+def test_numeric_paramtypes() -> None:
+    # We test every numeric type allowed in Parameter
+    for _numeric_type in [int, float, complex, np.int64, np.float64]:
```
Roland-djee, line 366: Didn't we define a type for this ?

Roland-djee (approved):
LGTM. Thanks @dominikandreasseitz

tests/qadence/test_parameters.py
dominikandreasseitz: we did but you cant loop over Union types it seems

PR #427, opened 2024-05-23
Rg/noisy sims pulser backend

Roland-djee:


Roland-djee:
Hey @rajaiitp I think this addresses your points in the other PR. Sorry for the confusion there. If you're OK, I'll add more tests with a range of noise probabilities and it should be good to go.

rajaiitp (changes requested):
Looks good to me. We are able to return both the expectation values and density matrices as and when required.

Something I noticed, we are applying readout errors after retrieving samples from pulser in qadence if specified. However pulser has its own readout error functionality. I'm not sure how we want to proceed with this in the future

Roland-djee:
> Something I noticed, we are applying readout errors after retrieving samples from pulser in qadence if specified. However pulser has its own readout error functionality. I'm not sure how we want to proceed with this in the future

Indeed. As discussed this is the plan for future developments.

rajaiitp (approved):
Looks good to me :)

PR #433, opened 2024-05-29
[Refac] Add device to model checkpoints

dominikandreasseitz:
When training the same model twice, however on different devices, `train_grad` fails due to the checkpoints differing in devices. This MR adds a "device" suffix to each created checkpoint

### DIFF
--- a/qadence/ml_tools/saveload.py
+++ b/qadence/ml_tools/saveload.py
@@ -61,7 +61,7 @@ def write_checkpoint(
     device = None
     try:
         # We extract the device from the pyqtorch native circuit
-        device = model.device
+        device = str(model.device).split(":")[0]  # in case of using several CUDA devices
     except Exception:
         pass
     model_checkpoint_name: str = (
--- a/qadence/ml_tools/saveload.py
+++ b/qadence/ml_tools/saveload.py
@@ -58,8 +58,18 @@ def write_checkpoint(
     from qadence.ml_tools.models import TransformedModule
     from qadence.models import QNN, QuantumModel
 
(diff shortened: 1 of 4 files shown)


awennersteen (approved):
great, thanks for fixing

qadence/ml_tools/saveload.py
```diff
@@ -58,8 +58,18 @@ def write_checkpoint(
     from qadence.ml_tools.models import TransformedModule
     from qadence.models import QNN, QuantumModel
 
-    model_checkpoint_name: str = f"model_{type(model).__name__}_ckpt_" + f"{iteration:03n}" + ".pt"
-    opt_checkpoint_name: str = f"opt_{type(optimizer).__name__}_ckpt_" + f"{iteration:03n}" + ".pt"
+    device = None
+    try:
+        # We extract the device from the pyqtorch native circuit
+        device = model.device
```
awennersteen, line 64:
```suggestion
        device = str(model.device).split(":")[0]
```
or sth like this?
To make sure we only get `cuda` and not `cuda:0` as a name?

awennersteen:
```
awennersteen:~$ python3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> torch.device("cuda:0")
device(type='cuda', index=0)
>>> a = torch.device("cuda:0")
>>> str(a)
'cuda:0'
>>> str(a).split(":")[0]
'cuda'
>>> 
```

awennersteen: great, thanks

PR #434, opened 2024-05-30
[Refac] Add device to ckpt load

dominikandreasseitz:


qadence/ml_tools/saveload.py
```diff
@@ -14,15 +14,17 @@
...
     files = [f for f in os.listdir(folder) if f.endswith(".pt") and type in f]
     if len(files) == 0:
         logger.error(f"Directory {folder} does not contain any {type} checkpoints.")
     if len(files) == 1:
         file = Path(files[0])
     else:
-        pattern = re.compile(".*_(\d+).pt$")
+        pattern = re.compile(f".*_(\d+)_device_{device}.pt$")
         max_index = -1
         for f in files:
```
awennersteen, line 29: This isn't going to be backwards compatible though. 
I'd say we should add a `try: ... except` here trying to old default format too (and add a warning if it succeeds with the old one)  to avoid lots of unnecessary breakages.

dominikandreasseitz: jep good point

dominikandreasseitz:
@inafergra @DanieleCucurachi feel free to finish this MR

Issue #214, opened 2023-11-23
[Feature] Add checkpoint_best_only in train_grad


dominikandreasseitz:
Qadence offers a `train_grad` ([link](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/train_grad.py)) method which abstracts some QNN training-related tasks away from the user.
It expects a `TrainConfig` ([link](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/config.py#L39)) which can be used to pass number of training epochs and various other training-related hyperparameters.

Task:
1. Implement logic in `train_grad` which only saves a checkpoint if the loss has improved by some user-provided `epsilon`. This can for example be done via maintaining a `current_best_validation_loss` and only save a checkpoint if `current_loss <= current_best_validation_loss - epsilon ` 
2. Add a test which makes sure your solution works
3. Add an example in our [tutorials](https://github.com/pasqal-io/qadence/blob/main/docs/tutorials/advanced_tutorials/custom-models.md) showcasing how to use it

madagra:
@dominikandreasseitz how do you define the "best" exactly? Given the creation date of the checkpoint, for example?

dominikandreasseitz:
@madagra i added a description

Roland-djee:
@dominikandreasseitz Can this be closed ?

dominikandreasseitz:
@Roland-djee no, needs to be added

Yash-10:
Hello @dominikandreasseitz , I have been trying to implement this, but I have a few questions. Could you help?

1. Is it intended to add `checkpoint_best_only` based on the train or the validation loss? I believe it's the latter since that would make more sense. If so, perhaps a new argument to the `train()` function can be added that accepts a validation data loader.
2. What is the use case of `DictDataLoader` as opposed to the (regular) `DataLoader`? (I believe this is not explained in the docs in detail, but I may have missed something).
    - For example, is it assumed that for a `DictDataLoader`, the first data loader should be used for training and the other for validation? I believe not because even if the data loader being passed is of type `DictDataLoader`, it is passed entirely to `optimize_step`. But if it's yes, then perhaps the validation can be run whenever the data loader being passed is of type `DictDataLoader`.

dominikandreasseitz:
> Hello @dominikandreasseitz , I have been trying to implement this, but I have a few questions. Could you help?
> 
[...]

PR #435, opened 2024-06-03
[Feature] Add `checkpoint_best_only` to `train_grad`

Yash-10:
This is regarding #214 (copying the tasks mentioned by @dominikandreasseitz there below).

- [X] Implement logic in train_grad which only saves a checkpoint if the loss has improved by some user-provided epsilon. This can for example be done via maintaining a `current_best_validation_loss` and only save a checkpoint if `current_loss <= current_best_validation_loss - epsilon` 
- [X] Add a test which makes sure your solution works
- [x] Add an example in our [tutorials](https://github.com/pasqal-io/qadence/blob/main/docs/tutorials/qml/ml_tools.md) showcasing how to use it

**Changes made**
- `TrainConfig` accepts new parameters: `val_every`, `val_epsilon`, and `validation_criterion`. This implementation implements logic for the case where `checkpoint_best_only=True` is set by the user, which was ignored previously since it was not implemented before.
- The code for `TrainConfig` is modified as follows:
    ```diff
    if self.validation_criterion is None:
    -     self.validation_criterion = lambda x: False    
    +    self.validation_criterion = lambda *x: False
    ```
    so that `validation_criterion` can take any no. of arguments.
- Validation is performed only if `val_every` is provided as an integer (i.e., if this holds true: `isinstance(config.val_every, int)`). If so, the data loader must be of type `DictDataLoader`. The two dataloaders are used for training and validation ("train" and "val" keys are required to access the respective dataloaders). If `val_every` is None (which is the default), the implementation falls to the default behavior as previously where no validation is performed, and so the dataloader is allowed to be either `DictDataLoader` or `DataLoader`. In this case, all the default behavior is retained as was before.
- When saving the best checkpoint when validation check is allowed to be performed (through `val_every`), the `iteration` argument to `write_checkpoint` is set to `"best"`. Also note that in this PR, `iteration` is now allowed to be either an integer or a string instead of only a string, which is why `iteration="best"` was made possible - this could make the names of the saved checkpoints slightly more customizable. The definition of "best" is also customizable through the `validation_criterion` function, which can be set through `TrainConfig`. It typically uses validation loss or accuracy but can be customized to any desired metric. Further, the function passed as `validation_criterion` can take any no. of arguments. The added example in the docs provides an example using validation loss.
- The `loss` argument to `write_tensorboard` is changed to only allow `float`'s and not `None` because otherwise, it throws an error. So, instead of `None`, it is required to pass NaN. This change is because `tensorboard` cannot handle None (as of now), it requires a float value (more discussion below).
- Note that currently, only one batch of validation data is used whenever validation is performed and not the entire validation data. This is the implementation we have agreed on for now since it keeps the code simplified.

Thanks!

### DIFF
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -94,7 +94,14 @@ config = TrainConfig(
 )
 ```
 
-If it is desired to only the save the "best" checkpoint, it must be ensured that (a) `checkpoint_best_only = True` is used while creating the configuration through `TrainConfig`, (b) `val_every` is set to a valid integer value (for example, `val_every = 10`) which controls the no. of iterations after which the validation data should be used to evaluate the model during training, which can also be set through `TrainConfig`, (c) a validation criterion is provided through the `validation_criterion`, set through `TrainConfig` to quantify the definition of "best", and (d) the dataloader passed to `train_grad` is of type `DictDataLoader`. In this case, it is expected that a validation dataloader is also provided along with the train dataloader since the validation data will be used to decide the "best" checkpoint. The dataloaders must be accessible with specific keys: "train" and "val". The criterion used to decide the "best" checkpoint can be customized by `validation_criterion`, which should be a function that can take any number of arguments and return a boolean value (True or False) indicating whether some validation metric is satisfied or not. Typical choices are to return True when the validation loss (accuracy) has decreased (increased) compared to smallest (largest) value from previous iterations at which a validation check was performed.
+If it is desired to only the save the "best" checkpoint, the following must be ensured:
+
+(a) `checkpoint_best_only = True` is used while creating the configuration through `TrainConfig`,
+(b) `val_every` is set to a valid integer value (for example, `val_every = 10`) which controls the no. of iterations after which the validation data should be used to evaluate the model during training, which can also be set through `TrainConfig`,
+(c) a validation criterion is provided through the `validation_criterion`, set through `TrainConfig` to quantify the definition of "best", and
+(d) the dataloader passed to `train_grad` is of type `DictDataLoader`. In this case, it is expected that a validation dataloader is also provided along with the train dataloader since the validation data will be used to decide the "best" checkpoint. The dataloaders must be accessible with specific keys: "train" and "val".
+
+The criterion used to decide the "best" checkpoint can be customized by `validation_criterion`, which should be a function that can take any number of arguments and return a boolean value (True or False) indicating whether some validation metric is satisfied or not. Typical choices are to return True when the validation loss (accuracy) has decreased (increased) compared to smallest (largest) value from previous iterations at which a validation check was performed.
 
 Let's see it in action with a simple example.
 
@@ -103,7 +110,7 @@ Let's see it in action with a simple example.
 In Quantum Machine Learning, the general consensus is to use `complex128` precision for states and operators and `float64` precision for parameters. This is also the convention which is used in `qadence`.
 However, for specific usecases, lower precision can greatly speed up training and reduce memory consumption. When using the `pyqtorch` backend, `qadence` offers the option to move a `QuantumModel` instance to a specific precision using the torch `to` syntax.
 
-Let's look at a complete example of how to use `train_with_grad` now.
+Let's look at a complete example of how to use `train_with_grad` now. Here we perform a validation check during training and use a validation criterion that checks whether the validation loss in the current iteration has decreased compared to the lowest validation loss from all previous iterations. For demonstration, the train and the validation data are kept the same here. However, it is beneficial and encouraged to keep them distinct in practice to understand model's generalization capabilities.
 
 ```python exec="on" source="material-block" html="1"
 from pathlib import Path
@@ -142,8 +149,6 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
     loss = criterion(out, y)
     return loss, {}
 
-# Here we use the validation criterion that checks whether the validation loss in the current
-# iteration has decreased compared to the lowest validation loss from all previous iterations.
 def validation_criterion(
     current_validation_loss: float, current_best_validation_loss: float, val_epsilon: float
 ) -> bool:
@@ -163,9 +168,6 @@ fn = lambda x, degree: .05 * reduce(add, (torch.cos(i*x) + torch.sin(i*x) for i 
 x = torch.linspace(0, 10, batch_size, dtype=torch.float32).reshape(-1, 1)
 y = fn(x, 5)
 
-# For demonstration, the train and the validation data are kept the same here.
-# However, it is beneficial and encouraged to keep them distinct in practice
-# to understand model's generalization capabilities.
 data = DictDataLoader(
     {
         "train": to_dataloader(x, y, batch_size=batch_size, infinite=True),
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -94,6 +94,8 @@ config = TrainConfig(
 )
 ```
 
+If it is desired to only the save the "best" checkpoint, it must be ensured that (a) `checkpoint_best_only = True` is used while creating the configuration through `TrainConfig`, (b) `val_every` is set to a valid integer value (for example, `val_every = 10`) which controls the no. of iterations after which the validation data should be used to evaluate the model during training, which can also be set through `TrainConfig`, (c) a validation criterion is provided through the `validation_criterion`, set through `TrainConfig` to quantify the definition of "best", and (d) the dataloader passed to `train_grad` is of type `DictDataLoader`. In this case, it is expected that a validation dataloader is also provided along with the train dataloader since the validation data will be used to decide the "best" checkpoint. The dataloaders must be accessible with specific keys: "train" and "val". The criterion used to decide the "best" checkpoint can be customized by `validation_criterion`, which should be a function that can take any number of arguments and return a boolean value (True or False) indicating whether some validation metric is satisfied or not. Typical choices are to return True when the validation loss (accuracy) has decreased (increased) compared to smallest (largest) value from previous iterations at which a validation check was performed.
+
 Let's see it in action with a simple example.
 
 ### Fitting a funtion with a QNN using `ml_tools`
@@ -114,7 +116,7 @@ import matplotlib.pyplot as plt
 from qadence import Parameter, QuantumCircuit, Z
 from qadence import hamiltonian_factory, hea, feature_map, chain
 from qadence.models import QNN
-from qadence.ml_tools import  TrainConfig, train_with_grad, to_dataloader
+from qadence.ml_tools import  TrainConfig, train_with_grad, to_dataloader, DictDataLoader
 
 DEVICE = torch.device('cuda') if torch.cuda.is_available() else torch.device('cpu')
 DTYPE = torch.complex64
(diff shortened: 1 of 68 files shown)


qadence/ml_tools/train_grad.py
```diff
@@ -61,6 +66,10 @@ def train(
             called every `config.write_every` iterations. The function must have
             the signature `write_tensorboard(writer, loss, metrics, iteration)`
             (see the example below).
+        epsilon: Safety margin to check if validation loss is smaller than the lowest
```
dominikandreasseitz, line 69: can we move this into the `TrainConfig` rather?

Yash-10: Yes, done, thanks!

qadence/ml_tools/train_grad.py
```diff
@@ -125,6 +128,13 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
     # initialize tensorboard
     writer = SummaryWriter(config.folder, purge_step=init_iter)
 
+    if config.perform_val_check and not isinstance(dataloader, DictDataLoader):
```
dominikandreasseitz, line 131: maybe we could get rid of `peform_val_check` and just check if the user passes `val_every` which can be None by default. and in that case we dont validate

Yash-10: Sure, that's a better design! Done.

qadence/ml_tools/config.py
```diff
@@ -38,6 +38,10 @@ class TrainConfig:
     """
     checkpoint_best_only: bool = False
     """Write model/optimizer checkpoint only if a metric has improved."""
+    val_every: Union[int, None] = None
+    """Calculate validation metric."""
+    epsilon: float = 1e-5
```
dominikandreasseitz, line 43: lets rename `epsilon` to `val_epsilon` for example

dominikandreasseitz:
ping @debrevitatevitae for his opinion

qadence/ml_tools/config.py
Yash-10: Done.

qadence/ml_tools/train_grad.py
```diff
@@ -125,6 +126,17 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
+    if perform_val and not isinstance(dataloader, DictDataLoader):
+        raise ValueError(
+            "If `config.val_every` is provided as an integer, dataloader must"
+            "be an instance of `DictDataLoader`"
+        )
+    if perform_val:
+        iter_keys = list(dataloader.dataloaders.keys())
+        val_dataloader = dataloader.dataloaders[iter_keys[1]]
+        dataloader = dataloader.dataloaders[iter_keys[0]]
+
```
debrevitatevitae, line 139: I would not by default assume that `val_dataloader` is at index 1. I would rather enforce a key, s.a. `val_dataloader.dataloaders['validation']`

Yash-10: Great, thanks! Done.

qadence/ml_tools/config.py
```diff
@@ -38,6 +38,16 @@ class TrainConfig:
     """
     checkpoint_best_only: bool = False
     """Write model/optimizer checkpoint only if a metric has improved."""
+    val_every: Union[int, None] = None
```
Roland-djee, line 41:
```suggestion
    val_every: int | None = None
```

Roland-djee (commented):
Thanks for this @Yash-10. Not much to say. I let @dominikandreasseitz have the final word on this.

qadence/ml_tools/config.py
Yash-10: Thanks, done!

qadence/ml_tools/train_grad.py
```diff
@@ -177,16 +198,32 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
                 if iteration % config.write_every == 0:
                     write_tensorboard(writer, loss, metrics, iteration)
 
+                if perform_val:
+                    if iteration % config.val_every == 0:
+                        # TODO: It may be desired that the entire validation set be used
```
Yash-10, line 203: This is a question that I would like to know someone's thoughts on. I believe it's more useful to iterate over the entire validation data whenever a validation check needs to be performed. Currently, this check is only done for a single batch, and that batch also changes at different iterations when a validation check is performed.

That can be done by a for/while loop to loop over the entire validation data with the below logic:

```py
def validate():
    val_loss = 0.0
    for data in val_dataloader:
        data_to_device = data_to_device(data, device=device, dtype=data_dtype)
        loss, _ = loss_fn(model, data_to_device)
        val_loss += loss
    return val_loss
```

qadence/ml_tools/train_grad.py
```diff
@@ -125,6 +126,22 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
     # initialize tensorboard
     writer = SummaryWriter(config.folder, purge_step=init_iter)
 
+    perform_val = isinstance(config.val_every, int)
+    if perform_val and not isinstance(dataloader, DictDataLoader):
```
dominikandreasseitz, line 130: should we move the first `if` into the second?

tests/ml_tools/test_train.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
```
dominikandreasseitz, line 3: why the `os` import?

dominikandreasseitz:
@Yash-10 can you fix the linting?

qadence/ml_tools/train_grad.py
dominikandreasseitz: i see your point. lets just validate on batches for now

qadence/ml_tools/train_grad.py
Yash-10: thanks, done!

Yash-10:
About adding this to the tutorials, would [this tutorial](https://github.com/pasqal-io/qadence/blob/main/docs/tutorials/advanced_tutorials/custom-models.md) or the general tutorial [here](https://github.com/pasqal-io/qadence/blob/main/docs/tutorials/qml/ml_tools.md) be the right place to describe `checkpoint_best_only`? #214 mentioned the former but I was wondering if it's better placed in the latter.

dominikandreasseitz:
> About adding this to the tutorials, would [this tutorial](https://github.com/pasqal-io/qadence/blob/main/docs/tutorials/advanced_tutorials/custom-models.md) or the general tutorial [here](https://github.com/pasqal-io/qadence/blob/main/docs/tutorials/qml/ml_tools.md) be the right place to describe `checkpoint_best_only`? #214 mentioned the former but I was wondering if it's better placed in the latter.

@Yash-10 i agree it should be the latter

qadence/ml_tools/train_grad.py
```diff
@@ -177,16 +198,28 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
+                    if iteration % config.val_every == 0:
+                        xs = next(dl_iter_val)
+                        xs_to_device = data_to_device(xs, device=device, dtype=data_dtype)
+                        val_loss, _ = loss_fn(model, xs_to_device)
+                        if config.validation_criterion(val_loss, best_val_loss, config.val_epsilon):
+                            best_val_loss = val_loss
+                            if config.folder and config.checkpoint_best_only:
+                                write_checkpoint(config.folder, model, optimizer, iteration="best")
+                            metrics["val_loss"] = val_loss
+                            write_tensorboard(writer, math.nan, metrics, iteration)
```
dominikandreasseitz, line 211: why `math.nan`?

Yash-10: `write_tensorboard(writer, None, metrics, iteration)` with `None` instead of NaN raised an error. This is the traceback.

```bash
    def make_np(x):
        """
        Convert an object into numpy array.
    
        Args:
          x: An instance of torch tensor or caffe blob name
    
        Returns:
            numpy.array: Numpy array
        """
        if isinstance(x, np.ndarray):
            return x
        if isinstance(x, str):  # Caffe2 will pass name of blob(s) to fetch
            return _prepare_caffe2(x)
        if np.isscalar(x):
            return np.array([x])
        if isinstance(x, torch.Tensor):
            return _prepare_pytorch(x)
>       raise NotImplementedError(
            f"Got {type(x)}, but numpy array, torch tensor, or caffe2 blob name are expected."
        )
E       NotImplementedError: Got <class 'NoneType'>, but numpy array, torch tensor, or caffe2 blob name are expected.

../../.local/share/hatch/env/virtual/qadence/G8biN769/qadence/lib/python3.11/site-packages/torch/utils/tensorboard/_convert_np.py:24: NotImplementedError
```

That's why I needed to change [this line in `printing.py`](https://github.com/pasqal-io/qadence/pull/435/files#diff-409c355dde25623d704f5e5d205ab9043700fff703bd4d28443f35d186d95dd1L15-R15). I have `tensorboard==2.16.2`.

Yash-10:
The linting error is slightly confusing:

```
qadence/ml_tools/train_grad.py:206: error: "None" not callable  [misc]
```
It's referring to this line:

```py
if config.validation_criterion(val_loss, best_val_loss, config.val_epsilon):
```

but as I see, if `validation_criterion=None` is passed, it's replaced with a lambda function that always returns False, so it should not be None. I'll try to investigate this.

Yash-10:
@dominikandreasseitz I have now also incorporated this into the tutorials, so this is ready for review.

docs/tutorials/qml/ml_tools.md
```diff
@@ -94,6 +94,8 @@ config = TrainConfig(
...
...h `TrainConfig`, (b) `val_every` is set to a valid integer value (for example, `val_every = 10`) which controls the no. of iterations after which the validation data should be used to evaluate the model during training, which can also be set through `TrainConfig`, (c) a validation criterion is provided through the `validation_criterion`, set through `TrainConfig` to quantify the definition of "best", and (d) the dataloader passed to `train_grad` is of type `DictDataLoader`. In this case, it is expected that a validation dataloader is also provided along with the train dataloader since the validation data will be used to decide the "best" checkpoint. The dataloaders must be accessible with specific keys: "train" and "val". The criterion used to decide the "best" checkpoint can be customized by `validation_criterion`, which should be a function that can take any number of arguments and return a boolean value (True or False) indicating whether some validation metric is satisfied or not. Typical choices are to return True when the validation loss (accuracy) has decreased (increased) compared to smallest (largest) value from previous iterations at which a validation check was performed.
```
jpmoutinho, line 97: Could you make the formatting a bit nicer? :)

E.g. maybe make the `(a), (b), (c)` points in a list of bulletpoints.

docs/tutorials/qml/ml_tools.md
```diff
@@ -140,18 +142,36 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
     loss = criterion(out, y)
     return loss, {}
 
+# Here we use the validation criterion that checks whether the validation loss in the current
+# iteration has decreased compared to the lowest validation loss from all previous iterations.
+def validation_criterion(
```
jpmoutinho, line 147: Typically it's best to avoid long comments in a documentation snippet. You can break the snippet in multiple ones and add this text as normal documentation description in between.

docs/tutorials/qml/ml_tools.md
```diff
@@ -140,18 +142,36 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
 )
 
 fn = lambda x, degree: .05 * reduce(add, (torch.cos(i*x) + torch.sin(i*x) for i in range(degree)), 0.)
 x = torch.linspace(0, 10, batch_size, dtype=torch.float32).reshape(-1, 1)
 y = fn(x, 5)
 
-data = to_dataloader(x, y, batch_size=batch_size, infinite=True)
+# For demonstration, the train and the validation data are kept the same here.
+# However, it is beneficial and encouraged to keep them distinct in practice
+# to understand model's generalization capabilities.
```
jpmoutinho, line 168: Same for this one.

tests/ml_tools/test_train.py
```diff
@@ -200,3 +207,122 @@ def test_fit_sin_adjoint(BasicAdjointQNN: torch.nn.Module) -> None:
 
     x_test = torch.rand(1, 1)
     assert torch.allclose(torch.sin(x_test), model(x_test), rtol=1e-1, atol=1e-1)
+
+
+def test_train_dataloader_raisesError_if_val_check_is_True_but_non_dict_dataloader(
```
jpmoutinho, line 212: Could you try to simplify these tests a little bit? There's some code duplication in all three of them, i.e. the `TrainConfig` seems to be the same, so maybe you can separate that into a function that is reused. 

Also no need for the super long test titles :) If you wish to have some extra context to what the test is doing that's cool of course, but best to write in a docstring or comment just below I think.

jpmoutinho (commented):
Thanks for the effort @Yash-10! @dominikandreasseitz will take a better look at the logic of your code, but just a few suggestions from my side mostly on the formatting.

tests/ml_tools/test_train.py
Yash-10: Thanks.

> There's some code duplication in all three of them, i.e. the TrainConfig seems to be the same, so maybe you can separate that into a function that is reused.

I agree that it helps reduce space, so I have factored out config creation.

> Also no need for the super long test titles :) If you wish to have some extra context to what the test is doing that's cool of course, but best to write in a docstring or comment just below I think.

Okay, great. I have reduced the lengths of the test titles. Hopefully, that makes it cleaner!

docs/tutorials/qml/ml_tools.md
Yash-10: Thanks for this suggestion, done!

docs/tutorials/qml/ml_tools.md
```diff
@@ -140,18 +149,31 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
     loss = criterion(out, y)
     return loss, {}
 
+def validation_criterion(
```
dominikandreasseitz, line 152: nice!

dominikandreasseitz:
@Yash-10 can you update the description?

qadence/ml_tools/train_grad.py
dominikandreasseitz: interesting, but thanks for fixing

docs/tutorials/qml/ml_tools.md
dominikandreasseitz: done, resolving

Yash-10:
@dominikandreasseitz I have updated the PR description.

dominikandreasseitz (approved):
great work @Yash-10 , approved. Thanks for contributing to Qadence!

PR #438, opened 2024-06-04
[Infra] Remove dlprof dependency out of default env

dominikandreasseitz:


awennersteen:
wow, GitHub just merges and doesn't wait for pipelines?
What craziness is this

dominikandreasseitz:
> wow, GitHub just merges and doesn't wait for pipelines?
> What craziness is this
> 
> Can we change this setting?

I thought we deactivated this behavior before? Yeah we should if its not the case

PR #440, opened 2024-06-05
[Performance] Cache non-parametric pulser observable

dominikandreasseitz:
In case the user provides a non-parametric observable, compute it once and reuse it


vytautas-a (approved):
Looks good

PR #442, opened 2024-06-05
[Performance] Improve horqrux sampling

dominikandreasseitz:
Increase speed of horqrux sampling:

New:
In [3]: %timeit sample(RX(20,'theta'),backend='horqrux')
45.3 ms ± 187 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

Old:
In [2]: %timeit sample(RX(20,'theta'),backend='horqrux')
4.28 s ± 25.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### DIFF
--- a/qadence/backends/horqrux/backend.py
+++ b/qadence/backends/horqrux/backend.py
@@ -204,8 +204,7 @@ class Backend(BackendInterface):
         )
         probs = jnp.abs(jnp.float_power(wf, 2.0)).ravel()
         key = jax.random.PRNGKey(0)
-        # JAX avoids implicit global random states,
-        # and instead tracks state explicitly via a random key
+        # JAX handles pseudo random number generation by tracking an explicit state via a random key
         # For more details, see https://jax.readthedocs.io/en/latest/random-numbers.html
         samples = jax.vmap(
             lambda subkey: jax.random.choice(
--- a/qadence/backends/horqrux/backend.py
+++ b/qadence/backends/horqrux/backend.py
@@ -195,28 +195,6 @@ class Backend(BackendInterface):
         if n_shots < 1:
             raise ValueError("You can only call sample with n_shots>0.")
 
-        def _sample(
-            _probs: ArrayLike, n_shots: int, endianness: Endianness, n_qubits: int
-        ) -> Counter:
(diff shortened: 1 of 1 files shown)


atiyo:
> @atiyo do you see a nice way to do this even better?

Looks good to me.

Only thing that comes to mind is trying to optimise that dictionary comprehension into something Jax native. e.g. vmapping that as well

 But maybe that's a bottleneck at all, in which case it looks good as is to me.

qadence/backends/horqrux/backend.py
```diff
@@ -225,16 +203,24 @@ def _smple(accumulator: ArrayLike, i: int) -> tuple[ArrayLike, None]:
...
         probs = jnp.abs(jnp.float_power(wf, 2.0)).ravel()
-        samples = [
-            _sample(
-                _probs=probs,
-                n_shots=n_shots,
-                endianness=endianness,
-                n_qubits=circuit.abstract.n_qubits,
-            ),
-        ]
+        key = jax.random.PRNGKey(0)
```
awennersteen, line 206: is this intentionally left in?

Roland-djee: @awennersteen it is used in 211 ?

Roland-djee (approved):
Nice! If I understand correctly, you sample from a ransom split ensemble of shots ?

qadence/backends/horqrux/backend.py
awennersteen: Spoke with Dominik and apparently this is how it's done in Jax. 
See e.g.
https://jax.readthedocs.io/en/latest/random-numbers.html

I'll approve but expecting a commit with a comment on this before merging for future reference.

awennersteen (approved):

qadence/backends/horqrux/backend.py
awennersteen: Yeah, Dominik explained @Roland-djee 

I think this should have a quick comment left in though for future reference for those not familiar with Jax.

Roland-djee: Agree.

dominikandreasseitz: agreed, added

PR #446, opened 2024-06-06
[Feature] Enable noisy sims in pulser backend

### LINKED ISSUE #445: [Feature] Enable noisy sims in Pulser backend
Enable noisy expectations in Pulser backend by adding functionality to compute expectations using density matrices in `PulserObservable` forward method.

Roland-djee:
Closes #445

### DIFF
--- a/qadence/backends/pulser/backend.py
+++ b/qadence/backends/pulser/backend.py
@@ -264,7 +264,10 @@ class Backend(BackendInterface):
         if isinstance(noise_probs, Iterable):
             noisy_batched_dms = []
             for noise_prob in noise_probs:
-                noisy_batched_dms.append(run_noisy_sim(noise_prob))
+                noisy_sim = run_noisy_sim(noise_prob)
+                if not param_values:
+                    noisy_sim = noisy_sim[0]
+                noisy_batched_dms.append(noisy_sim)
             noisy_batched_dms = torch.stack(noisy_batched_dms)
         else:
             noisy_batched_dms = run_noisy_sim(noise_probs)
--- a/qadence/backends/pulser/backend.py
+++ b/qadence/backends/pulser/backend.py
@@ -353,10 +353,10 @@ class Backend(BackendInterface):
                 # TODO: There should be a better check for batched density matrices.
                 if dms.size()[0] > 1:
                     res_list = [
-                        [obs.native(dm, param_values, qubit_support=support) for dm in dms]
+                        [obs.native(dm, param_values, qubit_support=support, noise=noise) for dm in dms]
                         for obs in observable
                     ]
-                    res = torch.stack([torch.transpose(torch.stack(res), 0, 1) for res in res_list])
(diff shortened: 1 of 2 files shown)


Roland-djee:
@rajaiitp @dominikandreasseitz Is this OK with you ? If so, I'd like to merge and release to fix protocols.

qadence/backends/pulser/convert_ops.py
```diff
@@ -39,13 +40,14 @@ def forward(
...
-        if self.block.is_parametric:
-            block_mat = block_to_tensor(
-                self.block, values, qubit_support=qubit_support, endianness=endianness
-            ).squeeze(0)
-        else:
-            block_mat = self.block_mat
-
-        return torch.sum(torch.matmul(state, block_mat) * state.conj(), dim=1)
+        # FIXME: cache this, it is very inefficient for non-parametric observables
+        block_mat = block_to_tensor(
```
dominikandreasseitz, line 47: ok so we cant cache here?

qadence/backends/pulser/backend.py
```diff
@@ -350,10 +353,18 @@ def expectation(
                 # TODO: There should be a better check for batched density matrices.
                 if dms.size()[0] > 1:
                     res_list = [
-                        [obs.native(dm, param_values, qubit_support=support) for dm in dms]
+                        [
+                            obs.native(
+                                dm.squeeze(), param_values, qubit_support=support, noise=noise
```
dominikandreasseitz, line 358: so we assume dim `0` is always 1? if not, that can result in undefined behaviour

Roland-djee: Yes. At this point, `dms` are batched density matrices by the noise parameters. So `dm` is a single density matrix.

qadence/backends/pulser/convert_ops.py
Roland-djee: I haven't tried to fix this, it predates this PR. Not sure though how I could cache the tensor. Any suggestions ?

dominikandreasseitz: the way it was before, which is a very recent MR of mine, did exactly that. in the constructor you check if its a parametric block, if not, you create the tensor and store it as a buffer. in the forward you check if the block is parametric and if not, use the buffer, otherwise build the tensor

qadence/backends/pulser/backend.py
dominikandreasseitz: ah ok

qadence/backends/pulser/convert_ops.py
Roland-djee: Alright, lemme have a look.

Roland-djee: Should be in 0c261c2.

rajaiitp (approved):
looks good to me

PR #450, opened 2024-06-07
[Feature] Add mlflow experiment tracker

debrevitatevitae:
Modify training routines and config to be able to log with ML-Flow.

TODOs:

- [x] Add MLConfig with username/pw fields which can be set using `os.getenv`
- [x] Add StrEnum `EXPERIMENT_TRACKER` with options `TENSORBOARD` and `MLFLOW`
- [x] Add EXPERIMENT_TRACKER to `TrainConfig` as `tracking_tool`
- [x] Modify `train` loop to either perform `write_tensorboard` or `write_mlflow` depending on the setting of `tracking_tool` is
- [x] Add log_params for configs
- [x] Add log_figures using callbacks
- [x] Add log_model (overlap with checkpointing?)
- [ ] Test integration with cloud

TODOS @debrevitatevitae:

- [x] FInalize interface/ features for mlflow tracking
- [x] Align `train_no_grad` with `train_grad`
- [x] Create a tutorial showcasing mlflow with remote tracking
- [x] Add tests with mlflow tracking


qadence/ml_tools/utils.py
```diff
@@ -1,16 +1,21 @@
...
 
 from qadence.blocks import AbstractBlock, parameters
 from qadence.circuit import QuantumCircuit
+from qadence.logger import get_script_logger
 from qadence.ml_tools.models import TransformedModule
 from qadence.models import QNN, QuantumModel
 from qadence.parameters import Parameter, stringify
 
+logger = get_script_logger(__name__)
+
```
awennersteen, line 18:
```suggestion
from typing import Any
from logging import get_logger

from torch import Tensor, rand

from qadence.blocks import AbstractBlock, parameters
from qadence.circuit import QuantumCircuit
from qadence.ml_tools.models import TransformedModule
from qadence.models import QNN, QuantumModel
from qadence.parameters import Parameter, stringify

logger = get_logger(__name__)

```

qadence/ml_tools/utils.py
```diff
@@ -44,3 +49,25 @@ def _(qnn: QNN, batch_size: int = 1) -> dict[str, Tensor]:
...
+    MLFLOW_TRACKING_URI: str = os.getenv("MLFLOW_TRACKING_URI", "")
+    MLFLOW_TRACKING_USERNAME: str = os.getenv("MLFLOW_TRACKING_USERNAME", "")
+    MLFLOW_TRACKING_PASSWORD: str = os.getenv("MLFLOW_TRACKING_PASSWORD", "")
+    EXPERIMENT: str = os.getenv("MLFLOW_EXPERIMENT", "")
+
+    def __post_init__(self) -> None:
+        if self.MLFLOW_TRACKING_USERNAME != "":
+            logger.info(
+                f"Intialized mlflow remote logging for user {self.MLFLOW_TRACKING_USERNAME}."
```
awennersteen, line 72: add experiment to the output too?

qadence/ml_tools/train_grad.py
```diff
@@ -122,9 +125,17 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
     else:
         model = model.to(device=device, dtype=dtype)
-    # initialize tensorboard
-    writer = SummaryWriter(config.folder, purge_step=init_iter)
+    # initialize tracking tool
+    if config.tracking_tool == ExperimentTrackingTool.TENSORBOARD:
+        writer = SummaryWriter(config.folder, purge_step=init_iter)
+    else:
+        mlflowconfig = MLFlowConfig()  # Set up credentials for mlflow tracking
+        writer = importlib.import_module("mflow")
```
awennersteen, line 133: Not sure about others but I personally prefer to do a 
```
try:
    import X
except ImportError
     X = None
     logger.debug("X not available")
```
At the top of the file together with normal imports

qadence/ml_tools/utils.py
dominikandreasseitz: why? dont you wanna use your creation :)

qadence/ml_tools/train_grad.py
dominikandreasseitz: you dont always wanna import mlflow, only in case the user passes it in the trainconfig. generally, i agree with you but in this case it avoids an unnecessary import

qadence/ml_tools/train_grad.py
```diff
@@ -29,7 +31,7 @@ def train(
     loss_fn: Callable,
     device: torch_device = None,
     optimize_step: Callable = optimize_step,
-    write_tensorboard: Callable = write_tensorboard,
+    write_tracker: Callable = write_tracker,
```
debrevitatevitae, line 34: @dominikandreasseitz , seems to me like we don't need `write_tracker` as an input, because the user defines the tracking tool via `config` and the rest is handled internally in `train_grad`.

Roland-djee:
@dominikandreasseitz @debrevitatevitae Is there review needed here ? Sorry, I lost track a little.

qadence/ml_tools/utils.py
awennersteen: well, it's part of core qadence, so it can use the standard logging mechanism

Roland-djee (approved):
Thanks @debrevitatevitae some cosmetic changes from me.

dominikandreasseitz:
@debrevitatevitae please try to finish the PR with @Roland-djee's help thanks!

qadence/ml_tools/config.py
```diff
@@ -60,3 +82,35 @@ def __post_init__(self) -> None:
...
+    RUN_NAME: str = os.getenv("MLFLOW_RUN_NAME", "test_0")
+
+    def __post_init__(self) -> None:
+        import mlflow
+
+        if self.MLFLOW_TRACKING_USERNAME != "":
+            logger.info(
+                f"Intialized mlflow remote logging for user {self.MLFLOW_TRACKING_USERNAME}."
+            )
+        mlflow.set_tracking_uri(self.MLFLOW_TRACKING_URI)
```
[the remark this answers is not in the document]
smitchaudhary: If it is empty, then it defaults to storing it at `./mlruns`. Will add it to the docstrings.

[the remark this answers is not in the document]
smitchaudhary: Added this and pushed with the cosmetic changes suggested.

debrevitatevitae:
@dominikandreasseitz, about model logging, there is a bit of an overlap with the checkpointing already there. I added a bool to log the final model with mlflow, which can be then loaded also through mlflow. See commit 714949a.

debrevitatevitae:
On another note, I am seeing that managing 2 tracking tools requires some tricks.

For instance, if there is an error before `train_grad` or `train_no_grad` hits `mlflow.end_run()`, then the run is not closed and mlflow might try to create a new run with the same ID, when `MLFlowConfig` has all defaults.

One way around that is to decorate `train_grad` and `train_no_grad` with something that handles the exception, e.g.
```python
def mlflow_run_decorator(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        try:
            mlflow.start_run()
            result = func(*args, **kwargs)
            return result
        except Exception as e:
            mlflow.end_run()
            raise e

    return wrapper
```

The only issue is that now the user has to import the decorator and decorate manually if they want to use mlflow, because we allow both for tb and mlflow.

Maybe in the future it might be worth considering if both should be maintained.

smitchaudhary (approved):

docs/tutorials/qml/ml_tools.md
```diff
@@ -306,3 +306,129 @@ def train(
...
+random.seed(hyperparams["seed"])
+
+
+def dataloader(batch_size: int = 25) -> DataLoader:
+    x = torch.linspace(0, 1, batch_size).reshape(-1, 1)
+    y = torch.cos(x)
+    return to_dataloader(x, y, batch_size=batch_size, infinite=True)
+```
+
+We continue with the usual QNN definition, as well as loss function and optimizer.
```
Roland-djee, line 369:
```suggestion
We continue with the regular QNN definition, together with the loss function and optimizer.
```

qadence/ml_tools/printing.py
```diff
@@ -1,7 +1,23 @@
...
+from torch.utils.data import DataLoader
 from torch.utils.tensorboard import SummaryWriter
 
+from qadence.ml_tools.data import DictDataLoader
+from qadence.types import ExperimentTrackingTool
+
+logger = getLogger(__name__)
+
+PlottingFunction = Callable[[Module, int], tuple[str, Figure]]
+InputData = Union[Tensor, dict[str, Tensor]]
```
Roland-djee, line 19:
```suggestion
InputData = Tensor | dict[str, Tensor]
```

qadence/ml_tools/printing.py
```diff
@@ -20,5 +36,112 @@ def write_tensorboard(
...
+) -> None:
+    for pf in plotting_functions:
+        descr, fig = pf(model, iteration)
+        writer.log_figure(fig, descr)
+
+
+def log_model_mlflow(
+    writer: Any,
+    model: Module,
+    dataloader: Union[None, DataLoader, DictDataLoader],
```
Roland-djee, line 85:
```suggestion
    dataloader: DataLoader | DictDataLoader | None
```

qadence/ml_tools/printing.py
```diff
@@ -20,5 +36,112 @@ def write_tensorboard(
...
+
+MODEL_LOGGER_MAPPING = {
+    ExperimentTrackingTool.TENSORBOARD: log_model_tensorboard,
+    ExperimentTrackingTool.MLFLOW: log_model_mlflow,
+}
+
+
+def write_tracker(
+    args: Any, tracking_tool: ExperimentTrackingTool = ExperimentTrackingTool.TENSORBOARD
+) -> None:
```
Roland-djee, line 128:
```suggestion
) -> dict:
```
Maybe defining the return type would help removing the `# type: ignore` directive ?

qadence/ml_tools/train_grad.py
```diff
@@ -122,8 +131,15 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
         model = model.to(device=device, dtype=dtype)
-    # initialize tensorboard
-    writer = SummaryWriter(config.folder, purge_step=init_iter)
+    # initialize tracking tool
+    if config.tracking_tool == ExperimentTrackingTool.TENSORBOARD:
+        writer = SummaryWriter(config.folder, purge_step=init_iter)
+    else:
+        writer = importlib.import_module("mlflow")
+
+        # writer.mlflow.pytorch.autolog(
```
Roland-djee, line 140: Remove.

qadence/ml_tools/printing.py
```diff
@@ -20,5 +36,112 @@ def write_tensorboard(
...
+}
+
+MODEL_LOGGER_MAPPING = {
+    ExperimentTrackingTool.TENSORBOARD: log_model_tensorboard,
+    ExperimentTrackingTool.MLFLOW: log_model_mlflow,
+}
+
+
+def write_tracker(
+    args: Any, tracking_tool: ExperimentTrackingTool = ExperimentTrackingTool.TENSORBOARD
```
Roland-djee, line 127:
```suggestion
    *args: Any, tracking_tool: ExperimentTrackingTool = ExperimentTrackingTool.TENSORBOARD
```

I would suggest to unpack here so that you can make it variadic and don't have to pass a tuple.

qadence/ml_tools/train_grad.py
```diff
@@ -166,14 +182,20 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
                 best_val_loss, metrics = loss_fn(model, xs_to_device)
 
                 metrics["val_loss"] = best_val_loss
-                write_tensorboard(writer, None, metrics, init_iter)
+                write_tracker(
+                    (writer, None, metrics, init_iter), tracking_tool=config.tracking_tool
```
Roland-djee, line 186:
```suggestion
                    writer, None, metrics, init_iter, tracking_tool=config.tracking_tool
```

If you make it variadic as above.

qadence/ml_tools/train_no_grad.py
```diff
@@ -63,8 +77,15 @@ def _update_parameters(
...
 
-    # initialize tensorboard
-    writer = SummaryWriter(config.folder, purge_step=init_iter)
+    # initialize tracking tool
+    if config.tracking_tool == ExperimentTrackingTool.TENSORBOARD:
+        writer = SummaryWriter(config.folder, purge_step=init_iter)
+    else:
+        writer = importlib.import_module("mlflow")
+
+        # writer.mlflow.pytorch.autolog(
```
Roland-djee, line 86: Remove.

Roland-djee (commented):
Thanks a lot @debrevitatevitae, that's really nice and clean. Some minor comments from me. Happy to approve once addressed.

qadence/ml_tools/config.py
```diff
@@ -38,14 +51,38 @@ class TrainConfig:
     """
     checkpoint_best_only: bool = False
     """Write model/optimizer checkpoint only if a metric has improved."""
-    validation_criterion: Optional[Callable] = None
+    validation_criterion: Callable | None = None
     """A boolean function which evaluates a given validation metric is satisfied."""
     trainstop_criterion: Optional[Callable] = None
```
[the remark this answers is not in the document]
debrevitatevitae: Fixed in 14373c7

docs/tutorials/qml/ml_tools.md
```diff
@@ -306,3 +306,129 @@ def train(
...
+export MLFLOW_TRACKING_URI=sqlite:///mlruns.db
+```
+
+Qadence can also read the following two environment variables to define the mlflow experiment name and run name
+```bash
+export MLFLOW_EXPERIMENT=test_experiment
+export MLFLOW_RUN_NAME=run_0
+```
+
+If no tracking URI is provided, mlflow stores runs and artifacts in the local `./mlflow` directory and if no names are defined, the experiment and run will be named with random UUIDs.
```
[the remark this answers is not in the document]
debrevitatevitae: resolved in 243366c, although in a slightly different way

qadence/ml_tools/printing.py
debrevitatevitae: Had to actually revert this in fdf05f1, because of invalid type definition error raised by mypy.

qadence/ml_tools/printing.py
debrevitatevitae: Fixed by defined some types for the mappings in fdf05f1

qadence/ml_tools/train_grad.py
debrevitatevitae: should I then just import mlflow normally and set 
```python
else:
    writer = mlflow
```
?

qadence/ml_tools/printing.py
debrevitatevitae: fixed it in eb9a98c

qadence/ml_tools/train_grad.py
debrevitatevitae: Fixed it ineb9a98c

PR #454, opened 2024-06-10
[Infra] Update pulser dependencies

jpmoutinho:


pyproject.toml
```diff
@@ -75,16 +75,11 @@ protocols = ["qadence-protocols"]
 libs = ["qadence-libs"]
 dlprof = ["nvidia-pyindex", "nvidia-dlprof[pytorch]"]
 all = [
-  "pulser-core==0.18.1",
-  "pulser-simulation==0.18.1",
-  "pasqal-cloud==0.8.1",
-  "amazon-braket-sdk<1.71.2",
```
Roland-djee, line 81: Maybe we can add braket to the test env ?

Roland-djee (approved):
Thanks. Just a quick comment as Braket would still be used for checking noisy sims. Ping @EthanObadia

pyproject.toml
jpmoutinho: I just refactored what was there, it's the same as before. The test env uses `all` which still has `braket` on it.

PR #474, opened 2024-06-20
[Fix] GPSR on gpu

dominikandreasseitz:


qadence/backends/gpsr.py
```diff
@@ -48,7 +48,13 @@ def single_gap_psr(
     Returns:
         Tensor: tensor containing derivative values
     """
-
+    device = torch.device("cpu")
+    try:
+        device = [v.device for v in param_dict.values()][0]
+    except Exception:
+        pass
```
vytautas-a, line 55: I would add some explicit message here

jpmoutinho (approved):
Well if this fixes it let's merge it asap so we can release with the fixed docs.

qadence/backends/gpsr.py
dominikandreasseitz: its just a hotfix, we will move this logic to pyq soon

PR #477, opened 2024-06-21
[Refac] Fix saving/loading models with device

### LINKED ISSUE #475: [Bug] get_latest_checkpoint_name() looks for the wrong ckpt name
The function `get_latest_checkpoint_name()` is not able to retrieve the last checkpoint name for both optimizers and models. I believe the issue is in `write_checkpoint()` when the `model_checkpoint_name` and `opt_checkpoint_name` are defined. When the `device` was included in the checkpoint name, it should have been included before `iteration_substring` in the string names, so that it matches the pattern defined by `pattern = re.compile(".*_(\d+).pt$")` in `get_latest_checkpoint_name`.

This is, the model's checkpoint name should be:
```
model_checkpoint_name: str = (
        f"model_{type(model).__name__}_ckpt"
        + f"_device_{device}"
        + f"_{iteration_substring}"
        + ".pt"
    )
```
and same with the optimizer's checkpoint. 

The tests in `test_checkpointing.py` are passing because they do not correctly test for `get_latest_checkpoint_name()`. When a checkpoint is not found in the `load_model()` function, it just returns the `model` passed as an argument, so that when comparing the parameters of the model pre- and post- loading the checkpoint they are equal even if the saved model was really not loaded. Same with optimizers I believe. For example, when running `pytest tests/ml_tools/test_checkpointing.py -s` to print the warnings, the following type of warning shows:
```
WARNING 2024-06-20 12:45:16 - qadence.ml_tools.saveload: Unable to load state dict due to [Errno 21] Is a directory: '/tmp/pytest-of-ina/pytest-21/test_basic_save_load_ckpts0'.               No corresponding pre-trained model found. Returning the un-trained model.
ERROR 2024-06-20 12:45:16 - qadence.ml_tools.saveload: Directory /tmp/pytest-of-ina/pytest-21/test_basic_save_load_ckpts0 does not contain any opt checkpoints.
```

I can make a draft PR with the changes.

inafergra:
Fixes #475 following up on the work done by @dominikandreasseitz in this unfinished [PR](https://github.com/pasqal-io/qadence/pull/434). The unfinished PR also included some refactoring of the `TrainConfig` class (not related to this bug) which was causing conflicts with main that were messy to solve, thus this new PR. As discussed, it would be better to add the `TrainConfig` refactoring in a separate PR.


qadence/ml_tools/saveload.py
```diff
@@ -41,19 +46,24 @@ def load_checkpoint(
...
     return model, optimizer, iter
 
 
 def write_checkpoint(
-    folder: Path, model: Module, optimizer: Optimizer | NGOptimizer, iteration: int | str
+    folder: Path,
+    model: Module,
+    optimizer: Optimizer | NGOptimizer,
+    iteration: int | str,
+    device: str | torch.device = "cpu",
```
DanieleCucurachi, line 66: The rest of the updates are great! There’s still an issue in the `write_checkpoint()` function though: it sets `device` to `None` at the start, ignoring both the provided `device` parameter and the default “cpu”.

Due to the try-except structure, failure to retrieve the device from the pyqtorch native circuit goes unnoticed by the user, leading to misleading training logs:

```
INFO  2024-06-27 11:12:44 - qadence.ml_tools.saveload: Writing PINN checkpoint model_PINN_ckpt_199_device_None.pt
INFO  2024-06-27 11:12:44 - qadence.ml_tools.saveload: Writing SGD to checkpoint opt_SGD_ckpt_199_device_None.pt
```

To address this, I would consider:

1. Omitting `device=None`, allowing the function to use the `device` specified in the input if the pyqtorch native circuit extraction does not succeed. Note: this change would necessitate providing a `device` to the  `write_checkpoint` calls in `train()` to prevent defaulting to “cpu” when a GPU is in use.
2. Keeping the current setup but adding an error notification to alert the user when `device` is mistakenly set to `None`

DanieleCucurachi (commented):
Thank you @inafergra ! 

Overall looks good, I just pointed out a minor issue in the original qadence code that is still to be fixed.

I do still get an error message when using the QNG optimizer. However, it doesn’t cause a failure anymore, allowing the simulation to proceed:
```
ERROR 2024-06-27 11:11:16 - qadence.ml_tools.saveload: Can't pickle local object 'Parametric.__init__.<locals>.parse_values'
Traceback (most recent call last):
  File "/home/daniele_wsl/.local/share/hatch/env/virtual/error-analysis/w9Kxc4_-/error-analysis/lib/python3.10/site-packages/qadence/ml_tools/saveload.py", line 103, in write_checkpoint
    torch.save(
  File "/home/daniele_wsl/.local/share/hatch/env/virtual/error-analysis/w9Kxc4_-/error-analysis/lib/python3.10/site-packages/torch/serialization.py", line 628, in save
    _save(obj, opened_zipfile, pickle_module, pickle_protocol, _disable_byteorder_record)
  File "/home/daniele_wsl/.local/share/hatch/env/virtual/error-analysis/w9Kxc4_-/error-analysis/lib/python3.10/site-packages/torch/serialization.py", line 840, in _save
    pickler.dump(obj)
AttributeError: Can't pickle local object 'Parametric.__init__.<locals>.parse_values'
```

It may be worth considering whether to address this within the current thread or to open a separate issue for better tracking. What do you think?

dominikandreasseitz (dismissed):
thanks @inafergra ,
1. can we add a test which loads both legacy (without a device) and current checkpoints (with device) and asserts that they are successfully loaded?
2.  i remember @awennersteen proposed on the previous PR to handle both cases and log a error/debug in case the user uses legacy checkpoints. the most important thing is that this implementation handles both. 
thanks!

inafergra:
> thanks @inafergra ,
> 
>     1. can we add a test which loads both legacy (without a device) and current checkpoints (with device) and asserts that they are successfully loaded?
> 
>     2. i remember @awennersteen proposed on the previous PR to handle both cases and log a error/debug in case the user uses legacy checkpoints. the most important thing is that this implementation handles both.
>        thanks!

Thanks for the comments Dominik. I added tests checking that both QNNs and QuantumModels can be loaded with the legacy ckpts names. The function `get_latest_checkpoint_name()` can find both legacy and device ckpts.

tests/ml_tools/test_checkpointing.py
```diff
@@ -25,6 +27,33 @@ def dataloader(batch_size: int = 25) -> DataLoader:
...
 
+def write_legacy_checkpoint(
+    folder: Path,
+    model: torch.Module,
+    optimizer: Optimizer | NGOptimizer,
+    iteration: int | str,
+) -> None:
+    iteration_substring = f"{iteration:03n}" if isinstance(iteration, int) else iteration
+    model_checkpoint_name: str = (
+        f"model_{type(model).__name__}_ckpt_" + f"{iteration_substring}" + ".pt"
```
gvelikova, line 38: For these strings, I think that you can just rely on the f"model_{type(model).__name__}_ckpt_{iteration_substring}.pt" to do the job.

tests/ml_tools/test_checkpointing.py
```diff
@@ -58,7 +87,7 @@ def test_random_basicqQM_save_load_ckpts(BasicQuantumModel: QuantumModel, tmp_pa
 
     def loss_fn(model: QuantumModel, data: torch.Tensor) -> tuple[torch.Tensor, dict]:
         next(cnt)
-        out = model.expectation({})
+        out = model.expectation({}).squeeze(dim=0)
```
gvelikova, line 90: Would this work in all cases?

gvelikova:
Hey @inafergra have you tested how this modification with the device works with GPUs?

tests/ml_tools/test_checkpointing.py
inafergra: I think the `type` object does not have a `name` attribute? I get `AttributeError: type object 'QuantumModel' has no attribute 'name'. But I will change it to just a single f-string, it's clearer

tests/ml_tools/test_checkpointing.py
inafergra: I don't think so, just for this particular dummy loss, since we are using `torch.rand(1)` as the target. I changed so that the dimensions match, otherwise the expectation values returns a shape [1,1]. I could also reshape the target if that's clearer?

qadence/ml_tools/saveload.py
```diff
@@ -14,20 +14,25 @@
 logger = getLogger(__name__)
 
 
-def get_latest_checkpoint_name(folder: Path, type: str) -> Path:
+def get_latest_checkpoint_name(folder: Path, type: str, device: str | torch.device = "cpu") -> Path:
```
awennersteen, line 17: I think if we're to allow `torch.device` here we should also perform the same transformation as we do for writing it
```
str(model.device).split(":")[0]
```

inafergra:
> Hey @inafergra have you tested how this modification with the device works with GPUs?

Hey @gvelikova, yes I tested it in my local (single) gpu and it works fine

qadence/ml_tools/saveload.py
```diff
@@ -102,13 +113,18 @@ def write_checkpoint(
...
 def load_model(
-    folder: Path, model: Module, model_ckpt_name: str | Path = "", *args: Any, **kwargs: Any
+    folder: Path,
+    model: Module,
+    model_ckpt_name: str | Path = "",
+    device: str | torch.device = "cpu",
+    *args: Any,
+    **kwargs: Any,
 ) -> tuple[Module, int]:
     from qadence import QNN, QuantumModel
```
awennersteen, line 123: Personally I'd expect here to be able to send `device = cuda:4`, and have the model loaded on GPU 4. 
Since we store the model as `cuda` though, that means that we need to change the logic here. 

I would suggest to pass `device` to the torch.load function, and then we'll have to sort out how to deal with the  
```
model._from_dict
```
and
```
model.load_state_dict
```
parts I guess.

awennersteen: If it's not obvious how to do it we can make it a new issue, it's a new feature I suppose, although it's also a bug :p

qadence/ml_tools/train_grad.py
```diff
@@ -110,8 +110,11 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
     """
     # load available checkpoint
     init_iter = 0
+    log_device = "cpu" if device is None else device
```
awennersteen, line 113: @dominikandreasseitz in Qadence is this guaranteed?
Because it's not in "regular" PyTorch.

qadence/ml_tools/saveload.py
inafergra: Yes true!

qadence/ml_tools/saveload.py
inafergra: We could handle that by simply adding `model.to(device)` after `model._from_dict` and `model.load_state_dict` maybe? That way, if e.g. `cuda:4` is passed then the model would be found and loaded no matter in which GPU the model is saved (since we only save and look for `cuda` when loading) and then it would be moved into GPU 4

qadence/ml_tools/train_grad.py
inafergra: What do you mean by guaranteed? I think this is needed because the default value for `device` argument in the `train_grad()` function is `None`, so with this if the user does not pass a device to `train_grad` it will just try to load a ckpt from `cpu`. Or do you mean there could a problem with the `if` statement itself?

awennersteen: device of `None` in PyTorch means don't change it.
So you have a tensor on GPU, and move it to `None` it remains on the same GPU
```python
awennersteen:~ $ python3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> a = torch.rand(5, device="cuda")
>>> a.to(None)
tensor([0.0810, 0.8175, 0.8455, 0.5090, 0.4081], device='cuda:0')
>>> 

```

awennersteen: So in general, someone could setup their code in advance to work on a certain GPU, and e.g. set `torch.default_device` to `"cuda:0"` and pass `device=None` but have the data on GPU, not cpu.

I just don't know if that is true on Qadence since we do mess around a bit with the parameters and such, and I hope @dominikandreasseitz knows

qadence/ml_tools/saveload.py
awennersteen: Yeah, it should probably be sufficient.

tests/ml_tools/test_checkpointing.py
inafergra: I changed it in this [commit](https://github.com/pasqal-io/qadence/pull/477/commits/66351310c2c8891a6f198cbedad681d721f2b5ac)

awennersteen: Yeah,  but why?
I guess because the tests failed, but I don' t understand why

inafergra: Tests were not failing but without squeezing they were throwing this kind of warning:
` UserWarning: Using a target size (torch.Size([1])) that is different to the input size (torch.Size([1, 1])). This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.
    return F.mse_loss(input, target, reduction=self.reduction)`

inafergra: I just changed it to get rid of the warning

dominikandreasseitz:
hey @inafergra , @Roland-djee will take over the review here!

qadence/ml_tools/train_grad.py
Roland-djee: I think it is. All examples of setting the device are mainly in the [examples](https://github.com/search?q=repo%3Apasqal-io%2Fqadence%20default_device&type=code). But I would seek for @dominikandreasseitz on this.

tests/ml_tools/test_checkpointing.py
```diff
@@ -25,6 +27,29 @@ def dataloader(batch_size: int = 25) -> DataLoader:
     return to_dataloader(x, y, batch_size=batch_size, infinite=True)
 
 
+def write_legacy_checkpoint(
+    folder: Path,
+    model: torch.Module,
+    optimizer: Optimizer | NGOptimizer,
+    iteration: int | str,
+) -> None:
+    iteration_substring = f"{iteration:03n}" if isinstance(iteration, int) else iteration
```
Roland-djee, line 36: Out of curiosity, is this a torch requirement that the interation is of that format ?

tests/ml_tools/test_checkpointing.py
```diff
@@ -149,3 +174,26 @@ def loss_fn(model: QuantumModel, data: torch.Tensor) -> tuple[torch.Tensor, dict
     for ckpt in ckpts:
         loaded_model, optimizer, _ = load_checkpoint(tmp_path, BasicQNN, optimizer, ckpt, "")
         assert torch.allclose(loaded_model.expectation(inputs), model.expectation(inputs))
+
+
+def test_basicqQM_save_load_legacy_ckpts(BasicQuantumModel: QuantumModel, tmp_path: Path) -> None:
```
Roland-djee, line 179:
```suggestion
def test_basic_qm_save_load_legacy_ckpts(BasicQuantumModel: QuantumModel, tmp_path: Path) -> None:
```

Roland-djee (approved):
Thanks for this @inafergra. Can't say much since I'm less versed in torch intricacies than @gvelikova or @awennersteen (but slowly curing myself). So I let them be better judges at this point. Approving now but subjected to their final approval.

qadence/ml_tools/saveload.py
```diff
@@ -102,20 +114,28 @@ def write_checkpoint(
...
+        model_ckpt_name = get_latest_checkpoint_name(folder, "model", device)
 
     try:
         iteration, model_dict = torch.load(folder / model_ckpt_name, *args, **kwargs)
         if isinstance(model, (QuantumModel, QNN)):
             model._from_dict(model_dict, as_torch=True)
         elif isinstance(model, Module):
             model.load_state_dict(model_dict, strict=True)
+        # Load to a specific gpu device if specified
+        if ":" in str(device):
```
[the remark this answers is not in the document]
gvelikova: Would `if not str(device).isalpha()` also work to distinguish between gpu with specific device number (like cuda:0) and cpu/gpu as strings?

qadence/ml_tools/saveload.py
```diff
@@ -14,20 +14,26 @@
...
         logger.error(f"Directory {folder} does not contain any {type} checkpoints.")
     if len(files) == 1:
         file = Path(files[0])
     else:
-        pattern = re.compile(".*_(\d+).pt$")
+        device = str(device).split(":")[0]
+        pattern = re.compile(f".*_(\d+)_device_{device}.pt$")
+        legacy_pattern = re.compile(".*_(\d+).pt$")
         max_index = -1
         for f in files:
```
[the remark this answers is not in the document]
inafergra: `file` is already the name of a variable previously defined:
https://github.com/pasqal-io/qadence/blob/9b3346467a0c182568c69a06596ca98ab5b69e3d/qadence/ml_tools/saveload.py#L18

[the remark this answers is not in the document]
Roland-djee: Meh, indeed. And it might be a reserved keyword. Just ignore please.

qadence/ml_tools/saveload.py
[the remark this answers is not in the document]
inafergra: True, using `re` now to match the string

tests/ml_tools/test_checkpointing.py
inafergra: Changed the name of all tests to match this style as well

tests/ml_tools/test_checkpointing.py
inafergra: I'm not sure but I'd say it's more of a style choice rather than a pytorch requirement. My understanding is that any string can be used as a ckpt name. But @dominikandreasseitz can correct me if wrong since he wrote the original line

qadence/ml_tools/saveload.py
```diff
@@ -63,8 +73,12 @@ def write_checkpoint(
     try:
         # We extract the device from the pyqtorch native circuit
         device = str(model.device).split(":")[0]  # in case of using several CUDA devices
-    except Exception:
-        pass
+    except Exception as e:
+        msg = (
+            f"Unable to identify in which device the QuantumModel is stored due to {e}."
+            + "Setting device to None"
```
gvelikova, line 79: I think you can do a multiline string without the need of the + here.

gvelikova (approved):

qadence/ml_tools/saveload.py
inafergra: True, thanks!

inafergra:
@awennersteen if you don't have any more comments this should be ready to merge now

inafergra:
Merging this now as discussed with @gvelikova, any additional changes can be included in a future PR solving #484 which also involves checkpointing.

Issue #368, opened 2024-03-13
[Refactoring] Avoid calling block to tensor on diagonal Hamiltonians


jpmoutinho:
When calling `HamEvo` on diagonal Hamiltonians the dense matrix is built but then only the diagonal is taken. 

This is currently done in the `PyQHamiltonianEvolution` class in [`qadence/backends/pyqtorch/convert_ops.py`](https://github.com/pasqal-io/qadence/blob/main/qadence/backends/pyqtorch/convert_ops.py), where the `__init__` method first builds the full matrix `hmat`, and then the `_unitary` method defines the evolution function for diagonal operators which runs `diagonal()` on the full hamiltonian matrix.

We should instead build only the diagonal. For non-parametric block this is directly possible with the function `block_to_diagonal`, as done in the `PyQObservable` class in the same file.

Handling parametric blocks is out of the scope of this issue, and can be done separately. It requires further development in the `block_to_tensor` side as detailed in https://github.com/pasqal-io/qadence/issues/394.

NOTE: Due to ongoing changes in the codebase, we might need to implement this directly in the PyQTorch backend (https://github.com/pasqal-io/pyqtorch/issues/175).

Roland-djee:
@jpmoutinho It feels like it's been handled somewhere unless I'm wrong ?

neogyk:
Dear all,
I am trying to solve this issue.
Is it will be enough to add the following lines to the `PyQHamiltonianEvolution` class constructor:

`        elif not block.generator.is_parametric and block.generator.diagonal:
            m = block.generator.to(dtype=cdouble)
            hmat = block_to_diagonal(
                MatrixBlock(m, qubit_support=block.qubit_support),
                qubit_support=self.qubit_support,
                use_full_support=False,
            )
            hmat = hmat.permute(1, 2, 0)
            self.register_buffer("hmat", hmat)
            self._hamiltonian = lambda self, values: self.hmat`

jpmoutinho:
Hey @neogyk, thanks for your interest in completing this issue!

The solution should be along those lines, but there is no need to use `MatrixBlock`. You should be able to run `block_to_diagonal` on the `block.generator`.

Feel free to open an MR and add me as a reviewer, then we can discuss it there.

neogyk:
Dear @jpmoutinho, could you please review the changes. I made the PR - #441

Roland-djee:
> Dear @jpmoutinho, could you please review the changes. I made the PR - #441

[...]

PR #441, opened 2024-06-05
[Refac] PyQHamiltonianEvolution - Added the case for diagonal Hamiltonian

neogyk:
Refactored the `PyQHamiltonianEvolution`. Added the case for diagonal Hamiltonian  #368

PR #478, opened 2024-06-23
[Refactor] Refactor the PyQHamiltonianEvolution. 

neogyk:
[Issue #368](https://github.com/pasqal-io/qadence/issues/368#issue-2183506409)
Continuation of the [#441 ](https://github.com/pasqal-io/qadence/pull/441#issue-2335666907)


.github/workflows/pylint.yml
jpmoutinho, line 1: This needs to be removed, we already have a linting workflow

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -350,13 +362,19 @@ def _hamiltonian(self: PyQHamiltonianEvolution, values: dict[str, Tensor]) -> Te
...
+            hamiltonian = diagonal(hamiltonian) if bool(diag_check) else hamiltonian
+        else:
+            self.batch_size = None
+            diag_check = tensor([True])
 
         def _evolve_diag_operator(hamiltonian: Tensor, time_evolution: Tensor) -> Tensor:
-            evol_operator = diagonal(hamiltonian) * (-1j * time_evolution).view((-1, 1))
+            evol_operator = hamiltonian * (-1j * time_evolution).view((-1, 1))
             evol_operator = diag_embed(exp(evol_operator))
             return transpose(evol_operator, 0, -1)
```
jpmoutinho, line 379: The logic here needs to be reworked a bit, and probably requires a bit of a refactor. There is no longer a need to use the `is_diag` imported from PyQTorch because you already checked for diagonality with the `block.generator._is_diag_pauli`. It seems to me like the `_evolve_diag_operator` needs to be decoupled from the `_unitary` method somehow. The point is that we need to avoid doing  `diag_embed(exp(evol_operator))` because that will instatiate a dense full matrix from the diagonal, which we don't want. We want to use the diagonal vector directly when applying it to the state through a vector - vector product.

jpmoutinho (commented):
Thanks for the effort @neogyk. It seems like this still requires a bit of work to properly fix the issue :)

qadence/backends/pyqtorch/convert_ops.py
neogyk: @jpmoutinho, yes I agree. 
In case of diagonal hamiltonian, we have 1-d vector representation, where evolution operator can be simply calculated as exponent function and applied to specific state through multiplication, in other way the state will be evolved using the function` pyq.apply_operator`.
This logic can be implemented in the `self.forward` function.

.github/workflows/pylint.yml
neogyk: @jpmoutinho, Ok I will remove.

tests/backends/pyq/test_quantum_pyq.py
```diff
@@ -844,3 +846,12 @@ def test_move_to_dtype(
...
     assert expval.dtype == torch.float64 if dtype == torch.cdouble else torch.float32
+
+
+
+@pytest.mark.parametrize("gen", [Z(0) + Z(1), Z(0) + Z(1) + X(0) - X(1), Z(0) + Z(1) + X(0) - X(0)])
+def test_hamiltonian_evolution(gen: AbstractBlock) -> None:
+    state = run(HamEvo(gen, torch.tensor([1.0])))
+    assert type(state) == torch.Tensor
+    assert state.dtype == torch.complex128
+    assert not torch.any(torch.isnan(state))
```
jpmoutinho, line 857: No need to include this as a test, there are already many tests that run `HamEvo`.

Also it seems this is conflicting with the main branch. Probably it will be solved once you remove the code here, but try to rebase anyway.

neogyk: Ok, it will be also removed

jpmoutinho:
Hi @neogyk, a quick update: we are going to close the issue #368 and give you credit for the completion, since what you have done so far in this MR is close to finishing the issue.

However, we will not be able to merge this MR. The extended deadline of the unitaryhack has clashed with some other work that we are doing on this file, and the code that you are changing here is going to be ported elsewhere. Once we complete that port, we will also port your solution :)

That being said, I will leave this MR open for a few more days so you are welcome to add more commits if you wish, and I will reply as best as I can.

PR #480, opened 2024-06-25
[Docs] Add digital-analog qcl tutorial

jpmoutinho:


qadence/register.py
```diff
@@ -272,14 +356,6 @@ def __eq__(self, other: object) -> bool:
...
-
-    Args:
-        n_qubits (int): number of nodes in the graph
-
-    Returns:
-        graph instance
-    """
     graph = nx.Graph()
     for i in range(n_qubits):
         graph.add_node(i, pos=(i, 0.0))
```
Doomsk, line 361: should the docstring be cut off from the function?

Doomsk (approved):
very nice @jpmoutinho ! Just one comment, but it's good to go

qadence/register.py
jpmoutinho: The reason I removed it was just so it didn't show up in the API reference. This is an auxiliary function that actually should be private, I just didn't feel like refactoring the whole code.

PR #482, opened 2024-06-27
[Fix] Handle nested parameter dicts

dominikandreasseitz:
Related to https://github.com/pasqal-io/pyqtorch/pull/204. Avoid inferring batch_size from dicts when tensors are expected


vytautas-a (approved):
Looks good

PR #490, opened 2024-07-02
[Feature] Add backend, diff mode etc to config constructor

smitchaudhary:
- [x] Update documentation with an example of passing diffmode and backend to QNN.from_configs


qadence/ml_tools/constructors.py
```diff
@@ -762,13 +791,28 @@ def build_qnn_from_configs(
...
+        inputs=inputs,
+        backend=backend,
+        diff_mode=diff_mode,
+        measurement=measurement,
+        noise=noise,
+        configuration=configuration,
+        input_diff_mode=input_diff_mode,
+    )
 
     return ufa
```
awennersteen, line 818: Sorry for hijacking but I think this function is becoming a bit hard to read. I propose to change the whole body to:

```python
    if observable_config is None:
        raise ValueError(
            "You need to provide at least one observable config in the QNN constructor"
        )

    if ansatz_config is None:
        ansatz_config = AnsatzConfig(depth=0)

    inputs: list[Basic | str] | None = None
    full_fm = chain()
    if fm_config:
        fm_blocks = create_fm_blocks(register=register, config=fm_config)
        full_fm = _interleave_ansatz_in_fm(
            register=register,
            fm_blocks=fm_blocks,
            ansatz_config=ansatz_config,
        )
        inputs = fm_config.inputs

    ansatz = create_ansatz(register=register, config=ansatz_config)

    if ansatz_config.ansatz_strategy == "rydberg":
        # Add a block before the Feature Map to move from 0 state to an
        # equal superposition of all states. This needs to be here only for rydberg
        # feature map and only as long as the feature map is not updated to include
        # a driving term in the Hamiltonian.

        num_qubits = register if isinstance(register, int) else register.n_qubits
        mixing_block = kron(*[H(i) for i in range(num_qubits)])
        full_fm = chain(mixing_block, full_fm)

    circ = QuantumCircuit(register, full_fm, ansatz)

    observable: AbstractBlock | list[AbstractBlock] = (
        [observable_from_config(register=register, config=cfg) for cfg in observable_config]
        if isinstance(observable_config, list)
        else observable_from_config(register=register, config=observable_config)
    )

    return QNN(
        circ,
        observable,
        inputs=inputs,
        backend=backend,
        diff_mode=diff_mode,
        measurement=measurement,
        noise=noise,
        configuration=configuration,
        input_diff_mode=input_diff_mode,
    )
```

awennersteen: Although I don't quite understand why fm_config and ansatz_config and observable_config should be allowed to be None. I think we should just not allow for `None`, and then the above simplifies even further

awennersteen:
```python
    fm_blocks = create_fm_blocks(register=register, config=fm_config)
    full_fm = _interleave_ansatz_in_fm(
        register=register,
        fm_blocks=fm_blocks,
        ansatz_config=ansatz_config,
    )
    ansatz = create_ansatz(register=register, config=ansatz_config)

    if ansatz_config.ansatz_strategy == "rydberg":
        # Add a block before the Feature Map to move from 0 state to an
        # equal superposition of all states. This needs to be here only for rydberg
        # feature map and only as long as the feature map is not updated to include
        # a driving term in the Hamiltonian.

        num_qubits = register if isinstance(register, int) else register.n_qubits
        mixing_block = kron(*[H(i) for i in range(num_qubits)])
        full_fm = chain(mixing_block, full_fm)

    circ = QuantumCircuit(register, full_fm, ansatz)

    observable: AbstractBlock | list[AbstractBlock] = (
        [observable_from_config(register=register, config=cfg) for cfg in observable_config]
        if isinstance(observable_config, list)
        else observable_from_config(register=register, config=observable_config)
    )

    return QNN(
        circ,
        observable,
        inputs=fm_config.inputs,
        backend=backend,
        diff_mode=diff_mode,
        measurement=measurement,
        noise=noise,
        configuration=configuration,
        input_diff_mode=input_diff_mode,
    )
```

smitchaudhary: @awennersteen 
> I think this function is becoming a bit hard to read.
Agreed 100%. The review is much appreciated. Many ad-hoc changes were made to this function.

> Although I don't quite understand why fm_config and ansatz_config and observable_config should be allowed to be None.
It wasn't allowed to be None before today. It is the latest addition. Users using it for Quantum Extremal Learning or other such cases, would want to be able to define a QNN without the feature encoding block.

An alternative is to require the user to specify that the number of features being encoded is 0 and then not have any of these fm_blocks. But with the None option, the user can just not provide a config for the feature map if they don't want any feature map.

Another option, is to make that the default be a feature map with 0 features (and thus have an empty block). The current default is 1 feature. But having a default feature map, which doesn't even do anything could be a bit confusing.

Thoughts?

dominikandreasseitz:
hey @smitchaudhary , @Roland-djee will take over the review here !

docs/tutorials/qml/config_qnn.md
```diff
@@ -89,16 +89,18 @@ For full details on the `ObservableConfig` class, see the [API documentation][qa
 
 ## Defining the QNN from the Configs
 
-To build the QNN, we can now use the `QNN` class.
+To build the QNN, we can now use the `QNN` class. In addition to the three parts of the QNN, we can also specify the `backend`, `diff_mode`, etc. For full details on the `QNN` class, see the [API documentation][qadence.ml_tools.models.QNN] or the documentation on the config constructor [here][qadence.ml_tools.models.QNN.from_configs].
```
Roland-djee, line 92: What are the three parts you're referring to ?

```suggestion
To build the QNN, we can now use the `QNN` class as a `QuantumModel` subtype. In addition to the three parts of the QNN, we can also specify options such as the `backend`, `diff_mode`, etc. For full details on the `QNN` class, see the [API documentation][qadence.ml_tools.models.QNN] or the documentation on the config constructor [here][qadence.ml_tools.models.QNN.from_configs].
```

qadence/ml_tools/constructors.py
```diff
@@ -733,42 +746,68 @@ def build_qnn_from_configs(
...
     """
-    fm_blocks = create_fm_blocks(register=register, config=fm_config)
-    full_fm = _interleave_ansatz_in_fm(
-        register=register,
-        fm_blocks=fm_blocks,
-        ansatz_config=ansatz_config,
-    )
 
-    ansatz = create_ansatz(register=register, config=ansatz_config)
+    if observable_config is None:
```
Roland-djee, line 762: Why making `observable_config` optional if you check existence later ? Wouldn't it be simpler to keep it as positional argument ?

qadence/ml_tools/constructors.py
```diff
@@ -733,42 +746,68 @@ def build_qnn_from_configs(
...
+        raise ValueError(
+            "You need to provide at least one observable config in the QNN constructor."
+        )
+
+    if ansatz_config is None:
+        ansatz_config = AnsatzConfig(depth=0)
+
+    inputs: list[Basic | str] | None = None
+    full_fm = chain()
```
Roland-djee, line 771: Do you need to instantiate `full_fm` as an empty chain ?

qadence/ml_tools/constructors.py
```diff
@@ -733,42 +746,68 @@ def build_qnn_from_configs(
...
+        inputs = fm_config.inputs
 
-    # Add a block before the Featuer Map to move from 0 state to an
-    # equal superposition of all states. This needs to be here only for rydberg
-    # feature map and only as long as the feature map is not updated to include
-    # a driving term in the Hamiltonian.
+    ansatz = create_ansatz(register=register, config=ansatz_config)
 
     if ansatz_config.ansatz_strategy == "rydberg":
+        # Add a block before the Featuer Map to move from 0 state to an
```
Roland-djee, line 784:
```suggestion
        # Add a block before the Feature Map to move from 0 state to an
```

qadence/ml_tools/constructors.py
Roland-djee: I had somewhat the same though as @awennersteen and commented about the `observable_config`. I think that if it makes the logic simpler, we should avoid optional args that are unnecessarily checked for existence except for the `FeatureMapConfig` if there is a user need as you pointed out. Then, it needs to be moved after the defaulted arguments which might be confusing. In this case, I'd be inclined to default it to 0 features, check for it and skip adding an empty block.

Roland-djee (commented):
Thanks @smitchaudhary. Comments from my side but generally would agree with @awennersteen.

docs/tutorials/qml/config_qnn.md
smitchaudhary: Thanks. Clarified the language.

qadence/ml_tools/constructors.py
smitchaudhary: @Roland-djee Thanks for the review!

The reason why the these configs default to `None` and then checked, although a bit ugly, is there to maintain the FM, ansatz, observable order.

And some users have asked to not have to include FM in their QNNs.

smitchaudhary:
@Roland-djee 
> I'd be inclined to default it to 0 features, check for it and skip adding an empty block.

Though forcing a default behavior for the fm_config, and thus, moving it after `observable_config` would be potentially confusing to the user I am afraid.

Or do you mean, make it a non-optional argument, thus preserving the FM->ansatz->Observable order, and force the user to provide a `FeatureMapConfig` object (which could then default to 0 features), even in cases where they don't want a FM>

qadence/ml_tools/constructors.py
Roland-djee: Yeah I got that point. But I guess we can get away without making them optional, as discussed below.

Roland-djee:
> @Roland-djee
> 
> > I'd be inclined to default it to 0 features, check for it and skip adding an empty block.
> 
> Though forcing a default behavior for the fm_config, and thus, moving it after `observable_config` would be potentially confusing to the user I am afraid.
> 
> Or do you mean, make it a non-optional argument, thus preserving the FM->ansatz->Observable order, and force the user to provide a `FeatureMapConfig` object (which could then default to 0 features), even in cases where they don't want a FM?

Yes, that's the solution I see. Then we can check the number of features and skip the empty block in the `full_fm` creation.

Roland-djee:
> > @Roland-djee
> > > I'd be inclined to default it to 0 features, check for it and skip adding an empty block.
> > 
> > 
> > Though forcing a default behavior for the fm_config, and thus, moving it after `observable_config` would be potentially confusing to the user I am afraid.
> > Or do you mean, make it a non-optional argument, thus preserving the FM->ansatz->Observable order, and force the user to provide a `FeatureMapConfig` object (which could then default to 0 features), even in cases where they don't want a FM?
> 
> Yes, that's the solution I see. Then we can check the number of features and skip the empty block in the `full_fm` creation.

Maybe a suggestion would be to use a datastructure for handling configs ? Could be an overkill though.

qadence/ml_tools/constructors.py
smitchaudhary: Thanks. Removed this to only include `full_fm` in a `blocks` list.

smitchaudhary:
Thanks @awennersteen and @Roland-djee for the comments.

I have made the following changes:

1. FeatureMapConfig now defaults to `num_features=0` and if a user does not want an FM, they can pass `FeatureMapConfig`. The downside being that a user who wants to have a single feature, will actively have to specify `num_features=1` (which is good practice already :P )
2. The default behavior of the `AnsatzConfig` is to have `depth=1` (all other ansatz constructors have this default, thus made the most sense).
3. Removed the ad-hoc block that changed the initial state in case of `RYDBERG` strategy being used. This was there due to the issue I raised in `qadence-libs` [here](https://github.com/pasqal-io/qadence-libs/issues/15). But I think, it is a lot more transparent to have the user specify the `state` when calling `forward`.
4. Modified tests to account for this changes.
5. Some other stylistic changes + fixing typos.

qadence/ml_tools/constructors.py
```diff
@@ -724,6 +731,12 @@ def build_qnn_from_configs(
     fm_config: FeatureMapConfig,
```
Roland-djee, line 731: Would it make sense to:

```suggestion
    fm_config: FeatureMapConfig = FeatureMapConfig(),
```

Hence you don't pass it if you don't need it.

tests/ml_tools/test_qnn.py
```diff
@@ -347,9 +347,46 @@ def test_variational_transformed_module(
     assert torch.allclose(tm_pred, (output_range[0] * pred) + output_range[1])
 
 
-def test_config_qnn_input_transform() -> None:
+def test_config_empty_fm() -> None:
     fm_config = FeatureMapConfig()
```
Roland-djee, line 351: If you default it, I don't think this would be necessary anymore.

Roland-djee (commented):
Thanks @smitchaudhary. More minor comments but looking nice. Will approve as soon as you've addressed them.

tests/ml_tools/test_qnn.py
smitchaudhary: Opps! True. Missed the point of the test altogether 😅

qadence/ml_tools/constructors.py
Roland-djee: Ah, yeah I always forget about Python idiosyncrasies for parameters.

smitchaudhary: Decided to keep it as is for now. Let me know if you would rather have it any other way.

smitchaudhary: Ok. Gathering quick feedback for deciding on one or the other. Will let the users decide. Don't mind either one.

smitchaudhary: @Roland-djee Made the relevant changes. Changed to order of arguments and made an empty default for the FM.

Roland-djee:
@smitchaudhary Feel free to request a review.

Roland-djee (approved):
Thanks @smitchaudhary. I would also seek for @awennersteen's approval but since he works Tue & Wed and fi you need that earlier, you could merge. Otherwise, let's see with him too.

smitchaudhary:
Re-requested his review for now.

smitchaudhary:
I will merge it today and bump the version to 1.7.1 with all the changes made in the last 3 weeks.

PR #496, opened 2024-07-11
[Bug] Fix backend import if in extensions.

Roland-djee:
Fix failing backend import when in extensions. Should fix the failing CI in extensions.


smitchaudhary (approved):
Thanks @Roland-djee 

Checked locally too. It fixes the problem.

PR #499, opened 2024-07-15
[Feature] Allow loss_fn in train_grad to return arbitrary number of values.

### LINKED ISSUE #498: [Bug] train_grad should not force the number of return values on loss_fn used
Now there is functionality to [print last value of the loss and the metrics](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/train_grad.py#L248) in `train_grad` but this enforces `loss_fn` to return `loss` and `metrics`. Similar return values are expected in case of [performing validation](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/train_grad.py#L227).

Before these changes, the `loss` and `metrics` were obtained only from `optimize_step`. Which allowed the user to define any custom `loss_fn` and a compatible `ooptimize_step`. Thus, the user can define any optimization strategy within `optimize_step` that potentially takes into account more than just `loss` and `metrics`.

smitchaudhary:
Fixes #498

### DIFF
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -59,7 +59,7 @@ use `train_with_grad` as example but the code can be used directly with the grad
 As every other training routine commonly used in Machine Learning, it requires
 `model`, `data` and an `optimizer` as input arguments.
 However, in addition, it requires a `loss_fn` and a `TrainConfig`.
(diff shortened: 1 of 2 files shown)


smitchaudhary:
It fixes #498 

However, not the best solution as it still enforces a structure on `loss_fn` (first return value being `loss` and the second return value being `metrics`).

chMoussa (approved):
Lgtm

Roland-djee (approved):
Thanks @smitchaudhary. Looking good.

PR #509, opened 2024-07-23
Move MlFlow to required dependency

### LINKED ISSUE #508: [Bug] Fix MLFlow import
`mlflow` is listed as an optional dependency and thus is not installed unless specified. However, it is imported from in `ml_tools.printing` which throws an error when importing `qadence`.

smitchaudhary:
Fixes #508

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ dependencies = [
     "pyyaml",
     "matplotlib",
     "Arpeggio==2.0.2",
+	"mlflow",
 ]
 
 [tool.hatch.metadata]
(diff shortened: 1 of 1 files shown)


awennersteen (changes requested):
I propose to close this in favour of
https://github.com/pasqal-io/qadence/pull/510

There I also fixed another issue I was hitting.

Reasing being that I don't think that MLFlow should be a required dependency, and the fix is very simple to keep it optional

smitchaudhary:
> I propose to close this in favour of #510
> 
> There I also fixed another issue I was hitting.
> 
> Reasing being that I don't think that MLFlow should be a required dependency, and the fix is very simple to keep it optional

Yes. Perfect. I will close this. Thanks.

PR #511, opened 2024-07-25
[FIX] Allow 0 to be passed as a config to print,write,plot,checkpoint_every and disable in that case

awennersteen:
This can be desired behavior in some cases, and is perhaps a better behaviour than the current where we end up raising DivideByZero errors.

We can perhaps use the `__post_int__` of the `TrainConfig` to do more validation here (like negative values), and print a warning?

Not sure what we think is best.

### DIFF
--- a/qadence/ml_tools/train_grad.py
+++ b/qadence/ml_tools/train_grad.py
@@ -5,16 +5,12 @@ import math
 from logging import getLogger
 from typing import Callable, Union
 
-from rich.progress import (
-    BarColumn,
-    Progress,
-    TaskProgressColumn,
-    TextColumn,
-    TimeRemainingColumn,
-)
-from torch import complex128, float32, float64
+from rich.progress import (BarColumn, Progress, TaskProgressColumn, TextColumn,
+                           TimeRemainingColumn)
+from torch import complex128
 from torch import device as torch_device
 from torch import dtype as torch_dtype
+from torch import float32, float64
 from torch.nn import DataParallel, Module
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader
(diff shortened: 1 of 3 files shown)


Roland-djee (commented):
Thanks @awennersteen. Stupid question: does it make sense to default it to 0 and only check if it is non-negative ?

awennersteen:
> Thanks @awennersteen. Stupid question: does it make sense to default it to 0 and only check if it is non-negative ?

I have no idea. My guess is that users usually will want these to be printed?

qadence/ml_tools/train_grad.py
```diff
@@ -231,18 +225,22 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
                     print_metrics(loss, metrics, iteration - 1)
 
-                if iteration % config.write_every == 0:
+                if config.write_every > 0 and iteration % config.write_every == 0:
                     write_tracker(
                         writer, loss, metrics, iteration, tracking_tool=config.tracking_tool
                     )
 
-                if iteration % config.plot_every == 0:
+                if config.plot_every > 0 and iteration % config.plot_every == 0:
```
Roland-djee, line 243: Is it possible to group them under a single `config.print_every > 0` check ?

qadence/ml_tools/train_no_grad.py
```diff
@@ -114,13 +108,13 @@ def _update_parameters(
...
-            if iteration % config.print_every == 0 and config.verbose:
+            if config.print_every > 0 and iteration % config.print_every == 0 and config.verbose:
                 print_metrics(loss, metrics, iteration)
 
-            if iteration % config.write_every == 0:
+            if config.write_every > 0 and iteration % config.write_every == 0:
                 write_tracker(writer, loss, metrics, iteration, tracking_tool=config.tracking_tool)
 
-            if iteration % config.plot_every == 0:
+            if config.plot_every > 0 and iteration % config.plot_every == 0:
```
Roland-djee, line 117: Same.

Roland-djee (approved):
Thanks @awennersteen. LGTM. Minor stylistic comments.

qadence/ml_tools/train_grad.py
Roland-djee: Ah yeah stupid me ofc.

PR #515, opened 2024-07-29
AnsatzConfig Updates, Correct

giovanniminutoNL:
Resolution mismatch in custom rotations: strategy_args.operations [definition](https://github.com/pasqal-io/qadence/blob/813d48dcc39c869b07457cddefa292e3f594d2b9/qadence/ml_tools/config.py#L403) does not match _create_hea_digital request for ["rotations" ](https://github.com/pasqal-io/qadence/blob/813d48dcc39c869b07457cddefa292e3f594d2b9/qadence/ml_tools/constructors.py#L490).


smitchaudhary:
Thanks @giovanniminutoNL. This looks good. Maybe it would be a good idea to also do another small cleanup in the same PR?

Renaming the function [here](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/constructors.py#L559) and thus also when it is called [here](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/constructors.py#L613)?

Instead of it being called `_create_hea_ansatz` it should simply be `_create_hea` since the 'a' in 'hea' is already for ansatz. It is an oversight on my end when creating this.

smitchaudhary:
Nice work @giovanniminutoNL! Thanks. And congrats on your first contribution to Qadence.

giovanniminutoNL:
thanks to you and @n-toscano for the guidance

PR #519, opened 2024-08-01
[Fix] Consistent naming and using of StrEnums instead of string literals

### LINKED ISSUE #518: [Bug] Inconsistent naming in StrEnums and constructors
Bug:
Currently, if trying to use serial encoding for multivariate functions, it would throw an error because the StrEnum for `MultivariateStrategy` has `SERIES` as the string instead of `series`. This can be fixed by not checking against a literal but only against the `MultivariateStrategy`.

Cosmetic and consistency related changes:

The naming of some of the StrEnums is inconsistent with the others. Additionally, there was even an inconsistency within a single Enum, namely the `MultivariateStrategy`.

Should only use `StrEnum`s and not the string literals.

smitchaudhary:
Fixes #518


chMoussa (changes requested):
To not forget to harmonize where previous strings were used. For instance `get_fm_qubits` in `utils.py`

smitchaudhary:
> To not forget to harmonize where previous strings were used. For instance `get_fm_qubits` in `utils.py`

Hmm. Good catch. This constructor is independent of the configs created here, but very similar use indeed. I will add it, and also rename `fm_strategy` to `multivariate_strategy` since that is what that property is called in `types`. It might be a slight breaking change for someone using this constructor (However, I don't think it is being used much, so should not cause too much bother on the user side).

chMoussa (approved):

PR #531, opened 2024-08-12
[Refactor] Use PyQ's time-independent Hamiltonian evolution

chMoussa:


qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -94,29 +100,43 @@ def convert_block(
...
+            )
+            return [
+                pyq.HamiltonianEvolution(
+                    qubit_support=qubit_support,
+                    generator=generator,
+                    time=time_param,
+                    generator_parametric=is_parametric,  # type: ignore[union-attr]
+                    cache_length=0,
+                )
+            ]
```
jpmoutinho, line 138: Hmm, in the `pyq.HamiltonianEvolution` some time ago I added this condition:
```
      elif isinstance(generator, (QuantumOperation, Sequence)):
            if qubit_support is not None:
                logger.warning(
                    "Taking support from generator and ignoring qubit_support input."
                )
(...)
```
I am wondering if we will run into things in Qadence where this warning will constantly appear? But in principle not I think, since `HamEvo` also has `qubit_support = None` by default.

jpmoutinho (approved):
Great, thanks a lot :)

qadence/backends/pyqtorch/convert_ops.py
chMoussa: In principle not. But we can remove the warning logger later.

Issue #202, opened 2023-11-20
[Refactoring] Alternative ways of defining tensorboard metrics


atiyo:
At the moment, `train` accepts a function `write_tensorboard(writer, loss, metrics, iteration)`, where metrics is returned from a custom loss function.

However, we might want to log non-training related metrics, e.g. logging a plot, or logging a mean squared error to a known benchmark. These sorts of metrics are a) potentially expensive to calculate at every training step and b) conceptually independent to the training loop.

It would be cool if there was an alternative way of logging custom metrics etc. to tensorboard without having to do so via the loss function.

nmheim:
I think the most general way of implementing this would be to redefine our `train_with_grad` function and do:
```python
    # outer epoch loop
    for iteration in progress.track(range(init_iter, init_iter + config.max_iter)):
        try:
            # in case there is not data needed by the model
            # this is the case, for example, of quantum models
            # which do not have classical input data (e.g. chemistry)
            if dataloader is None:
                loss, metrics = optimize_step(model, optimizer, loss_fn, None)
                loss = loss.item()

            elif isinstance(dataloader, (DictDataLoader, DataLoader)):
                data = data_to_device(next(dl_iter), device)  # type: ignore[arg-type]
                loss, metrics = optimize_step(model, optimizer, loss_fn, data)

            else:
                raise NotImplementedError(
                    f"Unsupported dataloader type: {type(dataloader)}. "
                    "You can use e.g. `qadence.ml_tools.to_dataloader` to build a dataloader."
                )

            iteration_callback()

        except KeyboardInterrupt:
            print("Terminating training gracefully after the current iteration.")
            break

    # Final writing and checkpointing
    final_callback()

    return model, optimizer
```
instead of whats currently being done with "hardcoded" functions.

smitchaudhary:
@DanieleCucurachi FYI!

chMoussa:
Closes with #533

PR #533, opened 2024-08-14
[Feature] Adding callback functionalities in train functions

chMoussa:
Related to a comment from #202, we can use callback functions in train functions to extend possibilities to do more during these functions and allow flexibility. Similar to scipy.optimize, we have an `OptimizeResult` class storing several optimization-related object/values.  The callback should take an `OptimizeResult` as input and apply a user-defined function.

Several booleans locate filter where the callbacks should be called within train. This also allow avoiding a few if statements for some already-defined callback-type functions (writing, checkpointing, printing, etc).

### DIFF
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -88,7 +88,7 @@ from qadence.ml_tools import TrainConfig, Callback
 batch_size = 5
 n_epochs = 100
 
-custom_callback = Callback(lambda opt_res: print(opt_res.model.parameters()) if opt_res.loss < 1.0e-03 else print('Loss threshold not yet reached.'), every=10, call_end_epoch=True)
+custom_callback = Callback(lambda opt_res: print(opt_res.model.parameters()), callback_condition=lambda opt_res: opt_res.loss < 1.0e-03, called_every=10, call_end_epoch=True)
 
 config = TrainConfig(
     folder="some_path/",
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -78,12 +78,17 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
 
 The [`TrainConfig`][qadence.ml_tools.config.TrainConfig] tells `train_with_grad` what batch_size should be used,
 how many epochs to train, in which intervals to print/log metrics and how often to store intermediate checkpoints.
+It is also possible to provide custom callback functions by instantiating a [`Callback`][qadence.ml_tools.config.Callback]
+with a function `callback` that only accept as argument an instance of [`OptimizeResult`][qadence.ml_tools.data.OptimizeResult] created within the `train` functions.
+One can also provide a `callback_condition` function, also only accepting an instance of [`OptimizeResult`][qadence.ml_tools.data.OptimizeResult], which returns True if `callback` should be called. If no `callback_condition` is provided, `callback` is called at every x epochs (specified by `Callback`'s `every` argument). We can also specify in which part of the training function the `Callback` will be applied.
 
 ```python exec="on" source="material-block"
-from qadence.ml_tools import TrainConfig
+from qadence.ml_tools import TrainConfig, Callback
 
 batch_size = 5
 n_epochs = 100
+
+custom_callback = Callback(lambda opt_res: print(opt_res.model.parameters()) if opt_res.loss < 1.0e-03 else print('Loss threshold not yet reached.'), every=10, call_end_epoch=True)
 
 config = TrainConfig(
     folder="some_path/",
@@ -91,6 +96,7 @@ config = TrainConfig(
     checkpoint_every=100,
     write_every=100,
     batch_size=batch_size,
+    callbacks = [custom_callback]
 )
 ```
 
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -80,7 +80,7 @@ The [`TrainConfig`][qadence.ml_tools.config.TrainConfig] tells `train_with_grad`
 how many epochs to train, in which intervals to print/log metrics and how often to store intermediate checkpoints.
 It is also possible to provide custom callback functions by instantiating a [`Callback`][qadence.ml_tools.config.Callback]
 with a function `callback` that only accept as argument an instance of [`OptimizeResult`][qadence.ml_tools.data.OptimizeResult] created within the `train` functions.
-One can also provide a `callback_condition` function, also only accepting an instance of [`OptimizeResult`][qadence.ml_tools.data.OptimizeResult], which returns True if `callback` should be called. If no `callback_condition` is provided, `callback` is called at every x epochs (specified by `Callback`'s `every` argument). We can also specify in which part of the training function the `Callback` will be applied.
+One can also provide a `callback_condition` function, also only accepting an instance of [`OptimizeResult`][qadence.ml_tools.data.OptimizeResult], which returns True if `callback` should be called. If no `callback_condition` is provided, `callback` is called at every x epochs (specified by `Callback`'s `called_every` argument). We can also specify in which part of the training function the `Callback` will be applied.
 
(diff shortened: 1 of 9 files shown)


qadence/ml_tools/config.py
```diff
@@ -28,6 +29,80 @@
...
+        self.call_before_opt = call_before_opt
+        self.call_end_epoch = call_end_epoch
+        self.call_after_opt = call_after_opt
+        self.call_during_eval = call_during_eval
+
+        self.every = every
+
+        if callback_condition is None:
+            if self.every > 1:
+                self.callback_condition = lambda opt_result: opt_result.iteration % self.every == 0
```
gvelikova, line 95: This will always be true for self.every==1, but I guess running it each time unnecessary would be costing extra.

gvelikova (approved):

qadence/ml_tools/config.py
```diff
@@ -28,6 +29,80 @@
...
+            else:
+                self.callback_condition = lambda opt_result: True
+        else:
+            self.callback_condition = callback_condition
+
+    def __call__(self, opt_result: OptimizeResult, *args: Any, **kwargs: Any) -> Any:
+        if self.callback_condition(opt_result):
+            return self.callback(opt_result, *args, **kwargs)
+
```
debrevitatevitae, line 104: As a general comment here, I would delegate the responsibility of triggering the callback condition to the callback condition itself. Here is a possible option
```python
@dataclass
class TrainingStep:
    """Information about which phase of the training we are in."""
    epoch_number: int | None = None
    before_opt: bool = True
    after_opt: bool = False
    during_eval: bool = False

CallbackFunction = Callable[[OptimizeResults], None]
CallbackCondition = Callable[[TrainingStep], bool]


class Callback:
    def __init__(self, callback_function: CallbackFunction, callback_condition: CallbackCondition):
        self._callback_function = callback_function
        self._callback_condition = callback_condition

    def __call__(self, training_step: TrainingStep, opt_result: OptimizeResult):
        if self.callback_condition(training_step):
            return self._callback_function(opt_result)
```

`TrainingStep` would be an internal 'clock' for the training phase, updated in `train_*` functions. As mentioned, we let `callback_condition` decide on whether the clock is right to trigger. In this way we could also remove `iteration` from `OptmizeResults`.

qadence/ml_tools/train_grad.py
```diff
@@ -160,115 +160,167 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
+    if not ((dataloader is None) or isinstance(dataloader, (DictDataLoader, DataLoader))):
+        raise NotImplementedError(
+            f"Unsupported dataloader type: {type(dataloader)}. "
+            "You can use e.g. `qadence.ml_tools.to_dataloader` to build a dataloader."
+        )
+
+    def next_loss_iter(dl_iter: Union[None, DataLoader, DictDataLoader]) -> Any:
+        xs = next(dl_iter) if dl_iter is not None else None
+        xs_to_device = data_to_device(xs, device=device, dtype=data_dtype)
+        return loss_fn(model, xs_to_device)
```
debrevitatevitae, line 172: Maybe include a docstring for explanation.

qadence/ml_tools/train_grad.py
```diff
@@ -160,115 +160,167 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
+    callbacks_before_opt_eval = [
+        callback for callback in callbacks if callback.call_before_opt and callback.call_during_eval
+    ]
+
     with progress:
         dl_iter = iter(dataloader) if dataloader is not None else None
 
         # Initial validation evaluation
         try:
+            opt_result = OptimizeResult(init_iter, model, optimizer, None, dict())
```
debrevitatevitae, line 272: can we avoid passing the last two arguments altogether here?

qadence/ml_tools/train_grad.py
```diff
@@ -160,115 +160,167 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
+                ),
+                every=config.checkpoint_every,
+                call_before_opt=True,
+                call_after_opt=True,
+                call_during_eval=True,
+            )
+        ]
+
+    def run_callbacks(callback_iterable: list[Callback], opt_res: OptimizeResult) -> None:
+        [callback(opt_res) for callback in callback_iterable]
```
debrevitatevitae, line 256: why the list comprehension?

debrevitatevitae (commented):
Cool work! Can we also add an example or a tutorial for the docs to showcase the callback definitions?

qadence/ml_tools/train_grad.py
chMoussa: Just liked one-liners but putting it back to a simple loop syntax

qadence/ml_tools/config.py
chMoussa: Hm, you can also trigger `callback_condition` based on the loss value or a condition coming from model or optimizer, so that is why I kept also the `callback_condition` to take as input the `OptimizeResult` instance. Although, I don't demonstrate it here directly. Also, the callback would be missing when it is called which would help us filter them before getting in the loop, and avoid unnecessary calling `callback_condition`.

chMoussa:
> Cool work! Can we also add an example or a tutorial for the docs to showcase the callback definitions?

Yes just did!

docs/tutorials/qml/ml_tools.md
```diff
@@ -78,19 +78,25 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
+custom_callback = Callback(lambda opt_res: print(opt_res.model.parameters()) if opt_res.loss < 1.0e-03 else print('Loss threshold not yet reached.'), every=10, call_end_epoch=True)
+
 config = TrainConfig(
     folder="some_path/",
     max_iter=n_epochs,
     checkpoint_every=100,
     write_every=100,
     batch_size=batch_size,
+    callbacks = [custom_callback]
 )
```
awennersteen, line 100: It would perhaps be nice to show the aforementioned `callback_condition`?

chMoussa: Sure, will change the example by making the if statement a `callback_condition` function!

qadence/ml_tools/train_grad.py
```diff
@@ -160,115 +160,178 @@ def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, d
...
-                        iteration,
-                        config.plotting_functions,
-                        tracking_tool=config.tracking_tool,
-                    )
                 if perform_val:
                     if iteration % config.val_every == 0:
-                        xs = next(dl_iter_val)
-                        xs_to_device = data_to_device(xs, device=device, dtype=data_dtype)
-                        val_loss, *_ = loss_fn(model, xs_to_device)
+                        val_loss, *_ = next_loss_iter(dl_iter_val)
```
awennersteen, line 328: we had some bug elsewhere where 0 is a valid value for these `val_every` type arguments, see #511 

IDK if this is guarded already, or if it's worth to add one

qadence/ml_tools/config.py
```diff
@@ -27,6 +28,80 @@
...
+    an OptimizeResult instance.
+
+    Attributes:
+        callback (CallbackFunction): Callback function accepting an
+            OptimizeResult as first argument.
+        callback_condition (CallbackConditionFunction | None, optional): Function that
+            conditions the call to callback. Defaults to None.
+        every (int, optional): Callback to be called each `every` epoch. Defaults to 1.
+            If callback_condition is None, we set
+            callback_condition to returns True when iteration % every == 0.
```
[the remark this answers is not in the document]
chMoussa: `called_every` sounds good to me

awennersteen (approved):
Some minor, optional, suggestions.

qadence/ml_tools/train_grad.py
chMoussa: I think we don't check. But I would raise an error in case it is 0 cause there is no point in defining a disabled callback.

PR #538, opened 2024-08-21
[Bug] Fix usage of the shift prefactor

chMoussa:
Reputting `n_eqs` in general PSR to limit the number of equations in case of multi-gap. Also, rescale the shift prefactor for cases of `Hamevo`.


awennersteen (approved):
lgtm

PR #554, opened 2024-09-02
[Feature] Add possibility to modify the OptimizeResult during training in ml_tools

chMoussa:
Simply adding an extra `modify_optimize_result` argument in `Callback` to change the `OptimizeResult` instance passed during training. A showcase is available in docs.

### DIFF
--- a/docs/tutorials/qml/ml_tools.md
+++ b/docs/tutorials/qml/ml_tools.md
@@ -80,15 +80,19 @@ The [`TrainConfig`][qadence.ml_tools.config.TrainConfig] tells `train_with_grad`
 how many epochs to train, in which intervals to print/log metrics and how often to store intermediate checkpoints.
 It is also possible to provide custom callback functions by instantiating a [`Callback`][qadence.ml_tools.config.Callback]
 with a function `callback` that only accept as argument an instance of [`OptimizeResult`][qadence.ml_tools.data.OptimizeResult] created within the `train` functions.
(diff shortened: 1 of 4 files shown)


docs/tutorials/qml/ml_tools.md
```diff
@@ -80,23 +80,27 @@ The [`TrainConfig`][qadence.ml_tools.config.TrainConfig] tells `train_with_grad`
...
 
 config = TrainConfig(
     folder="some_path/",
     max_iter=n_epochs,
     checkpoint_every=100,
     write_every=100,
     batch_size=batch_size,
-    callbacks = [custom_callback]
+    callbacks = [custom_callback, custom_callback_extra]
 )
```
DanieleCucurachi, line 104: Thanks for the quick work! The implementation works well. However, since the custom callback extra is still a `CallbackFunction`, it necessitates providing additional arguments as global variables. For instance, in the provided code snippet, if I wanted to use an arbitrary `external_argument`, I would need to define it as a global variable and add it to `extra` as done in `add_bias_extra`.

While this approach might be necessary if these additional arguments need to be modified multiple times during training, it would be beneficial to allow for some additional (fixed) arguments to be passed directly to `Callback`, which can then initialize certain entries of the `extra` dictionary with these arguments by default. @chMoussa what do you think about it?

DanieleCucurachi (commented):
Nice job, thanks @chMoussa , I left just a small suggestion.

docs/tutorials/qml/ml_tools.md
chMoussa: @DanieleCucurachi Ok, so maybe I should replace the modify_optimize_result to also be a dictionary that just update the extra field?

DanieleCucurachi: Sounds like a better solution to me, this way we can pass arbitrary arguments directly to `Callback`. The previous solution worked but, personally, I would avoid having to rely on global variables.

chMoussa: @DanieleCucurachi ok, just modified now to allow for a dict

PR #560, opened 2024-09-05
[BugFix] Fix writer callback

### LINKED ISSUE #559: [Bug] Only one data point in TB despite multiple logging iterations
Since the introduction of callbacks (#533), when using the `train` function, only a single data point is visible in the TensorBoard logs after training. This occurs regardless of the `write_every` parameter value, which is intended to control the frequency of intermediate result logging.

chMoussa:
Closes #559 . The writer callback would be only called during evaluation given the boolean values. So just need to separate the writer for evaluation and outside evaluation.

### DIFF
--- a/qadence/ml_tools/train_grad.py
+++ b/qadence/ml_tools/train_grad.py
@@ -221,10 +221,23 @@ def train(
                 writer,
                 opt_res.loss,
                 opt_res.metrics,
-                opt_res.iteration - 1,  # same reason as priting above
+                opt_res.iteration - 1,  # loss returned be optimized_step is at -1
                 tracking_tool=config.tracking_tool,
             ),
             called_every=config.write_every,
+            call_end_epoch=True,
+        ),
+        Callback(
+            lambda opt_res: write_tracker(
+                writer,
+                opt_res.loss,
+                opt_res.metrics,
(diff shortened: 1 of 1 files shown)


qadence/ml_tools/train_grad.py
```diff
@@ -332,12 +346,19 @@ def next_loss_iter(dl_iter: Union[None, DataLoader, DictDataLoader]) -> Any:
...
             loss, metrics, *_ = next_loss_iter(dl_iter)
             if iteration % config.print_every == 0 and config.verbose:
                 print_metrics(loss, metrics, iteration)
+            write_tracker(
+                writer,
+                loss,
+                metrics,
+                iteration,
+                tracking_tool=config.tracking_tool,
+            )
```
DanieleCucurachi, line 361: Whenever you add a new entry to metrics, the last value is never logged or printed as callbacks are not called here. I would add an additional step where (at least) the custom callbacks are called before `write_tracker`.

chMoussa: Ok, I changed that now to go back to having this as part of the final callbacks by specifying a callback `write_tracker` for after optimization, to handle the last loss and metrics.

DanieleCucurachi (approved):
It works nicely now, thank you @chMoussa

Issue #532, opened 2024-08-12
Implement constructor function for alternating layer ansatz


ducthanh1991:
Hello,

Have you considered implementing a function to construct the alternating layer ansatz defined in the papers of [Nakaji et al.]( https://quantum-journal.org/papers/q-2021-04-19-434/) and [Cerezo et al.](https://www.nature.com/articles/s41467-021-21728-w) that would be similar to the existing `hea` function of the module `qadence.constructors.ansatze`?

chMoussa:
Hi @ducthanh1991, we have not considered until you asked. It is probably a matter of changing slightly changing the function yielding the entangler, right?

ducthanh1991:
Hello @chMoussa, 

Yes, I guess that one would need to specify all the same parameters as for the `hea` function plus one additional parameter for the size of the support of the "local" block of the alternating layer ansatz.

something like:

```
ansatz = alt_digital(
        depth=`4,
        n_qubits=12,
        m_qubits_block=4,
        operations=[RX],
        entangler=CNOT,
        )
display(ansatz,size="15,15")
```
would yield:
![image](https://github.com/user-attachments/assets/feaa2755-34bf-499b-a729-19641e03a6c7)

chMoussa:
> Hello @chMoussa,
> 
> Yes, I guess that one would need to specify all the same parameters as for the `hea` function plus one additional parameter for the size of the support of the "local" block in the alternating layer ansatz.
> 
> something like:
> 
> ```
> ansatz = alt_digital(
>         depth=`4,
>         n_qubits=12,
>         m_qubits_block=4,
>         operations=[RX],
>         entangler=CNOT,
>         )
> display(ansatz,size="15,15")
> ```
> 
[...]

PR #565, opened 2024-09-08
[Feature] Implement constructor function for alternating layer ansatz

ducthanh1991:
Proposition of resolution of #532: Implementation of a constructors function that create a Alternating Layer Ansatz circuit.

Done:
- Implemented the digital strategy to create an ALT circuit

image
Example of an alt circuit generated with the function for `n_qubits=10,m_qubits_block=4,depth=2`

To do:
- implement the sdaqc strategy
- implementation of the bdaqc strategy
- implementation of the analog strategy

### DIFF
--- a/qadence/constructors/ansatze.py
+++ b/qadence/constructors/ansatze.py
@@ -411,7 +411,7 @@ def alt_digital(
 
         
 )->AbstractBlock:
-    print(depth)
+    
     rot_list = _rotations_digital(
         n_qubits=n_qubits,
         
--- a/qadence/constructors/ansatze.py
+++ b/qadence/constructors/ansatze.py
@@ -349,7 +349,12 @@ def _entanglers_block_digital(
         if not d%2:
            ents.append(
                kron(
-                   _entangler(control=support[i+j],target=support[i+j+1],param_str=param_prefix + f"_ent_{next(iterator)}",op=entangler)
+                   _entangler(
+                       control=support[i+j],
+                       target=support[i+j+1],
+                       param_str=param_prefix + f"_ent_{next(iterator)}",
+                       op=entangler
+                       )
                    
                    for i in range(0,n_qubits,m_block_qubits)
                    for j in range(0,m_block_qubits,2)  
@@ -360,7 +365,12 @@ def _entanglers_block_digital(
            if m_block_qubits > 2:
                ents.append(
                 kron(
-                    _entangler(control=support[i+j],target=support[i+j+1],param_str=param_prefix + f"_ent_{next(iterator)}",op=entangler)
+                    _entangler(
+                        control=support[i+j],
+                        target=support[i+j+1],
+                        param_str=param_prefix + f"_ent_{next(iterator)}",
+                        op=entangler
+                        )
                     
                     for i in range(0,n_qubits,m_block_qubits) 
                     for j in range(1,m_block_qubits,2) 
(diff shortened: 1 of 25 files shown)


awennersteen:
Thanks a lot!

We might take a couple of days to review properly due to slightly busy period, and some holidays being taken on top of that

qadence/constructors/ansatze.py
```diff
@@ -320,3 +320,222 @@ def hea_bDAQC(*args: Any, **kwargs: Any) -> Any:
...
+                            param_str=param_prefix + f"_ent_{next(iterator)}",
+                            op=entangler,
+                        )
+                        for i in range(0, n_qubits, m_block_qubits)
+                        for j in range(1, m_block_qubits, 2)
+                        if i + j + 1 < n_qubits and j + 1 < m_block_qubits
+                    )
+                )
+
+        elif d % 2:
```
chMoussa, line 432: Why elif and not just do a else?

qadence/constructors/ansatze.py
```diff
@@ -320,3 +320,222 @@ def hea_bDAQC(*args: Any, **kwargs: Any) -> Any:
...
+                        target=support[i + j + 1],
+                        param_str=param_prefix + f"_ent_{next(iterator)}",
+                        op=entangler,
+                    )
+                    for i in range(-m_block_qubits // 2, n_qubits, m_block_qubits)
+                    for j in range(0, m_block_qubits, 2)
+                    if i + j + 1 < n_qubits and j + 1 < m_block_qubits and i + j >= 0
+                )
+            )
+            if m_block_qubits > 2:
```
chMoussa, line 446: Given the repeating blocks of code, I would suggest a syntax change by introducing at the beginning of the loop:
- `start_i, start_j` variables to be set depending on the `d %2` case and `m_block_qubits`
- entangler_indices = [(i,j) for i in range(start_i, n_qubits, m_block_qubits) for j in range(start_j, m_block_qubits, 2) if ...] and adjust the if conditions accordingly
- then you will only write the `kron(_entangler(...) for i,j in entangler_indices)` once!

chMoussa (changes requested):

qadence/constructors/ansatze.py
ducthanh1991: Will be updated in the next commit

qadence/constructors/ansatze.py
```diff
@@ -320,3 +320,222 @@ def hea_bDAQC(*args: Any, **kwargs: Any) -> Any:
...
+    support: tuple[int, ...] = None,
+    strategy: Strategy = Strategy.DIGITAL,
+    **strategy_args: Any,
+) -> AbstractBlock:
+    """
+    Factory function for the Alternating Layer Ansatz (ALT).
+
+    Args:
+        n_qubits: number of qubits in the block
+        m_block_qubits: number of qubits in the local entangling block
```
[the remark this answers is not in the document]
ducthanh1991: Indeed the paper of Nakaji et al. makes the assumption of `n/m` integer. But it does not look like it's vital for the construction of the ALT ansatze. We should compare the expressibility and trainability of the cases with and without `n/m` being an integer.

qadence/constructors/ansatze.py
ducthanh1991: see https://github.com/pasqal-io/qadence/pull/565#discussion_r1767045762

chMoussa (approved):
Nice! LGTM, thanks @ducthanh1991

PR #568, opened 2024-09-12
[Refact] Get rid of PyQTimeDependentEvolution

vytautas-a:
Substitute `PyQTimeDependentEvolution` class with proper call to `pyqtorch`'s time-dependent functionality.

### DIFF
--- a/qadence/backends/pyqtorch/convert_ops.py
+++ b/qadence/backends/pyqtorch/convert_ops.py
@@ -1,37 +1,22 @@
 from __future__ import annotations
 
+from functools import partial, reduce
 from itertools import chain as flatten
-from math import prod
-from typing import Any, Sequence, Tuple
+from typing import Callable, Sequence
 
 import pyqtorch as pyq
 import sympy
 import torch
-from pyqtorch.embed import Embedding
-from pyqtorch.matrices import _dagger
-from pyqtorch.time_dependent.sesolve import sesolve
-from pyqtorch.utils import is_diag
+from pyqtorch.embed import ConcretizedCallable
 from torch import (
     Tensor,
     cdouble,
     complex64,
-    diag_embed,
-    diagonal,
-    exp,
     float64,
-    linalg,
     tensor,
-    transpose,
 )
-from torch import device as torch_device
-from torch import dtype as torch_dtype
-from torch.nn import Module, ParameterDict
+from torch.nn import Module
 
-from qadence.backends.utils import (
-    finitediff,
-    pyqify,
-    unpyqify,
-)
 from qadence.blocks import (
     AbstractBlock,
     AddBlock,
@@ -43,11 +28,7 @@ from qadence.blocks import (
     ScaleBlock,
     TimeEvolutionBlock,
 )
-from qadence.blocks.block_to_tensor import (
-    _block_to_tensor_embedded,
-)
 from qadence.blocks.primitive import ProjectorBlock
-from qadence.blocks.utils import parameters
 from qadence.operations import (
     U,
     multi_qubit_gateset,
@@ -59,6 +40,18 @@ from qadence.operations import (
 from qadence.types import OpName
 
 from .config import Configuration
+
+SYMPY_TO_PYQ_MAPPING = {
+    sympy.Pow: "pow",
+    sympy.cos: "cos",
+    sympy.Add: "add",
+    sympy.Mul: "mul",
+    sympy.sin: "sin",
+    sympy.log: "log",
+    sympy.tan: "tan",
+    sympy.tanh: "tanh",
+}
+
 
 # Tdagger is not supported currently
 supported_gates = list(set(OpName.list()) - set([OpName.TDAGGER]))
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
     return config.get_param_name(block)[0]
 
 
+def sympy_to_pyq(expr: sympy.Expr) -> ConcretizedCallable:
+    """Convert sympy expression to pyqtorch ConcretizedCallable object.
+
+    Args:
+        expr (sympy.Expr): sympy expression
+
+    Returns:
+        ConcretizedCallable: expression encoded as ConcretizedCallable
+    """
+
+    # base case - independent argument
+    if len(expr.args) == 0:
+        res = (
+            float(expr)
+            if str(expr).replace(".", "", 1).replace("-", "", 1).isdigit()
+            else str(expr)
+        )
+        if isinstance(res, str):
+            if "/" in res:
+                # found a rational
+                res = float(sympy.Rational(res).evalf())
+            if "fix" in res:  # type: ignore [operator]
+                # found a fixed parameter - convert to float
+                res = torch.as_tensor(float(res.split("_")[1]))  # type: ignore [attr-defined]
+        return res
+
+    # iterate through current function arguments
+    all_results = []
+    for arg in expr.args:
+        res = sympy_to_pyq(arg)
+        all_results.append(res)
+
+    # deal with multi-argument (>2) sympy functions: converting to nested
+    # ConcretizedCallable objects
+    if len(all_results) > 2:
+
+        def fn(x: str | ConcretizedCallable, y: str | ConcretizedCallable) -> Callable:
+            return partial(ConcretizedCallable, call_name=SYMPY_TO_PYQ_MAPPING[expr.func])(  # type: ignore [no-any-return]
+                abstract_args=[x, y]
+            )
+
+        cc = reduce(fn, all_results)
+    else:
+        cc = ConcretizedCallable(SYMPY_TO_PYQ_MAPPING[expr.func], all_results)
+    return cc
+
+
 def convert_block(
     block: AbstractBlock, n_qubits: int = None, config: Configuration = None
 ) -> Sequence[Module | Tensor | str | sympy.Expr]:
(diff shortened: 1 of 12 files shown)


vytautas-a:
This refactoring disables the embedding function call for `pyqtorch` backend, thus currently only `HamEvo` block works properly. I suggest we could completely get rid of the embedding call when dealing with `pyqtorch`. This can be achieved by utilizing the new `sympy` -> `ConcretizedCallable` converter.

Roland-djee:
Hey @vytautas-a can I mark this PR as ready ?

vytautas-a:
@Roland-djee, well this PR here is currently more in the proposal stage. It depends also on another PR that I'll open in `pyqtorch`.

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
...
+    # base case - independent argument
+    if len(expr.args) == 0:
+        res = (
+            float(expr)
+            if str(expr).replace(".", "", 1).replace("-", "", 1).isdigit()
+            else str(expr)
+        )
+        if isinstance(res, str):
+            if "/" in res:
+                # found a rational
```
Roland-djee, line 113:
```suggestion
            if "/" in res:  # Found a rational
```

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
...
+    Returns:
+        ConcretizedCallable: expression encoded as ConcretizedCallable
+    """
+
+    # base case - independent argument
+    if len(expr.args) == 0:
+        res = (
+            float(expr)
+            if str(expr).replace(".", "", 1).replace("-", "", 1).isdigit()
+            else str(expr)
```
Roland-djee, line 109: Would it make sense not to convert to a `str` here and use Sympy functionalities later ? Ie. when checking for rationality ?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
...
+        if isinstance(res, str):
+            if "/" in res:
+                # found a rational
+                res = float(sympy.Rational(res).evalf())
+            if "fix" in res:  # type: ignore [operator]
+                # found a fixed parameter - convert to float
+                res = torch.as_tensor(float(res.split("_")[1]))  # type: ignore [attr-defined]
+        return res
+
+    # iterate through current function arguments
```
Roland-djee, line 120:
```suggestion
    # Recursively iterate through current function arguments
```

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
...
+    all_results = []
+    for arg in expr.args:
+        res = sympy_to_pyq(arg)
+        all_results.append(res)
+
+    # deal with multi-argument (>2) sympy functions: converting to nested
+    # ConcretizedCallable objects
+    if len(all_results) > 2:
+
+        def fn(x: str | ConcretizedCallable, y: str | ConcretizedCallable) -> Callable:
```
Roland-djee, line 130:
```suggestion
        def convert(x: str | ConcretizedCallable, y: str | ConcretizedCallable) -> Callable:
```

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
     return config.get_param_name(block)[0]
 
 
+def sympy_to_pyq(expr: sympy.Expr) -> ConcretizedCallable:
+    """Convert sympy expression to pyqtorch ConcretizedCallable object.
```
Roland-djee, line 95:
```suggestion
    """Convert sympy expression to a possibly nested pyqtorch ConcretizedCallable object.
```

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +91,53 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
...
+    # deal with multi-argument (>2) sympy functions: converting to nested
+    # ConcretizedCallable objects
+    if len(all_results) > 2:
+
+        def fn(x: str | ConcretizedCallable, y: str | ConcretizedCallable) -> Callable:
+            return partial(ConcretizedCallable, call_name=SYMPY_TO_PYQ_MAPPING[expr.func])(  # type: ignore [no-any-return]
+                abstract_args=[x, y]
+            )
+
+        cc = reduce(fn, all_results)
```
Roland-djee, line 135:
```suggestion
        concretized_callables = reduce(fn, all_results)
```

qadence/execution.py
```diff
@@ -76,10 +76,14 @@ def _(
         diff_mode = DiffMode.AD
     bknd = backend_factory(backend, diff_mode=diff_mode, configuration=configuration)
     conv = bknd.convert(circuit)
+    if backend == BackendName.PYQTORCH:
+        vals = values
```
Roland-djee, line 80:
```suggestion
        embedded_values = values
```

Roland-djee (commented):
Thanks a lot @vytautas-a looks nice and clean. Minor renaming suggestions and comments. Will approve once addressed.

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -114,36 +157,47 @@ def convert_block(
 
     if isinstance(block, ScaleBlock):
         scaled_ops = convert_block(block.block, n_qubits, config)
-        scale = extract_parameter(block, config)
-        return [pyq.Scale(pyq.Sequence(scaled_ops), scale)]
+        scale = extract_parameter(block, config=config)
+
+        # replace underscore by dot when underscore is between two numbers in string
+        if isinstance(scale, str):
+            scale = re.sub(r"(?<=\d)_(?=\d)", ".", scale)
```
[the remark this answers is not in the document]
vytautas-a: This substitution is required due to the method how `qadence` stringifies `sympy` expressions. Your corner case is interesting - I have to test it, but it would seem that current implementation would substitute a dot in this variable. And that is not desirable. The reason for this regular expression is to convert floats from form `0_1234` to `0.1234` that is understood by `parse_expr` method of `sympy`. But yes, it has to be updated to handle your case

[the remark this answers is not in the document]
vytautas-a: fixed

tests/engines/test_torch.py
```diff
@@ -100,7 +100,7 @@ def test_embeddings() -> None:
         "orig_param_values", {}
     )  # TODO: remove this when embedding system is updated
 
-    assert len(list(low_level_params.keys())) == 9
+    assert len(list(low_level_params.keys())) == 9 + len(inputs)
```
jpmoutinho, line 103: Are you sure this is the right fix?

The line above has a comment to be removed "when embedding system is updated". Could it be that removing that will make the `assert` pass without the extra manual addition?

tests/qadence/test_quantum_model.py
```diff
@@ -52,7 +52,9 @@ def test_quantum_model_parameters(parametric_circuit: QuantumCircuit) -> None:
     embedded_params_ad.pop(
         "orig_param_values", {}
     )  # TODO: remove this when embedding system is updated
-    assert len(embedded_params_ad) == 5
+    assert (
+        len(embedded_params_ad) == 5 + 1
+    )  # adding one because original param x is included for PYQ + AD
```
jpmoutinho, line 57: Same as above

tests/qadence/test_quantum_model.py
```diff
@@ -71,7 +73,7 @@ def test_quantum_model_duplicate_expr(duplicate_expression_circuit: QuantumCircu
     embedded_params_ad.pop(
         "orig_param_values", {}
     )  # TODO: remove this when embedding system is updated
-    assert len(embedded_params_ad) == 2
+    assert len(embedded_params_ad) == 2 + 4  # adding 4 because all original params are included
```
jpmoutinho, line 76: Same as above

jpmoutinho (commented):
Thanks @vytautas-a :)

Few minor things and then it should be good to go.

One thing, I see that the docs for the time-dependent HamEvo reads:

```
# create parameterized HamEvo block
hamevo = HamEvo(generator_td, 0.0, duration=duration)
```

Is passing `0.0` correct there? I thought the idea was to pass the time parameter in that argument?

tests/engines/test_torch.py
vytautas-a: good catch, I'll update the test

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -98,6 +98,71 @@ def extract_parameter(block: ScaleBlock | ParametricBlock, config: Configuration
...
+    """
+
+    # Function to replace the underscore with a dot
+    def underscore_to_dot(match: re.Match) -> Any:
+        return match.group(0).replace("_", ".")
+
+    # Compile the regular expression
+    pattern = re.compile(float_with_underscore_pattern, re.VERBOSE)
+
+    return pattern.sub(underscore_to_dot, s)
```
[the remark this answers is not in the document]
vytautas-a: Ok, I'll add it.

vytautas-a:
> Thanks @vytautas-a :)
> 
> Few minor things and then it should be good to go.
> 
> One thing, I see that the docs for the time-dependent HamEvo reads:
> 
> ```
> # create parameterized HamEvo block
> hamevo = HamEvo(generator_td, 0.0, duration=duration)
> ```
> 
> Is passing `0.0` correct there? I thought the idea was to pass the time parameter in that argument?

Yes, the docs have to updated.

tests/engines/test_torch.py
vytautas-a: It turns out that I already removed the `orig_param_values` key, thus the test is actually correct.

vytautas-a: I just removed the unnecessary code from these tests.

qadence/backends/pyqtorch/convert_ops.py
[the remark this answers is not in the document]
vytautas-a: done

jpmoutinho (approved):
LGTM!

Issue #583, opened 2024-10-14
[Refactoring] [Testing] Noise 


chMoussa:
Folowing the digital noise MR #563, a few extra tasks are to be carried out:

- Properly making consistent the DigitalNoise with the Noise class and etc... Some are backend independent (Readout), others are analog/digital-specific. It makes sense to separate the different types.
- Including the noise in each block string: if added, check how it behaves with visualization.
- Serialization: make `_to_dict` and `_from_dict` functions to noise class
- Add tests! For transpiler functions such as `set_noise`

chMoussa:
At the same time, tackle #369

mlahariya:
@chMoussa closing this as most things are already implemented. Visualization which is left will be added to mega visualization issue. #613

PR #591, opened 2024-10-25
[Refactoring] [Testing] [Feature] NoiseHandler

chMoussa:
Supersedes #584  and solves #583 (besides plotting). We introduce a `NoiseHandler` class containing information about the noise. For digital noise, this means now that we can do a conversion in `convert_ops` to the right class in Pyq. @jpmoutinho also proposed to separate noise protocol types into Analog, Digital, and Readout (given that a same protocol name may be available in different settings such as depolarizing).

Note that a substantial harmonisation of naming may be needed between different backend (Readout is basically called SPAM in Pulser).

Todos:

- [x] Redesign the previous `Noise` class into the `NoiseHandler` class to allow multiple combination of noise protocols.
- [x] Change typings and replace `DigitalNoise` in blocks.
- [x] Add conversion function for Pyq backend.
- [x] Serialize noise in blocks
- [x] Add Tests
- [x] Elaborate the docs.


qadence/mitigations/readout.py
```diff
@@ -88,7 +88,12 @@ def mitigation_minimization(
     Returns:
         Mitigated counts computed by the algorithm
     """
-    noise_matrices = noise.options.get("noise_matrix", noise.options["confusion_matrices"])
+    readoutnoise = noise if isinstance(noise, NoiseSource) else noise.noise_sources[-1]
```
RolandMacDoland, line 91:
```suggestion
    readout_noise = noise if isinstance(noise, NoiseSource) else noise.noise_sources[-1]
```

qadence/mitigations/readout.py
```diff
@@ -88,7 +88,12 @@ def mitigation_minimization(
...
+    if readoutnoise.protocol != "Readout":
```
RolandMacDoland, line 92: Can't you compare to a `NoiseType` enum instead of a hardcoded `str` ?

qadence/noise/protocols.py
```diff
@@ -1,29 +1,62 @@
...
-    DEPHASING = "dephasing"
-    DEPOLARIZING = "depolarizing"
-    READOUT = "readout"
+class NoiseSource:
+    """A container for a single source of noise."""
+
+    def __init__(self, protocol: str, options: dict = dict(), protocol_type: str = "") -> None:
+        if protocol not in supported_noise_protocols:
+            raise ValueError(
+                "Protocol {protocol} is not supported. Choose from {supported_noise_protocols}."
```
RolandMacDoland, line 23:
```suggestion
                f"Protocol {protocol} is not supported. Choose from {supported_noise_protocols}."
```

qadence/noise/protocols.py
```diff
@@ -33,27 +66,161 @@ def get_noise_fn(self) -> Callable:
...
+        return {
+            "protocol": self.protocol,
+            "options": self.options,
+            "protocol_type": self.protocol_type,
+        }
 
     @classmethod
-    def _from_dict(cls, d: dict) -> Noise | None:
+    def _from_dict(cls, d: dict | None) -> NoiseSource | None:
         if d:
```
RolandMacDoland, line 77:
```suggestion
        if d is not None:
```

qadence/noise/protocols.py
```diff
@@ -33,27 +66,161 @@ def get_noise_fn(self) -> Callable:
...
+    def __repr__(self) -> str:
+        return f"{self.protocol_type} NoiseSource({self.protocol}, {str(self.options)})"
+
+
+class NoiseHandler:
+    """A container for multiple sources of noise."""
+
+    def __init__(
+        self,
+        protocol: str | NoiseSource | list[str] | list[NoiseSource],
```
RolandMacDoland, line 95: Do we still need `str` ?

qadence/noise/protocols.py
```diff
@@ -33,27 +66,161 @@ def get_noise_fn(self) -> Callable:
...
+        else:
+            protocol = protocol if isinstance(protocol, list) else [protocol]
+            options = options if isinstance(options, list) else [options] * len(protocol)
+            types = (
+                protocol_type
+                if isinstance(protocol_type, list)
+                else [protocol_type] * len(protocol)
+            )
+
+            if len(options) != len(protocol) or len(types) != len(protocol):
```
RolandMacDoland, line 113: There might be noises that do not require options ? In that case, you'll fall in the first of you predicate cases without necessarily being wrong ?

qadence/mitigations/readout.py
RolandMacDoland: For instance `NoiseProtocolType` ?

qadence/types.py
```diff
@@ -459,3 +459,28 @@ class ExperimentTrackingTool(StrEnum):
 
 
 LoggablePlotFunction = Callable[[Module, int], tuple[str, Figure]]
+
+
+class NoiseProtocolType(StrEnum):
```
RolandMacDoland, line 464: Would it make sense to merge this two enums somehow ? So that you could do smtg like `NoiseType.ANALOG.DEPOLARIZING` ?

tests/qadence/test_noise/test_digital_noise.py
```diff
@@ -0,0 +1,71 @@
...
+    hamiltonian_factory,
+    kron,
+    set_noise,
+)
+
+list_noises = [DigitalNoiseType(noise.value) for noise in DigitalNoiseType]
+
+
+@pytest.mark.parametrize("protocol", list_noises)
+@given(st.digital_circuits())
```
RolandMacDoland, line 26: Nice :)

tests/qadence/test_noise/test_digital_noise.py
```diff
@@ -0,0 +1,71 @@
...
+    NoiseHandler,
+    QuantumCircuit,
+    QuantumModel,
+    Z,
+    hamiltonian_factory,
+    kron,
+    set_noise,
+)
+
+list_noises = [DigitalNoiseType(noise.value) for noise in DigitalNoiseType]
```
RolandMacDoland, line 22: Since it is reused in more than one test, maybe it could become a fixture ?

RolandMacDoland (commented):
Thanks a lot for this @chMoussa ! My own impression is that there is a profusion of `StrEnum` that maybe we should think about simplifiying, `eg.` `NoiseType`, `NoiseProtocoltype` etc. It'll rapidly get out of hand if we don't try to reduce the number of them. Otherwise it looks good. I would also try out some deterministic test with noise just to make sure it gives correct values back. I'll also seek for @jpmoutinho and @mlahariya feedback.

qadence/types.py
chMoussa: Made a dataclass `NoiseProtocol` with different levels of noises.

qadence/noise/protocols.py
chMoussa: I am forcing noise without options to be an empty dict

qadence/noise/protocols.py
chMoussa: Not anymore with the new definition of NoiseProtocol

tests/qadence/test_noise/test_digital_noise.py
chMoussa: So far, better to keep the parameterize on the list instead of doing a for loop in the test.

docs/tutorials/realistic_sims/noise.md
```diff
@@ -1,7 +1,37 @@
...
+One can also define a `NoiseHandler` as a list of `NoiseSource` instances:
+
+```python exec="on" source="material-block" session="noise" result="json"
+from qadence import NoiseHandler, NoiseSource
+from qadence.types import NoiseProtocol
+
+digital_noise = NoiseSource(protocol=NoiseProtocol.DIGITAL.DEPOLARIZING, options={"error_probability": 0.1})
+readout_noise = NoiseSource(protocol=NoiseProtocol.READOUT.READOUT, options={"error_probability": 0.1, "seed": 0})
+
+noise_combination = NoiseHandler([digital_noise, readout_noise])
```
RolandMacDoland, line 28: If the `NoiseHandler` can accept a single protocol (as shown in line 61 below) why would it need a list of `NoiseSource` ? It feels a bit on an unnecessary intermediate step unless there is a reason I don't see now.

docs/tutorials/realistic_sims/noise.md
```diff
@@ -69,16 +98,45 @@ print(f"noiseless = {noiseless_exp}") # markdown-exec: hide
...
 
 ```python exec="on" source="material-block" session="noise" result="json"
-from qadence import NoiseType, DigitalNoise, RX, run
+from qadence import NoiseProtocol, RX, run
 import torch
 
-noise = DigitalNoise(NoiseType.BITFLIP, error_probability = 0.2)
-noise = DigitalNoise.bitflip(error_probability = 0.2) # equivalent
+noise = NoiseHandler(NoiseProtocol.DIGITAL.BITFLIP, {"error_probability": 0.2})
+noise = NoiseHandler.bitflip({"error_probability": 0.2}) # equivalent
```
RolandMacDoland, line 139: Would it make sense to build on this to chain noise models ? Smtg like:

```python
noise = NoiseHandler.bitflip(...).phaseflip(...).<other-noise-model>(...)
```

That keep adding in the list.

pyproject.toml
```diff
@@ -26,7 +26,7 @@ authors = [
 ]
 requires-python = ">=3.9"
 license = { text = "Apache 2.0" }
-version = "1.7.8"
+version = "1.7.9"
```
RolandMacDoland, line 29: I think you can bump the minor.

qadence/backends/pulser/backend.py
```diff
@@ -235,27 +235,27 @@ def run(
...
-        if noise_probs is None:
-            KeyError("A `noise probs` option should be passed to the <class QuantumModel>.")
-        if not (isinstance(noise_probs, float) or isinstance(noise_probs, Iterable)):
-            KeyError(
-                "A single or a range of noise probabilities"
-                " should be passed. Got {type(noise_probs)}."
-            )
+        noise_source = noise.noise_sources[-1]
+        if not isinstance(noise_source.protocol, NoiseProtocol.ANALOG):
+            raise TypeError("Noise must be Analog.")
```
RolandMacDoland, line 246:
```suggestion
            raise TypeError("Noise must be of type `NoiseProtocol.ANALOG`.")
```

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -286,17 +300,35 @@ def convert_block(
...
     else:
         raise NotImplementedError(
             f"Non supported operation of type {type(block)}. "
             "In case you are trying to run an `AnalogBlock`, make sure you "
             "specify the `device_specs` in your `Register` first."
         )
+
+
+def convert_digital_noise(noise: NoiseHandler | None) -> pyq.noise.NoiseProtocol | None:
+    if noise is None:
```
RolandMacDoland, line 324: Maybe do this check outside this function ? So that you don't have to do it every time it is called and you won't have to handle the `None` return type ?

qadence/noise/protocols.py
```diff
@@ -1,29 +1,37 @@
...
 
 PROTOCOL_TO_MODULE = {
-    "readout": "qadence.noise.readout",
+    "Readout": "qadence.noise.readout",
 }
 
-# Temporary solution
-DigitalNoise = NoiseProtocol
 
+class NoiseSource:
```
RolandMacDoland, line 13: I have the feeling all the checking `NoiseSource` provides can be done in the handler. Is that true ?

qadence/types.py
```diff
@@ -459,3 +461,29 @@ class ExperimentTrackingTool(StrEnum):
...
+
+
+class AnalogNoise(StrEnum):
+    """Type of noise protocol."""
+
+    DEPOLARIZING = "Depolarizing"
+    DEPHASING = "Dephasing"
+
+
+class ReadoutNoise(StrEnum):
```
RolandMacDoland, line 473: Yeah unsure we need to maintain an `StrEnum` for a single variant.

RolandMacDoland (commented):
Thanks @chMoussa more comments from me as I _think_ we can simplify further. Another point is about testing for composed noise models. Have you tried it out ?

docs/tutorials/realistic_sims/noise.md
chMoussa: This is especially for the case of Digital noise where we can combine several types (if the user decide to apply let's say a bitflip followed by a phase flip, etc)

docs/tutorials/realistic_sims/noise.md
chMoussa: Originally I intented for these methods to instantiate easily a handler with one type of noise. Maybe I could add a `append` method to add to the list?

qadence/noise/protocols.py
chMoussa: I wanted to separate the checking of options at the NoiseSource level, and for the handler to check more the combination of several types.

qadence/types.py
chMoussa: At the moment no, but there may be other readout types implemented at a later stage.

docs/tutorials/realistic_sims/mitigation.md
```diff
@@ -45,7 +45,7 @@ observable = hamiltonian_factory(circuit.n_qubits, detuning=Z)
 model = QuantumModel(circuit=circuit, observable=observable)
 
 # Define a noise model to use:
-noise = Noise(protocol=Noise.READOUT)
+noise = NoiseHandler(protocol=NoiseProtocol.READOUT.READOUT)
```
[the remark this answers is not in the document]
chMoussa: Made it work by setting `NoiseProtocol.READOUT` to be just a string for now.

qadence/types.py
chMoussa: But I will replace it for now and do this change when we introduce another readout protocol

chMoussa:
> Thanks @chMoussa more comments from me as I _think_ we can simplify further. Another point is about testing for composed noise models. Have you tried it out ?

Yes, I just added a parameterized test for this in `qadence/tests/qadence/test_noise/test_digital_noise.py`

docs/tutorials/realistic_sims/noise.md
```diff
@@ -1,7 +1,37 @@
...
+noise_combination = NoiseHandler([digital_noise, readout_noise])
+print(noise_combination)
+```
+
+!!! warning "NoiseHandler scope"
+    Note it is not possible to define a `NoiseHandler` instances with both digital and analog noise sources, both readout and analog noise sources, several analog noise sources, several readout noise sources, or a readout noise source that is not the last defined `NoiseSource` within `NoiseHandler`.
+
 ## Readout errors
 
 State Preparation and Measurement (SPAM) in the hardware is a major source of noise in the execution of
```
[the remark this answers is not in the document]
chMoussa: NoiseHandler.bitflip would only create an instance with only one noisesource.

qadence/mitigations/analog_zne.py
```diff
@@ -116,11 +116,12 @@ def noise_level_experiment(
     circuit: QuantumCircuit,
     observable: list[AbstractBlock],
     param_values: dict[str, Tensor],
-    noise: Noise,
+    noise: NoiseHandler,
     endianness: Endianness,
     state: Tensor | None = None,
 ) -> Tensor:
-    noise_probs = noise.options.get("noise_probs")
+    noise_source = noise.noise_sources[-1]
```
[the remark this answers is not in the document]
chMoussa: You cannot create a NoiseHandler with empty list. You get a raised error when instantating a NoiseHandler. And I made the convention that, if a readout is specified, it has to be the last noisesource (also raises an error at initialization).

docs/tutorials/realistic_sims/noise.md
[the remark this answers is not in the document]
chMoussa: And just added

docs/tutorials/realistic_sims/noise.md
chMoussa: Added the `append`

jpmoutinho (approved):
Very nice @chMoussa, this looks much cleaner :)

I did not go into the full details of the code, let me know if there is something specific that you would like to discuss. But in general, this interface has my approval.

qadence/backends/pulser/backend.py
```diff
@@ -24,12 +24,12 @@
 from qadence.measurements import Measurements
 from qadence.mitigations import Mitigations
 from qadence.mitigations.protocols import apply_mitigation
-from qadence.noise import Noise
-from qadence.noise.protocols import apply_noise
+from qadence.noise import NoiseHandler
+from qadence.noise.protocols import apply_readout_noise
```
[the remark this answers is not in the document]
chMoussa: Yes

chMoussa:
> Thanks @chMoussa. Sorry for being a pain but I think we still have too many types that makes the whole implementation difficult to follow. Ideally, I _think_ we can get away with max two types. Also, would it make sense to turn the `NoiseHandler` into a pydantic model for all the checking ? It might also help in serialization. Just a suggestion.

@RolandMacDoland, I followed your suggestions. Now we only have the `NoiseHandler` and the classmethods are becoming mutation methods. Let me know what you think of it

qadence/backends/pulser/backend.py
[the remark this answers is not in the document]
chMoussa: And in a future MR following a Pyqtorch release, we will have an interface ReadoutNoise that should instantiate the corresponding backend object.

qadence/noise/protocols.py
```diff
@@ -33,27 +52,231 @@ def get_noise_fn(self) -> Callable:
...
+    def _from_dict(cls, d: dict | None) -> NoiseHandler | None:
+        if d is not None and d.get("protocol", None):
+            return cls(d["protocol"], d["options"])
         return None
 
     @classmethod
     def list(cls) -> list:
         return list(filter(lambda el: not el.startswith("__"), dir(cls)))
 
+    def filter(self, protocol_type: NoiseEnum) -> NoiseHandler | None:
```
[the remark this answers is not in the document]
chMoussa: Changed to `protocol`

qadence/noise/protocols.py
chMoussa: I get it, now it's removed.

docs/tutorials/realistic_sims/noise.md
chMoussa: Gotcha, now it's been reworked.

PR #592, opened 2024-10-29
Add cloud integration features to qadence

pimvenderbosch:
This PR aims to add features to qadence that enable the user to upload their workloads to Pasqal's cloud platform.

### DIFF
--- a/docs/tutorials/advanced_tutorials/cloud-submission.md
+++ b/docs/tutorials/advanced_tutorials/cloud-submission.md
@@ -1,0 +1,120 @@
+# Submission of Qadence Jobs to Pasqal Cloud
+
+It is possible to submit quantum computational jobs to execute remotely on Pasqal's [cloud platform](https://portal.pasqal.cloud).
+The qadence module `qadence.pasqal_cloud_connection` offers functionality to easily specify the computation, upload the specification and retrieve the result when the computation has finished execution on the cloud platform.
+In this tutorial a simple quantum circuit will be defined as an example to show how the submission process for remote computations.
+The same process can be applied to run more complex quantum circuits on the cloud platform.
+
+Let's first define a very simple quantum circuit that creates a Bell state.
+
+```python
+from qadence import CNOT, H, QuantumCircuit
+
+circuit = QuantumCircuit(2, H(0), CNOT(0, 1))
+```
+
+If we want to upload this circuit to the cloud platform we need to follow 4 steps:
+- Authentication and making connection to cloud
+- Defining workload specification
+- Submission
+- Retrieval of results
+
+## Authentication and connection
+To setup a connection the cloud platform, use the `SDK` object present in `qadence.pasqal_cloud_connection`. The email and password are the ones used to login to the webportal. The project-id can be found in the webportal under "Projects".
+
+```python
+from qadence.pasqal_cloud_connection import SDK
+
+connection = SDK("john.doe@email.com", "my-password", project_id="proj1")
+```
+
+## Defining workload specification
+To define quantum computation we create a `WorkloadSpec` object that contains all information that is needed to execute the computation.
+To create a workload specification we need some extra information on top the circuit itself.
+We need to specify the backend, which here we will choose for PyQTorch. The cloud platform only supports PyQTorch and EMU-C.
+We need to define the result types that we want.
+These define the run method (`run`, `sample` or `expectation`) that is used on the `QuantumModel` that will be build on the basis of this workload specification.
+We can choose them through the enum value `ResultTypes`.
+These are provided in a list, so that multiple result types can be requested in a single submission.
+
+```python
+from qadence import BackendName
+from qadence.pasqal_cloud_connection import WorkloadSpec, ResultTypes
+
+workload = WorkloadSpec(circuit, BackendName.PYQTORCH, [ResultTypes.SAMPLE, ResultTypes.RUN])
+```
+
+If you already have your quantum computation defined as a `QuantumModel`, it is possible to create a workload specification directly from the model using `get_spec_from_model`.
+Then, the circuit and backend specifications will be extracted from the model, the other values need to be provided as extra arguments.
+
+### Using a Quantum Model
+
+```python
+from qadence.pasqal_cloud_connection import get_spec_from_model
+
+model = QuantumModel(...)
+workload = get_spec_from_model(model, [ResultType.SAMPLE])
+```
+
+### Observable Expectation Value
+
+For the result type `ResultType.EXPECTATION` it is mandatory to provide an observable to the workload specification.
+In the example below we use the trivial identity observable `I(0) * I(1)`.
+
+```python
+workload = WorkloadSpec(circuit, BackendName.PYQTORCH, [ResultTypes.EXPECTATION], observable=I(0)*I(1))
+```
+
+### Parametric Circuits
+In the case of a parametric circuit, i.e. a circuit that contains feature parameters or variational parameters, values for these parameters need to be provided.
+For feature parameters is it is mandatory to set a value in the workload specification, for variational parameters it is optional.
+If no values are set for a variational parameter, one will be assigned randomly.
+The parameter values are defined in a dictionary, where the keys are equal to the parameter name and the values are equal to the parameter value.
+The parameter values should be passed as torch tensors.
+It is possible to set multiple values by using a 1-D tensor, to the parameters, in that case the computation is executed for each value in the tensor.
+A mix of 0-D and 1-D tensors can be provided to keep some parameters constant and others changed during this process.
+However, all 1-D tensors need to have the same length.
+
+
+```python
+parametric_circuit = ...
+parameter_values = {"param1": tensor(0), "param2": tensor([0, 1, 2]), "param3": tensor([5, 6, 7])}
+workload = WorkloadSpec(parametric_circuit, BackendName.PYQTORCH, [ResultTypes.EXPECTATION], parameter_values=parameter_values)
+```
+
+## Submission
+Submission to the cloud platform is done very easily using the `upload_workload` function.
+The workload id will be provided by executing the function.
+This id is needed later, to request the status of the given workload.
+
+```python
+from qadence.pasqal_cloud_connection import upload_workload
+
+workload_id = upload_workload(connection, workload)
+```
+
+## Retrieval of Results
+The next step is to retrieve the results from the cloud platform.
+This can however only be done once the job is finished with it's execution on the platform.
+The `check_status` function can be used to see if the workload is finished already.
+If so, the results of the computation will be provided in a `WorkloadResult` object.
+The result of the computation itself can be found in the `result` attribute of this object.
+If the workload has not finished yet, or resulted in an error, `check_status` will raise an exception, either a `WorkloadStoppedError` or `WorkloadNotDoneError`.
+
+```python
+from qadence.pasqal_cloud_connection import check_status
+
+workload_result = check_status(connection, workload_id)
+print(workload_result.result)
+```
+
+A more convenient way of interacting retrieving the result, especially if the execution the cloud is not expected to take long, is to use the `get_result` function.
+This function checks in set intervals the status of the workload until the workload is finished or the function has timed out.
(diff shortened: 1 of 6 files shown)


pimvenderbosch:
Hi @RolandMacDoland,

I've pushed a prototype of how the cloud integration could look like in qadence. The actual implementation is not there yet. Could you quickly review to see that this goes into the desired direction?

RolandMacDoland (commented):
Thanks for this @pimvenderbosch. LGTM. Minor comments but I think they might be irrelevant after I went through the code. Happy to approve after threads resolution. Another point, I think there is a way to test the cloud connection internally. But we need to discuss this.

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class WorkloadStoppedError(Exception):
+    """Is raised when a workload has stopped running on remote for some reason."""
+
+    pass
+
+
+def check_status(connection: SDK, workload_id: str) -> WorkloadResult:
+    """Checks if the workload is succesfully finished on remote connection.
```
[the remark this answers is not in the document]
pimvenderbosch: Thanks, will make a change to the Google format!

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,109 @@
...
+        case "DONE":
+            return result
+        case "CANCELED" | "TIMED_OUT" | "ERROR":
+            raise WorkloadStoppedError(
+                f"Workload with id {workload_id} couldn't finish, the status is {result.status}"
+            )
+        case _:
+            raise ValueError(
+                f"Undefined workload status ({result.status}) was returned for "
+                + f"workload ({result.id})"
```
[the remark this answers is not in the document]
pimvenderbosch: Ah yeah indeed, will remove.

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,109 @@
...
+    """Uploads a workload to Pasqal's Cloud and returns the created workload ID."""
+    workload_json = workload_spec_to_json(workload)
+    remote_workload = connection.create_workload(
+        workload_json.workload_type, workload_json.backend_type, workload_json.config
+    )
+    workload_id: str = remote_workload.id
+    return workload_id
+
+
+class WorkloadNotDoneError(Exception):
```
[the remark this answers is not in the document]
pimvenderbosch: From the `pasqal_cloud` package's perspective these aren't errors. The workload in the cloud database has a status, which gets updated when the workload is progressed. From a qadence user perspective, you can't proceed with the analysis of the results unless the workload status is `"DONE"`. That's why I created these custom errors.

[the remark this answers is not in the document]
pimvenderbosch: There is one case where we need to add context from what is given from the cloud : status: `”ERROR”`. In that case there is error information in the result object that we need to shot to the user. I will add this.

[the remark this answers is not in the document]
pimvenderbosch: Added this feature in commit [57265a4](https://github.com/pasqal-io/qadence/pull/592/commits/57265a4d4ae0ceaaf1ce55549549b8940d4c4a80)

docs/tutorials/advanced_tutorials/cloud-submission.md
```diff
@@ -0,0 +1,120 @@
...
+```python
+workload = WorkloadSpec(circuit, BackendName.PYQTORCH, [ResultTypes.EXPECTATION], observable=I(0)*I(1))
+```
+
+### Parametric Circuits
+In the case of a parametric circuit, i.e. a circuit that contains feature parameters or variational parameters, values for these parameters need to be provided.
+For feature parameters is it is mandatory to set a value in the workload specification, for variational parameters it is optional.
+If no values are set for a variational parameter, one will be assigned randomly.
+The parameter values are defined in a dictionary, where the keys are equal to the parameter name and the values are equal to the parameter value.
+The parameter values should be passed as torch tensors.
```
RolandMacDoland, line 73:
```suggestion
The parameter values are defined in a dictionary, where keys are the parameter name and values are parameter value passed as torch tensors.
```

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,238 @@
...
+    """Specification of a workload to be executed on Pasqal Cloud.
+
+    This data class defines a single workload specification that is to be executed on Pasqal's
+    cloud platform.
+
+    Args:
+        circuit: The quantum circuit to be executed.
+        backend: The backend to execute the workload on. Not all backends are available on the
+            cloud platform. Currently the supported backends are `BackendName.PYQTORCH` and
+            `"emu_c"`.
```
RolandMacDoland, line 33:
```suggestion
            `BackendName.EMUC`.
```
Although I am wondering if we should mention a private emulator in a public repo.

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,238 @@
...
+            The observable field is mandatory in this case. If not, the value of this field will
+            be ignored. Only a single observable can be passed for cloud submission; providing a
+            list of observables is not supported.
+    """
+
+    circuit: QuantumCircuit
+    backend: BackendName | str
+    result_types: list[ResultType]
+    parameter_values: Optional[dict[str, Tensor]] = None
+    observable: Optional[AbstractBlock] = None
```
RolandMacDoland, line 52:
```suggestion
    parameter_values: dict[str, Tensor] | None = None
    observable: AbstractBlock | None = None
```

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,238 @@
...
+    backend_type: str
+    config: dict[str, Any]
+    workload_type = "qadence_circuit"
+
+
+def _parameter_values_to_json(parameter_values: Optional[dict[str, Tensor]]) -> str:
+    parameter_values_dict: dict[str, str] = dict()
+    if parameter_values is not None:
+        for key, value in parameter_values.items():
+            parameter_values_dict[key] = value.tolist()
```
RolandMacDoland, line 118: Maybe use a dict comprehension here ?

RolandMacDoland: Also: couldn't that be done in-place ?

tests/test_pasqal_cloud_connection.py
```diff
@@ -0,0 +1,194 @@
...
+    circuit = BasicQuantumCircuit
+    result_types = [ResultType.RUN, ResultType.SAMPLE]
+    workload = WorkloadSpec(
+        circuit, BackendName.PYQTORCH, result_types, {"my-parameter": tensor(3)}, I(0) * I(1)
+    )
+    result = upload_workload(mock_connection, workload)
+    assert result == expected_workload_id
+
+
+@pytest.fixture
```
RolandMacDoland, line 128: Move to `conftest.py` ?

RolandMacDoland (approved):
Thanks a lot @pimvenderbosch. Various minor comments throughout otherwise LGTM.

RolandMacDoland:
Ah @pimvenderbosch I forgot to mention: we should remove mentions to EMUC and put them in extensions I think. Also, `Optional -> |` whenever possible. Thanks !

tests/test_pasqal_cloud_connection.py
pimvenderbosch: This fixture is very specific to the tests defined in this module. I'd rather keep the fixture here.

docs/tutorials/advanced_tutorials/cloud-submission.md
```diff
@@ -0,0 +1,124 @@
...
+```python
+from qadence.pasqal_cloud_connection import submit_workload
+
+workload_id = submit_workload(connection, workload)
+```
+
+## Check Workload Status
+
+The `check_status` function can be used to see if the workload is finished already.
+The status of a workload can be: done, pending, running, paused, canceled, timed out or error.
```
MatthieuMoreau0, line 101:
```suggestion
The status of a workload can be: done, pending, running, paused, canceled, timed out or error.
The `check_status` function can be used to see if the workload is finished already.
```

So that it fits nicely with the next sentence

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,235 @@
...
+    observable: AbstractBlock | None = None,
+) -> WorkloadSpec:
+    """Creates a `WorkloadSpec` from a quantum model.
+
+    This function creates a `WorkloadSpec` from a `QuantumModel` and the other arguments provided.
+    The circuit, that is extracted from the model, is the original circuit that was used to
+    initialize the model, not the backend converted circuit in `model.circuit`. The backend set in
+    the model will be used in the workload specification.
+
+    It is important to note that in case there is an observable defined in the model, it is ignored
```
MatthieuMoreau0, line 84: I don't know much about qadence so this may be obvious; but why do we ignore the observable defined in the model and not reuse it for the workload?

docs/tutorials/advanced_tutorials/cloud-submission.md
```diff
@@ -0,0 +1,124 @@
...
+connection = SDK("john.doe@email.com", "my-password", project_id="proj1")
+```
+## Defining workload specification
+
+A workload is a quantum calculation to execute on a Pasqal cloud backend.
+To create a workload specification, we need some extra information on top the circuit itself.
+We need to specify the backend, chosen here to be PyQTorch.
+The cloud platform currently only supports PyQTorch.
+Moreover, the requested result type needs to be defined.
+Based on the workload specification, the appropriate run methods (`run`, `sample` or `expectation`) will be called by the `QuantumModel` by passing them through the enum value `ResultTypes` argument.
```
MatthieuMoreau0, line 39: Out of curiosity (not for this MR), are there other types of result that we should support in the cloud for qadence workloads?

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,235 @@
...
+def get_spec_from_model(
+    model: QuantumModel,
+    result_types: list[ResultType],
+    parameter_values: dict[str, Tensor] | None = None,
+    observable: AbstractBlock | None = None,
+) -> WorkloadSpec:
+    """Creates a `WorkloadSpec` from a quantum model.
+
+    This function creates a `WorkloadSpec` from a `QuantumModel` and the other arguments provided.
+    The circuit, that is extracted from the model, is the original circuit that was used to
```
MatthieuMoreau0, line 80: Again out of curiosity, since we extract the backend from the quantum model why do we use the original circuit and not the backend converted one?

qadence/pasqal_cloud_connection.py
```diff
@@ -0,0 +1,235 @@
...
+    workload_type = "qadence_circuit"
+
+
+def _parameter_values_to_json(parameter_values: dict[str, Tensor] | None) -> str:
+    if parameter_values is None:
+        return json.dumps(dict())
+    return json.dumps({key: value.tolist() for key, value in parameter_values.items()})
+
+
+def workload_spec_to_json(workload: WorkloadSpec) -> WorkloadSpecJSON:
```
MatthieuMoreau0, line 119: nit: AFAICT, this is not meant to be called by the user so I would suggest making this function "private" and prepending it with an underscore

qadence/pasqal_cloud_connection.py
pimvenderbosch: In qadence you can define the observable either in the model or when you call `.expectation(...)`. I opted for only supporting the latter option here, because then I could more easily check if it is there. Because it is mandatory to provide one, if you use `expectation`.

docs/tutorials/advanced_tutorials/cloud-submission.md
pimvenderbosch: No, these are the only three that exist in qadence.

qadence/pasqal_cloud_connection.py
pimvenderbosch: The workload handles takes in a raw circuit, not a converted one. I agree that this will cause duplicate work, therefor the preferred route should be through the quantum circuit and not from a model.

Issue #570, opened 2024-09-17
[Refactoring] Training functions and TrainConfig can be made more modular


debrevitatevitae:
### Describe the feature

Refactoring the training system (training functions and training configuration) in more internal classes or functions.

### It should be implemented because

Currently, there are implementation and readability problems about training functions and `TrainConfig`

- training functions are too long (`train_grad` counts more than 360 lines of code!). Therefore, on one hand, they have too many responsibilities, which makes them hard to maintain and extend. On the other hand they are hard to read.
- `TrainConfig` unloads too much responsibility to the `__post_init__`, which causes unwanted behavior such as [this one](https://github.com/pasqal-io/qadence/issues/569).

Modularization and responsibility separation will help in the readability, maintenance and extensibility of the training system

### Would you like to work on this issue?

None

mlahariya:
Hey @debrevitatevitae , Thanks for raising this. For the refactoring, I had a few ideas and suggestions. 

1) **Logger**: We can move the logging part of the training functions outside into a separate class (Logger). An instance of this class could take care of all logging operations - including defining callbacks and providing methods to log at different steps in the training loop
2) **ConfigHandler**: We can define a config handler class that would be a collection of methods to handle TrainConfig. This would help us in initial and end stages of the training - where we can define separate methods based on the configuration provided by the user. We can also mode the `__post_init__` methods into this class, allowing us to resolve [this issue](https://github.com/pasqal-io/qadence/issues/569)

The simple outline of the train function would look something like this

`Trainer`
    
    Train(args):
      logger = Logger(config)
  
      confighandle = ConfigHandle(config)
  
      # training 
      # and logging 
  
      # end logging and close the logger`



I had two thoughts where I needed inputs though:
 Q1:  Currently this is defined as a Train function. Can we move it to a Train class ? (a structure similar to the trainer function from PyTorch-lightning [trainer](https://lightning.ai/docs/pytorch/stable/common/trainer.html)). We would be able to define a base fit() method inside this class - which could also be inherited for other purposes?
 Q2: Do we need the separate functions - train_with_grad and train_without_grad? Can we have a single Train (function/class) - that can be used for either based on a user defined argument?
 So, do we need the trainer to be a function and do we need separate functions for train_with_grand/train_without_grad? 
 
  
Let me know what you think. Once these ideas are refined - I will start working on a PR for this. 


@chMoussa @Roland-djee @smitchaudhary @n-toscano 
Thanks
M

[...]

Issue #569, opened 2024-09-16
[ml_tools] A train config used more than once overwrites previous results.


debrevitatevitae:
### Short description

If we train a model more than once in a row and we want to use the same training settings (i.e. same `TrainConfig`), every training will overwrite the results of the previous ones.

### What is the expected result?

When the option `create_subfolder_per_run=True` and training a second time with the **same** `TrainConfig`, a new subdirectory should be created, where the results of the second training are stored. In other words, the results of the second training should not overwrite the results of the first one.

### What is the actual result?

In the situation above, the results of the first training are overwritten.

### Steps/Code to reproduce

```python
from itertools import count

import torch
from qadence import (
    feature_map,
    hamiltonian_factory,
    hea,
    QNN,
    QuantumCircuit,
    TrainConfig,
    Z,
)
from qadence.ml_tools import to_dataloader, train_with_grad
from torch.utils.data import DataLoader


def dataloader(data_size: int = 25, batch_size: int = 5, infinite: bool = True) -> DataLoader:
    x = torch.linspace(0, 1, data_size).reshape(-1, 1)
    y = torch.sin(x)
    return to_dataloader(x, y, batch_size=batch_size, infinite=infinite)


n_qubits = 2
fm = feature_map(n_qubits)
ansatz = hea(n_qubits=n_qubits, depth=2)
observable = hamiltonian_factory(n_qubits, detuning=Z)
circuit = QuantumCircuit(n_qubits, fm, ansatz)

model = QNN(circuit, observable, backend="pyqtorch", diff_mode="ad")
batch_size = 100
input_values = {"phi": torch.rand(batch_size, requires_grad=True)}

cnt = count()
criterion = torch.nn.MSELoss()
optimizer = torch.optim.Adam(model.parameters(), lr=0.1)


def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, dict]:
    next(cnt)
    x, y = data[0], data[1]
    out = model(x)
    loss = criterion(out, y)
    return loss, {}


config = TrainConfig(max_iter=10, print_every=2, folder="dev", create_subfolder_per_run=True)

# train a first time
train_with_grad(model, dataloader(), optimizer, config, loss_fn)

# train a second time
train_with_grad(model, dataloader(), optimizer, config, loss_fn)

```

### Would you like to work on this issue?

None

PR #593, opened 2024-10-30
[Refactor][Feature] Train functions refactoring

mlahariya:
Solves #570 and #569. We introduce a Trainer class that will handle the training process. Along with this, we also introduce a callback system - including callbacks for different operations, as well a callback manager tied closely with Trainer. A config manager handles the TrainConfig (which solves the logging directory problem). 

image


mlahariya:
Hey @chMoussa , @RolandMacDoland , @debrevitatevitae, 

I have updated the ml_tools according to the refactor above, along with modifying the docs and tests. Please have a look.

docs/tutorials/qml/ml_tools/callbacks.md
```diff
@@ -0,0 +1,231 @@
+
+# Callbacks for Trainer
+
+Qadence ml_tools provides a powerful callback system for customizing various stages of the training process. With callbacks, you can monitor, log, save, and alter your training workflow efficiently. A `CallbackManager` is used with [`Trainer`][qadence.ml_tools.Trainer] to excute the training process with defined callbacks. The default callbacks already provided in the [`Trainer`][qadence.ml_tools.Trainer] are
```
chMoussa, line 4: Small typos (excute -> execute and finish sentence

mkdocs.yml
```diff
@@ -42,7 +42,9 @@ nav:
 
     - Variational quantum algorithms:
       - tutorials/qml/index.md
-      - Training tools: tutorials/qml/ml_tools.md
+      - Training: tutorials/qml/ml_tools/trainer.md
+      - Training Callbacks: tutorials/qml/ml_tools/callbacks.md
+      - Data and Configrations: tutorials/qml/ml_tools/data_and_config.md
```
chMoussa, line 47: Data and Configurations

qadence/ml_tools/callbacks/callbackmanager.py
```diff
@@ -0,0 +1,216 @@
...
+        callback_map = {
+            "PrintMetrics": PrintMetrics,
+            "WriteMetrics": WriteMetrics,
+            "PlotMetrics": PlotMetrics,
+            "SaveCheckpoint": SaveCheckpoint,
+            "LoadCheckpoint": LoadCheckpoint,
+            "LogModelTracker": LogModelTracker,
+            "LogHyperparameters": LogHyperparameters,
+            "SaveBestCheckpoint": SaveBestCheckpoint,
+        }
```
chMoussa, line 148: Instead of creating the dict every time at each call of `get_callback_by_name`, can we just set it outside and index from it?

qadence/ml_tools/callbacks/callbackmanager.py
```diff
@@ -0,0 +1,216 @@
...
+        Returns:
+            Any: Results of the executed callbacks.
+        """
+        return [
+            callback(when=trainer.state, trainer=trainer, config=self.config, writer=self.writer)
+            for callback in self.callbacks
+            if callback.on == trainer.state
+        ]
+
+    def start_training(self, trainer: Any) -> None:
```
chMoussa, line 167: Why Any?

qadence/ml_tools/callbacks/callbackmanager.py
```diff
@@ -0,0 +1,216 @@
...
+
+        trainer.opt_result = OptimizeResult(trainer.global_step, trainer.model, trainer.optimizer)
+        trainer.is_last_iteration = False
+
+        # Load checkpoint if available
+        load_checkpoint_callback = LoadCheckpoint(on="on_train_start", called_every=1)
+        loaded_result = load_checkpoint_callback.run_callback(
+            trainer=trainer,
+            config=self.config,
+            writer=MLFlowWriter(),  # adding empty writer to avoid [arg-type] error
```
chMoussa, line 185: Why Mlflowwriter is instantieted?

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+from qadence.types import ExperimentTrackingTool
+
+logger = getLogger(__name__)
+
+# Type aliases
+PlottingFunction = Callable[[Module, int], tuple[str, Figure]]
+InputData = Union[Tensor, dict[str, Tensor]]
+
+
+class BaseWriter:
```
chMoussa, line 27: Add inheritance from ABC ?

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+    """Writer for logging to TensorBoard.
+
+    Attributes:
+        writer (SummaryWriter): The TensorBoard SummaryWriter instance.
+    """
+
+    def __init__(self) -> None:
+        self.writer = None
+
+    def open(self, config: TrainConfig, iteration: int = None) -> SummaryWriter:
```
chMoussa, line 138: for typing iteration, use int | None = None

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+
+        Args:
+            config: Configuration object containing settings for logging.
+            iteration (int, optional): The iteration step to start logging from.
+                Defaults to None.
+
+        Returns:
+            mlflow: The MLflow module instance.
+        """
+        import mlflow
```
chMoussa, line 249: why the import inside?

chMoussa (commented):
@mlahariya Quite an impressive refactoring, overall it looks quite well organised with the Trainer class. Small question on the config, what could you have in folder in addition to the _log_folder?

In terms of tests, do we have tests for checking what happens when we run twice the optimization?

docs/tutorials/qml/ml_tools/callbacks.md
```diff
@@ -0,0 +1,231 @@
+
+# Callbacks for Trainer
+
+Qadence ml_tools provides a powerful callback system for customizing various stages of the training process. With callbacks, you can monitor, log, save, and alter your training workflow efficiently. A `CallbackManager` is used with [`Trainer`][qadence.ml_tools.Trainer] to execute the training process with defined callbacks. The default callbacks already provided in the [`Trainer`][qadence.ml_tools.Trainer].
```
RolandMacDoland, line 4:
```suggestion
Qadence `ml_tools` provides a powerful callback system for customizing various stages of the training process. With callbacks, you can monitor, log, save, and alter your training workflow efficiently. A `CallbackManager` is used with [`Trainer`][qadence.ml_tools.Trainer] to execute the training process with defined callbacks. The default callbacks already provided in the [`Trainer`][qadence.ml_tools.Trainer].
```

qadence/ml_tools/callbacks/writer_registry.py
debrevitatevitae: and `@abstractmethod`s

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+
+        Returns:
+            SummaryWriter: The initialized TensorBoard writer.
+        """
+        log_dir = str(config._log_folder)
+        if isinstance(iteration, int):
+            self.writer = SummaryWriter(log_dir=log_dir, purge_step=iteration)
+        else:
+            self.writer = SummaryWriter(log_dir=log_dir)
+        return self.writer
```
debrevitatevitae, line 155: Maybe refactor as
```suggestion
        purge_step = iteration if isinstance(iteration, int) else None
        self.writer = SummaryWriter(log_dir=log_dir, purge_step=purge_step)
```

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+
+        Args:
+            result (OptimizeResult): The optimization results to log.
+        """
+        # Not writing loss as loss is available in the metrics
+        # if result.loss is not None:
+        #     self.writer.add_scalar("loss", float(result.loss), result.iteration)
+        if self.writer:
+            for key, value in result.metrics.items():
+                self.writer.add_scalar(key, value, result.iteration)
```
debrevitatevitae, line 174: what happens in the non-happy path? (`bool(self.writer) is False`)? Should we include an error?

qadence/ml_tools/callbacks/writer_registry.py
debrevitatevitae, line 138: I see there are a lot of `if self.writer` checks in the other methods, which require the caller to call `open` on the writer. Is there a reason why `open` shouldn't be executed directly in the `__init__`?

docs/tutorials/qml/ml_tools/callbacks.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Default Callbacks
+Below is a list of the default callbacks already implemented in the `CallbackManager` used with [`Trainer`][qadence.ml_tools.Trainer]:
+
+- **`on_train_start`**: `PlotMetrics`, `SaveCheckpoint`, `WriteMetrics`
```
RolandMacDoland, line 9: Maybe the `on` can be omitted ? Since it is the argument name, it feels a bit redundant.

docs/tutorials/qml/ml_tools/callbacks.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+### 1.3. `PlotMetrics`
+
+Plots metrics based on user-defined plotting functions.
+
+```python
+from qadence.ml_tools.callbacks import PlotMetrics
+
+plot_metrics_callback = PlotMetrics(on="on_train_epoch_end", called_every=100)
```
RolandMacDoland, line 59:
```suggestion
plot_metrics_callback = PlotMetrics(on="train_epoch_end", every=100)
```

docs/tutorials/qml/ml_tools/callbacks.md
```diff
@@ -0,0 +1,231 @@
...
+    callbacks=[log_model_callback]
+)
+```
+
+
+## 2. Custom Callbacks
+
+The base `Callback` class in Qadence allows defining custom behavior that can be triggered at specified events (e.g., start of training, end of epoch). You can set parameters such as when the callback runs (`on`), frequency of execution (`called_every`), and optionally define a `callback_condition`.
+
+### defining Callbacks
```
RolandMacDoland, line 147:
```suggestion
### Defining Callbacks
```

qadence/ml_tools/callbacks/callbackmanager.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def __init__(self, config: TrainConfig):
+        """
+        Initializes the CallbacksManager with a training configuration.
+
+        Args:
+            config (TrainConfig): The training configuration object.
+        """
+        self.config = config
+        self.callbacks: List[Callback] = []
```
RolandMacDoland, line 47:
```suggestion
        self.callbacks: list[Callback] = []
```

qadence/ml_tools/callbacks/callbackmanager.py
```diff
@@ -0,0 +1,216 @@
...
+            self.add_callback("LogHyperparameters", "on_train_end")
+        # Log model
+        if self.config.log_model:
+            self.add_callback("LogModelTracker", "on_train_end")
+        # only save the last checkpoint if not checkpoint_best_only
+        if not self.config.checkpoint_best_only:
+            self.add_callback("SaveCheckpoint", "on_train_end")
+        self.add_callback("WriteMetrics", "on_train_end")
+
+    def add_callback(self, callback: Union[Callback, str], on: str, called_every: int = 1) -> None:
```
RolandMacDoland, line 103:
```suggestion
    def add_callback(self, callback: Callback | str, on: str, called_every: int = 1) -> None:
```

qadence/ml_tools/callbacks/saveload.py
```diff
@@ -18,6 +18,7 @@ def get_latest_checkpoint_name(folder: Path, type: str, device: str | torch.devi
     file = Path("")
     files = [f for f in os.listdir(folder) if f.endswith(".pt") and type in f]
     if len(files) == 0:
+        pass
```
RolandMacDoland, line 21: Wouldn't it skip the error logging ?

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+        write(result): Writes the optimization results to the tracking tool.
+        log_hyperparams(hyperparams): Logs the hyperparameters to the tracking tool.
+        plot(model, iteration, plotting_functions): Logs model plots using provided
+            plotting functions.
+        log_model(model, dataloader): Logs the model and any relevant information.
+    """
+
+    run: Run  # [attr-defined]
+
+    def open(self, config: TrainConfig, iteration: int = None) -> Any:
```
RolandMacDoland, line 45: Shouldn't these methods be `@abstractmethod`s ?

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+
+        Args:
+            config: Configuration object containing settings for logging.
+            iteration (int, optional): The iteration step to start logging from.
+                Defaults to None.
+
+        Returns:
+            SummaryWriter: The initialized TensorBoard writer.
+        """
+        log_dir = str(config._log_folder)
```
debrevitatevitae, line 150: If `_log_folder` is accessed outside of `TrainConfig`, perhaps it shouldn't be a protected variable.

qadence/ml_tools/loss/loss.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import annotations
+
+from typing import Callable, Dict, Tuple, Union
+
+import torch
+import torch.nn as nn
+
+
+def mse_loss(
+    model: nn.Module, batch: Tuple[torch.Tensor, torch.Tensor]
```
RolandMacDoland, line 10:
```suggestion
    model: nn.Module, batch: tuple[torch.Tensor, torch.Tensor]
```

qadence/ml_tools/callbacks/writer_registry.py
debrevitatevitae, line 45: I see there are a lot of `if self.writer` checks in the other methods, which require the caller to call `open` on the writer. Is there a reason why `open` shouldn't be executed directly in the `__init__`?

From `callbackmanager.py`:
```python
writer = get_writer(tracking_tool)
writer.open(self.config, iteration=trainer.global_step)
```
from which I see that `open` is called just after initialization. I think then that we can `open` _at_ initialization.

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -0,0 +1,375 @@
...
+                    xs = {key: val.detach().cpu().numpy() for key, val in xs.items()}
+                    preds = {key: val.detach().cpu().numpy() for key, val in preds.items()}
+                try:
+                    from mlflow.models import infer_signature
+
+                    signature = infer_signature(xs, preds)
+                except ImportError:
+                    logger.warning(
+                        "MLflow's infer_signature is not available. Please install mlflow."
+                    )
```
debrevitatevitae, line 350: This looks a bit long and nested. Can we split it?

RolandMacDoland (commented):
Thanks @mlahariya. This is a massive so I only went half-way. It looks good, minor comments about renaming and the use of annotations allowed by `from __future__ import annotations` where part of `typing` is included in the standard library with aliases that make Python code a bit more readable. I encourage you to use it as much as possible as it is also idiomatic to Qadence and getting rid as much as possible of `typing` features.

debrevitatevitae (commented):
Hey @mlahariya, really impressive work.

I left a bunch of comments, but for now only on docs, config and writer. I am happy to discuss them and I will review the rest of the changes ASAP.

docs/tutorials/qml/ml_tools/trainer.md
```diff
@@ -0,0 +1,615 @@
...
+cnt = count()
+criterion = torch.nn.MSELoss()
+
+def loss_fn(model: torch.nn.Module, data: torch.Tensor) -> tuple[torch.Tensor, dict]:
+    next(cnt)
+    x, y = data
+    out = model(x)
+    loss = criterion(out, y)
+    return loss, {}
+```
```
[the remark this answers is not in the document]
mlahariya: Added.

qadence/ml_tools/train_utils/config_manager.py
```diff
@@ -0,0 +1,159 @@
...
+                self.config.hyperparams.pop(key)
+
+    def _derive_parameters(self) -> None:
+        """
+        Derive additional parameters for the training configuration.
+
+        Sets the stopping criterion if it is not already defined.
+        """
+        if self.config.trainstop_criterion is None:
+            self.config.trainstop_criterion = lambda x: x <= self.config.max_iter
```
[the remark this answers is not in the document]
mlahariya: It would be better to keep an open ended nomenclature - in case we would like to derive other parameters based on config in the future.
trainstop_criterion could be only one of the derivations. What do you think?

qadence/ml_tools/callbacks/saveload.py
mlahariya: Indeed. Corrected it.

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: We can make it public. My rational to keep log_folder and subfolder private in the TrainConfig was - the user will not be able to set these values - they will be for internal use, and will be updated internally by the configmanager. All other attributes of the TrainConfig can be modified by the user, but not these. What do you think?

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: We will need to have a method to set up the writer in case the log_folder changes during runtime. For example when create_subfolders_per_run = True, a new writer will have to be set up for each run (based on a new log_dir), or we will have to modify the existing writer - both of which will need a writer opening method at the start of training.

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: Sure. We can raise a runtime error.

docs/tutorials/qml/ml_tools/trainer.md
```diff
@@ -0,0 +1,615 @@
...
+
+
+config = TrainConfig(
+    folder="mlflow_demonstration",
+    max_iter=10,
+    checkpoint_every=1,
+    plot_every=2,
+    write_every=1,
+    log_model=True,
+    tracking_tool=ExperimentTrackingTool.MLFLOW,
```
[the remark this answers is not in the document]
mlahariya: What would this look like? If no tracking is available, only printing and checkpointing will be possible. Do we want this behaviour?

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: 1) We will need to have a method to set up the writer in case the log_folder changes during runtime. For example when create_subfolders_per_run = True, a new writer will have to be set up for each run (based on a new log_dir), or we will have to modify the existing writer - both of which will need a writer opening method at the start of training. 

2) I wanted to keep the init for the writers as clean as possible - so that the get_writer function only has one argument - which is the tracking tool - and returns the object of a particular writer. After which we can open the writer using TrainConfig and iterations - the user can also open the writer with a different Config, etc. 
This might be useful if the user wants to train - without Trainer - but still want to use our writers.

```
writer = get_writer(tracking_tool)
writer.open(self.config, iteration=trainer.global_step)

writer.close()

writer.open(Differfent_config, iteration=trainer.global_step)
writer.close()
```

docs/tutorials/qml/ml_tools/trainer.md
```diff
@@ -0,0 +1,615 @@
...
+The available hooks include:
+
+- `on_train_start`: Called at the beginning of the training process.
+- `on_train_end`: Called at the end of the training process.
+- `on_train_epoch_start`: Called at the start of each training epoch.
+- `on_train_epoch_end`: Called at the end of each training epoch.
+- `on_train_batch_start`: Called at the start of each training batch.
+- `on_train_batch_end`: Called at the end of each training batch.
+
+Example of using a hook to log a message at the end of each epoch:
```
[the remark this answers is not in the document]
mlahariya: Indeed. 
Callbacks are mainly for logging/tracking purposes - but these hooks are generic. 
Eg. In every train batch start, 
1. perform on_train_batch_start callbacks (using the decorator BaseTrainer.callback("train_batch"))
2. call the on_train_batch_start hook
3. do the batch training
4. call the on_train_batch_end hook
5. perform on_train_batch_end callbacks (using the decorator BaseTrainer.callback("train_batch"))

The use of these hooks is not specifically to add extra callbacks, but for any other generic pre/post processing 
1. eg, reshaping input batch in case of RNNs/LSTMs, post processing loss and adding an extra metric. 
2. they could also be used to add more callbacks (which is not recommended - as we provide methods to add extra callbacks in the TrainCofig)

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: Updated.

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: Updated.

qadence/ml_tools/train_utils/base_trainer.py
```diff
@@ -0,0 +1,518 @@
...
+    def set_use_grad(cls, value: bool) -> None:
+        """
+        Sets the global use_grad flag.
+
+        Args:
+            value (bool): Whether to use gradient-based optimization.
+        """
+        if not isinstance(value, bool):
+            raise TypeError("use_grad must be a boolean value.")
+        cls.use_grad = value
```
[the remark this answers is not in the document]
mlahariya: Currently we have:
1.  Contextmanagers (enable_grad_opt/disable_grad_opt) to change use_grad at an object level. 

2. And set_use_grad() to change use_grad at a class level. 
      > This might be useful in case of parallel processing, where we would like to change use_grad for all instances of the class. Also helps avoiding an extra line of code to set use_grad in case we have multiple trainings - where we can set it up for all instances of Trainer. 


However, we can indeed make use_grad a property with a setter method. This would serve the same purpose as the context managers - to change use_grad at a object level. We can keep either the contextmanager or the setter or both. I added the context managers because they are more pythonic - and the confusion a setter method (trainer.use_grad = True) could cause with classmethod (Trainer.set_use_grad())

Should i just add a setter for use_grad as well?

qadence/ml_tools/callbacks/callbackmanager.py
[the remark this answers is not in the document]
mlahariya: Updated.

docs/tutorials/qml/ml_tools/callbacks.md
mlahariya: Updated.

docs/tutorials/qml/ml_tools/callbacks.md
```diff
@@ -0,0 +1,231 @@
...
+
+Qadence ml_tools provides a powerful callback system for customizing various stages of the training process. With callbacks, you can monitor, log, save, and alter your training workflow efficiently. A `CallbackManager` is used with [`Trainer`][qadence.ml_tools.Trainer] to execute the training process with defined callbacks. The default callbacks already provided in the [`Trainer`][qadence.ml_tools.Trainer].
+
+### Default Callbacks
+Below is a list of the default callbacks already implemented in the `CallbackManager` used with [`Trainer`][qadence.ml_tools.Trainer]:
+
+- **`on_train_start`**: `PlotMetrics`, `SaveCheckpoint`, `WriteMetrics`
+- **`on_train_epoch_end`**: `SaveCheckpoint`, `PrintMetrics`, `PlotMetrics`, `WriteMetrics`
+- **`on_val_epoch_end`**: `SaveBestCheckpoint`, `WriteMetrics`
+- **`on_train_end`**: `LogHyperparameters`, `LogModelTracker`, `WriteMetrics`, `SaveCheckpoint`
```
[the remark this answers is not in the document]
mlahariya: I have added the plot metrics to the train_end. However, I am a bit concerned about the amount of default callbacks we have in the trainer. Currently we have the following callbacks as default: which might seem like an overkill. Maybe we should reduce them in the future - and only keep the necessary ones. 

```
        # Train Start
        self.add_callback("PlotMetrics", "train_start")
        if self.config.val_every:
            self.add_callback("WriteMetrics", "train_start")
            # only save the first checkpoint if not checkpoint_best_only
            if not self.config.checkpoint_best_only:
                self.add_callback("SaveCheckpoint", "train_start")

        # Checkpointing
        if self.config.checkpoint_best_only:
            self.add_callback("SaveBestCheckpoint", "val_epoch_end", self.config.val_every)
        elif self.config.checkpoint_every:
            self.add_callback("SaveCheckpoint", "train_epoch_end", self.config.checkpoint_every)

        # Printing
        if self.config.verbose and self.config.print_every:
            self.add_callback("PrintMetrics", "train_epoch_end", self.config.print_every)

        # Plotting
        if self.config.plot_every:
            self.add_callback("PlotMetrics", "train_epoch_end", self.config.plot_every)

        # Writing
        if self.config.write_every:
            self.add_callback("WriteMetrics", "train_epoch_end", self.config.write_every)
        if self.config.val_every:
            self.add_callback("WriteMetrics", "val_epoch_end", self.config.val_every)

        # Train End
        # Hyperparameters
        if self.config.hyperparams:
            self.add_callback("LogHyperparameters", "train_end")
        # Log model
        if self.config.log_model:
            self.add_callback("LogModelTracker", "train_end")
        if self.plot_every:
            self.add_callback("PlotMetrics", "train_end")
        # only save the last checkpoint if not checkpoint_best_only
        if not self.config.checkpoint_best_only:
            self.add_callback("SaveCheckpoint", "train_end")
        self.add_callback("WriteMetrics", "train_end")
```

qadence/ml_tools/train_utils/base_trainer.py
```diff
@@ -0,0 +1,518 @@
...
+
+    Attributes:
+        use_grad (bool): Indicates if gradients are used for optimization. Default is True.
+
+        _model (nn.Module): The neural network model.
+        _optimizer (Union[optim.Optimizer, NGOptimizer, None]]): The optimizer for training.
+        _config (TrainConfig): The configuration settings for training.
+        _train_dataloader (Optional[DataLoader]): DataLoader for training data.
+        _val_dataloader (Optional[DataLoader]): DataLoader for validation data.
+        _test_dataloader (Optional[DataLoader]): DataLoader for testing data.
```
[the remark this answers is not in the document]
mlahariya: Ah, this was a mistake. I wanted to add properties as attributes, not the protected variables. I'll update it.

qadence/ml_tools/callbacks/callbackmanager.py
mlahariya: Updated.

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: Updated.

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: Moved outside. 

Kept it inside to avoid unnecessarily importing mlflow in case it was not needed.

qadence/ml_tools/callbacks/callbackmanager.py
mlahariya: This is because of circular imports between Trainer and CallbackManager.

qadence/ml_tools/train_utils/config_manager.py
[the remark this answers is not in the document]
debrevitatevitae: Fair, but the name remains a bit obscure to me. E.g. one could think of `parameters` as the trainable ones. Why not calling it `setup_additional_configuration` or something like that?

qadence/ml_tools/train_utils/base_trainer.py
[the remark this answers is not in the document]
debrevitatevitae: I see. Perhaps it makes more sense to me if the two functionalities were inverted. If `enable_grad_opt` was independent from `Trainer`, then inside that context manager, every new instance of `Trainer` would be spawned with `use_grad=True` (perhaps using a global variable). Correspondingly for `disable_grad_opt`.

This would remove the need of `set_use_grad` and we could have `@property` and `@setter.` around `use_grad` to allow modifications at the instance level.

docs/tutorials/qml/ml_tools/callbacks.md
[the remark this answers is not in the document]
debrevitatevitae: I agree. That's related to my suggestion of having a no-tracking option (for that matters, maybe also a no-checkpointing options) and even to keep these zero settings as default. In this way the default `Trainer` would be as lean as possible and it's up to the user to add plugins to `Trainer`, when they really want to record something.

docs/tutorials/qml/ml_tools/trainer.md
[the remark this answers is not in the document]
debrevitatevitae: Right, thanks for the explanation. Is the difference clear from the docs? Maybe I missed it.

qadence/ml_tools/callbacks/writer_registry.py
debrevitatevitae: Sure. My suggestion would be to use a read-only mechanism.
```python
class TrainConfig:
    def __init__(..., log_folder,...)
        self._log_folder = log_folder

    @property
    def log_folder(self):
        return self.log_folder
``` 
This makes `log_folder` readable but not settable.

qadence/ml_tools/callbacks/writer_registry.py
debrevitatevitae: Sure, sounds good.

qadence/ml_tools/train_utils/config_manager.py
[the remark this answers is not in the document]
mlahariya: Updated.

qadence/ml_tools/train_utils/base_trainer.py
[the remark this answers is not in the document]
mlahariya: I would be opposed to keeping class level context managers. This will cause problems in case of multiprocessing and threading, as the class context will change during the code - and might interfere with other processes. 
Let's keep the context managers `enable_grad_opt` and `disable_grad_opt` at instance level. For convenience, I will further add a setter method for `use_grad` - in case the user wants to change it during run time. 

So the user will be able to do the following. 

```python 
trainer = Trainer(model, None, config)
with trainer.enable_grad_opt(optimizer):
    trainer.fit()

with trainer.diable_grad_opt(ng_optimizer):
    trainer.fit()
```
or 

```python 
trainer = Trainer(model, None, config)
trainer.use_grad = True
trainer.optimizer = optimizer
trainer.fit()

trainer.use_grad = False
trainer.optimizer = ng_optimizer
trainer.fit()
```


For class level changes - let's keep `set_use_grad` intact. this will help reduce the amount of code needed in case a single optimization method is used throughout.

qadence/ml_tools/callbacks/writer_registry.py
mlahariya: Alright, here the proposed solution, I have added this to the docs as well. 
The user can specify three parameters for logging now. 

`root_folder`,  `create_subfolder_per_run`, `log_folder`.

- The `root_folder` and `create_subfolder_per_run` will work as expected, where logs will be saved inside a subfolder inside root folder. 

- The user can also specify a  `log_folder` (which will only work when `create_subfolder_per_run` = False (default)) - in which case the logs will be saved inside the log_folder. 

The additional `log_folder` argument might be useful in case someone wants load checkpoints and continue training from a specific run.

docs/tutorials/qml/ml_tools/trainer.md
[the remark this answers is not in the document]
mlahariya: We can this issue for future. You can simply text things out now with a simple code. 


```python
config = TrainConfig(max_iter=5)
trainer = Trainer(model, optimzer, config)
trainer.fit(train_dataloader)
```

This will only save the last checkpoint at the end of training in the default log folder - nothing else.

qadence/ml_tools/train_utils/base_trainer.py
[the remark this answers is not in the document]
debrevitatevitae: Sounds good

debrevitatevitae (approved):

mlahariya:
Hey @RolandMacDoland , @chMoussa, Please let me know if there is something else I need to revisit, else I will finish this merge.

chMoussa (approved):

qadence/ml_tools/callbacks/callbackmanager.py
```diff
@@ -0,0 +1,214 @@
...
+        elif isinstance(callback, Callback):
+            callback.on = on
+            callback.called_every = called_every
+            self.callbacks.append(callback)
+        else:
+            logger.warning(
+                f"Invalid callback type: {type(callback)}. Expected str or Callback instance."
+            )
+
+    def run_callbacks(self, trainer: Any) -> Any:
```
RolandMacDoland, line 149:
```suggestion
    def run_callbacks(self, trainer: Trainer) -> list:
```

qadence/ml_tools/callbacks/saveload.py
```diff
@@ -79,10 +79,8 @@ def write_checkpoint(
         )
         device = str(device).split(":")[0]  # in case of using several CUDA devices
     except Exception as e:
-        msg = (
-            f"Unable to identify in which device the QuantumModel is stored due to {e}."
-            "Setting device to None"
-        )
+        msg = f"""Unable to identify in which device the QuantumModel is stored due to {e}.
+                            Setting device to None"""
```
RolandMacDoland, line 83:
```suggestion
        msg = f"Unable to identify in which device the QuantumModel is stored due to {e}."
                            "Setting device to 'None'."
```

qadence/ml_tools/callbacks/saveload.py
```diff
@@ -146,8 +146,8 @@ def load_model(
             model.to(device)
 
     except Exception as e:
-        msg = f"Unable to load state dict due to {e}.\
-               No corresponding pre-trained model found. Returning the un-trained model."
+        msg = f"""Unable to load state dict due to {e}.
+                            No corresponding pre-trained model found."""
```
RolandMacDoland, line 150:
```suggestion
        msg = f"Unable to load state dict due to {e}."
                            "No corresponding pre-trained model found."
```

qadence/ml_tools/loss/loss.py
```diff
@@ -0,0 +1,87 @@
...
+    Returns:
+        Callable: The corresponding loss function.
+
+    Raises:
+        ValueError: If `loss_fn` is a string but not a supported loss function name.
+    """
+    if callable(loss_fn):
+        return loss_fn
+    elif isinstance(loss_fn, str):
+        if loss_fn == "mse":
```
RolandMacDoland, line 80: I generally refrain from hardcoding strings in favour of `StrEnums` to avoid typing errors.

RolandMacDoland (approved):
Thanks @mlahariya for this massive contribution. LGTM. Minor comments that you could ignore as you've already answered elsewhere.

qadence/ml_tools/loss/loss.py
mlahariya: I believe we will have to revisit the loss module in the ml_tools at some point. We can remove the hard coding at that point.

PR #599, opened 2024-11-05
[Refactoring] Use Readout from Pyqtorch

chMoussa:
Related to #593, we moved the readout at the backend level, so far in Pyqtorch (see [MR](https://github.com/pasqal-io/pyqtorch/pull/286) and now we do the conversion from `NoiseHandler` to be able to use it.


RolandMacDoland (approved):
Thanks @chMoussa LGTM.

Issue #603, opened 2024-11-07
[Bug] Tomography fix the rotate method


chMoussa:
### Short description

Tomography does not work in the case of X observables as the rotations should just be the H gate instead of Z * H. Currently, the returned result has a different sign.

### Steps/Code to reproduce

```
import torch
from torch import tensor
from qadence import BackendName, DiffMode
from qadence import Parameter, chain, kron, RX, RY, Z, QuantumCircuit, QuantumModel, X, Y
from qadence.measurements import Measurements
from qadence import product_state
torch.manual_seed(42)


# Define parameters for a circuit.
theta1 = Parameter("theta1", trainable=False)
theta2 = Parameter("theta2", trainable=False)
theta3 = Parameter("theta3", trainable=False)
theta4 = Parameter("theta4", trainable=False)

blocks = chain(
    kron(RX(0, theta1), RY(1, theta2)),
    kron(RX(0, theta3), RY(1, theta4)),
)

values = {
    "theta1": tensor([0.5]),
    "theta2": tensor([1.5]),
    "theta3": tensor([2.0]),
    "theta4": tensor([2.5]),
}

# Create a circuit and an observable.
circuit = QuantumCircuit(2, blocks)
observable = X(1)
# Create a model.
model = QuantumModel(
    circuit=circuit,
    observable=observable,
    backend=BackendName.PYQTORCH,
    diff_mode=DiffMode.AD,
)
```

### Would you like to work on this issue?

Yes

chMoussa:
Closes with #604

PR #604, opened 2024-11-07
[Bugfix] Fix rotations in tomography

chMoussa:
Solves #603 and also adds the extra test to reflect the change.


RolandMacDoland (approved):
Thanks @chMoussa !

Issue #607, opened 2024-11-11
[Bug] Shadows not working


chMoussa:
### Short description

Classical shadows are not implemented correctly. When asking for instance an expectation over `X(1)` that should be non-zero, it returns 0.

### What is the expected result?

A result close to the analytical expectation value.

### What is the actual result?

0

### Steps/Code to reproduce

import torch
from torch import tensor
from qadence import hamiltonian_factory, BackendName, DiffMode
from qadence import Parameter, chain, kron, RX, RY, Z, QuantumCircuit, QuantumModel, X, Y
from qadence.measurements import Measurements
from qadence import product_state
torch.manual_seed(42)

# Define parameters for a circuit.
theta1 = Parameter("theta1", trainable=False)
theta2 = Parameter("theta2", trainable=False)
theta3 = Parameter("theta3", trainable=False)
theta4 = Parameter("theta4", trainable=False)

blocks = chain(
    kron(RX(0, theta1), RY(1, theta2)),
    kron(RX(0, theta3), RY(1, theta4)),
)

values = {
    "theta1": tensor([0.5]),
    "theta2": tensor([1.5]),
    "theta3": tensor([2.0]),
    "theta4": tensor([2.5]),
}

# Create a circuit and an observable.
circuit = QuantumCircuit(2, blocks)
observable = hamiltonian_factory(2, detuning=Z)
observable = X(1)
# Create a model.
model = QuantumModel(
    circuit=circuit,
    observable=observable,
    backend=BackendName.PYQTORCH,
    diff_mode=DiffMode.AD,
)

print(model.expectation(values, state= product_state("01"), ))

shadow_options = {"accuracy": 0.1, "confidence": 0.1} 
shadow_measurement = Measurements(protocol=Measurements.SHADOW, options=shadow_options)
model.expectation(values, state= product_state("01"), measurement=shadow_measurement)

### Would you like to work on this issue?

Yes

chMoussa:
Closes with #608

PR #608, opened 2024-11-11
[Fix] Change shadow algorithm for using only samples

chMoussa:
Solves #607 
Since we use shadows for computing expectation values only, the randomized version is just about using samples and averaging them for the case the sampled pauli-string correspond to the target pauli-string.

### DIFF
--- a/qadence/measurements/shadow.py
+++ b/qadence/measurements/shadow.py
@@ -109,7 +109,7 @@ def extract_unitaries(unitary_ids: np.ndarray, n_qubits: int) -> list:
     """
     unitaries = nested_operator_indexing(unitary_ids)
     if n_qubits > 1:
-        unitaries = [kron(*l) for l in unitaries]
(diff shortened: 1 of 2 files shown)


qadence/measurements/shadow.py
```diff
@@ -119,123 +129,85 @@ def classical_shadow(
...
     K: int,
-    shadow: list,
+    unitary_shadow_ids: np.ndarray,
+    shadow_samples: Tensor,
     observable: AbstractBlock,
-    endianness: Endianness = Endianness.BIG,
 ) -> Tensor:
     """
-    Return estimators (traces of observable times mean density matrix).
+    Return estimators (aka traces calculated by matching the sampled unitaries to the.
```
RolandMacDoland, line 174: The docstring needs formatting.

RolandMacDoland (approved):
Thanks for this ! Minor comment and LGTM !

PR #610, opened 2024-11-12
[Feature] Setting digital noise in addition to readout noise in Pyq backend functions sample and expectation

chMoussa:
Complementing #591 and #599 where digital and readout noise can be set in the converted circuit from `sample` and `expectation`, without the user calling `set_noise` beforehand.

### DIFF
--- a/qadence/backends/pyqtorch/backend.py
+++ b/qadence/backends/pyqtorch/backend.py
@@ -109,7 +109,8 @@ class Backend(BackendInterface):
         if len(passes) > 0:
             circuit = transpile(*passes)(circuit)
         # setting noise on blocks
-        set_noise(circuit, self.config.noise)
+        if self.config.noise:
+            set_noise(circuit, self.config.noise)
 
         ops = convert_block(circuit.block, n_qubits=circuit.n_qubits, config=self.config)
         readout_noise = (
--- a/qadence/backends/pyqtorch/backend.py
+++ b/qadence/backends/pyqtorch/backend.py
@@ -44,7 +44,8 @@ def converted_circuit_with_noise(
 ) -> ConvertedCircuit:
     """For backend functions, get a ConvertedCircuit with noise.
 
-    Note that if config.noise is not None, we try to append
+    Note that if config.noise is not None, we try to append to noise and
+    this may raise an error.
 
     Args:
         circuit (ConvertedCircuit): Original ConvertedCircuit.
@@ -62,13 +63,15 @@ def converted_circuit_with_noise(
             except ValueError as e:
                 raise ValueError(f"Cannot append provided noise with the config noise. Error: {e}")
         new_convcirc = ConvertedCircuit(circuit.native, circuit.abstract, circuit.original)
-        set_noise(new_convcirc.abstract, noise)
+        set_noise(new_convcirc.abstract, noise_combination)
+
+        # the config should not add the noise
         ops = convert_block(
             new_convcirc.abstract.block, n_qubits=new_convcirc.abstract.n_qubits, config=config
         )
-        readout = None
+        readout = new_convcirc.native.readout_noise
         if new_convcirc.native.readout_noise is None:
-            readout = convert_readout_noise(new_convcirc.abstract.n_qubits, noise)
+            readout = convert_readout_noise(new_convcirc.abstract.n_qubits, noise_combination)
         native = pyq.QuantumCircuit(
             new_convcirc.abstract.n_qubits,
             ops,
@@ -105,6 +108,8 @@ class Backend(BackendInterface):
         original_circ = circuit
         if len(passes) > 0:
             circuit = transpile(*passes)(circuit)
+        # setting noise on blocks
+        set_noise(circuit, self.config.noise)
 
         ops = convert_block(circuit.block, n_qubits=circuit.n_qubits, config=self.config)
         readout_noise = (
@@ -130,8 +135,8 @@ class Backend(BackendInterface):
             scale_primitive_blocks_only,
         ]
         block = transpile(*transpilations)(observable)  # type: ignore[call-overload]
-        # prevent setting noise on observable
-        operations = convert_block(block, n_qubits, self.config, False)
+        # we do not set noise on the observable blocks
+        operations = convert_block(block, n_qubits, self.config)
         native = pyq.Observable(operations=operations)
         return ConvertedObservable(native=native, abstract=block, original=observable)
 
--- a/qadence/backends/pyqtorch/backend.py
+++ b/qadence/backends/pyqtorch/backend.py
@@ -28,6 +28,7 @@ from qadence.transpile import (
     flatten,
     invert_endianness,
     scale_primitive_blocks_only,
+    set_noise,
     transpile,
 )
 from qadence.types import BackendName, Endianness, Engine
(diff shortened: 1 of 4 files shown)


qadence/backends/pyqtorch/backend.py
```diff
@@ -38,6 +39,51 @@
...
+    Returns:
+        ConvertedCircuit: A new ConvertedCircuit with noise.
+    """
+
+    if noise:
+        noise_combination = noise
+        if config.noise:
+            try:
+                noise_combination.append(config.noise)
+            except ValueError as e:
```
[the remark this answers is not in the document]
chMoussa: For instance if this leads to adding up two Readout Noise. Any of the exceptions raised by `append`

qadence/noise/protocols.py
```diff
@@ -149,13 +149,17 @@ def list(cls) -> list:
...
+            is_protocol = [p == protocol for p in self.protocol]
+        else:
+            is_protocol = [isinstance(p, protocol) for p in self.protocol]  # type: ignore[arg-type]
         return (
             NoiseHandler(
                 list(compress(self.protocol, is_protocol)),
                 list(compress(self.options, is_protocol)),
             )
-            if len(is_protocol) > 0
+            if len(is_protocol) > 0 and sum(is_protocol) > 0
```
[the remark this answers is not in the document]
chMoussa: It's to avoid the case where the result of compress is the empty list which throws an error

tests/qadence/test_noise/test_digital_noise.py
```diff
@@ -92,6 +92,34 @@ def test_run_digital(noisy_config: NoiseProtocol | list[NoiseProtocol]) -> None:
...
 
 
+@pytest.mark.parametrize(
+    "noisy_config",
+    [
+        NoiseProtocol.DIGITAL.BITFLIP,
+        [NoiseProtocol.DIGITAL.BITFLIP, NoiseProtocol.DIGITAL.PHASEFLIP],
+    ],
+)
+def test_expectation_digital_noise(noisy_config: NoiseProtocol | list[NoiseProtocol]) -> None:
```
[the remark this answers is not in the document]
chMoussa: Sure

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -175,8 +175,23 @@ def fn(x: str | ConcretizedCallable, y: str | ConcretizedCallable) -> Callable:
...
+    Args:
+        block (AbstractBlock): Block to convert.
+        n_qubits (int, optional): Number of qubits. Defaults to None.
+        config (Configuration, optional): Backend configuration instance. Defaults to None.
+
+    Raises:
+        NotImplementedError: For non supported blocks.
+
+    Returns:
+        Sequence[Module | Tensor | str | sympy.Expr]: Lis of native operations.
```
mlahariya, line 193: List*

mlahariya (commented):
Thank @chMoussa, nice work. Looks alright to me. Just a minor typo.

qadence/backends/pyqtorch/backend.py
```diff
@@ -38,6 +39,51 @@
 logger = getLogger(__name__)
 
 
+def converted_circuit_with_noise(
```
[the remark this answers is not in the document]
jpmoutinho: I am a little unclear on why we need this function. Is because of the logic on the functions in `qadence.execution`? E.g. the fact that they don't instantiate a `QuantumModel` but instead a backend directly.

qadence/backends/pyqtorch/backend.py
```diff
@@ -87,6 +136,7 @@ def observable(self, observable: AbstractBlock, n_qubits: int) -> ConvertedObser
             scale_primitive_blocks_only,
         ]
         block = transpile(*transpilations)(observable)  # type: ignore[call-overload]
+        # we do not set noise on the observable blocks
```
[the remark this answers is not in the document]
jpmoutinho: I would just remove this comment, I don't think that's a functionality people would consider right?

qadence/backends/pyqtorch/backend.py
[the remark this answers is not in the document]
chMoussa: So it is because we can pass the noise in either the config, or as an argument of expectation or sample. For these last two, the input converted circuit may not have the noise set.

qadence/backends/pyqtorch/backend.py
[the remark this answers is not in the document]
chMoussa: Eventually, I removed this check. I assume the noise passed should also be the same from config or replaces it completely.

qadence/backends/pyqtorch/backend.py
```diff
@@ -38,6 +39,51 @@
...
+    """
+
+    if noise:
+        noise_combination = noise
+        if config.noise:
+            try:
+                noise_combination.append(config.noise)
+            except ValueError as e:
+                raise ValueError(f"Cannot append provided noise with the config noise. Error: {e}")
+        new_convcirc = ConvertedCircuit(circuit.native, circuit.abstract, circuit.original)
```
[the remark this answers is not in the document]
chMoussa: @RolandMacDoland I decided to add the ConvertedCircuit class in set_noise. I also separated in the backend all the steps into dedicated functions. I would avoid though putting everything into set_noise as these instructions are specific to the pyq backend using its conversion functions.

qadence/noise/protocols.py
```diff
@@ -149,15 +149,18 @@ def list(cls) -> list:
         return list(filter(lambda el: not el.startswith("__"), dir(cls)))
 
     def filter(self, protocol: NoiseEnum | str) -> NoiseHandler | None:
-        is_protocol: list = [p == protocol or isinstance(p, protocol) for p in self.protocol]  # type: ignore[arg-type]
-        return (
-            NoiseHandler(
+        is_protocol: list = list()
```
RolandMacDoland, line 152: `is_<smtg>` generally conveys the idea of a predicate. Here, I'd use `protocols` instead.

qadence/noise/protocols.py
```diff
@@ -149,15 +149,18 @@ def list(cls) -> list:
...
+        if protocol == NoiseProtocol.READOUT:
+            is_protocol = [p == protocol for p in self.protocol]
+        else:
+            is_protocol = [isinstance(p, protocol) for p in self.protocol]  # type: ignore[arg-type]
```
RolandMacDoland, line 156: Isn't possible to use the `filter` functionality ?

qadence/noise/protocols.py
```diff
@@ -149,15 +149,18 @@ def list(cls) -> list:
...
+        # if we have at least a match
+        if sum(is_protocol) > 0:
```
RolandMacDoland, line 158:
```suggestion
        if protocols:
```

qadence/transpile/noise.py
```diff
@@ -23,24 +24,26 @@ def _set_noise(
...
         target_class: optional class to selectively add noise to.
     """
-    is_circuit_input = isinstance(circuit, QuantumCircuit)
-
-    input_block: AbstractBlock = circuit.block if is_circuit_input else circuit  # type: ignore
+    to_convert = circuit.abstract if isinstance(circuit, ConvertedCircuit) else circuit
+    is_circuit_input = isinstance(to_convert, QuantumCircuit)
 
-    output_block = apply_fn_to_blocks(input_block, _set_noise, noise, target_class)
+    input_block: AbstractBlock = to_convert.block if is_circuit_input else to_convert  # type: ignore
```
RolandMacDoland, line 46: Uff...I feels quite a complicated bit of logic that might be simplified ? I don't have a straight answer off the top of my head but it really feels that you only need the `.block` when this is a `QuantumCircuit` and the `circuit` otherwise.

RolandMacDoland (commented):
Thanks @chMoussa looks nice but I think there are bits and pieces that might be simplified. Happy to approve once addressed.

qadence/backends/pyqtorch/backend.py
```diff
@@ -38,6 +39,46 @@
 logger = getLogger(__name__)
 
 
+def set_noise_abstract_to_native(circuit: ConvertedCircuit, config: Configuration) -> None:
```
RolandMacDoland, line 42: I don't quite get why this is here while `set_noise` is in `transpile`.

qadence/noise/protocols.py
chMoussa: Replaced by `if True in`

qadence/noise/protocols.py
chMoussa: Updated with `protocol_matches`

qadence/noise/protocols.py
chMoussa: Well `filter` would return the objects, but I need to know also the corresponding `options` for filtered protocols. Hence why I keep the list of booleans

qadence/backends/pyqtorch/backend.py
chMoussa: It's because I return a pyq object, and I'd avoid showing outside pyq objects outside of backends.

RolandMacDoland (approved):
Thanks @chMoussa. LGTM. Minor comments.

qadence/noise/protocols.py
```diff
@@ -149,15 +149,19 @@ def list(cls) -> list:
...
-        is_protocol: list = [p == protocol or isinstance(p, protocol) for p in self.protocol]  # type: ignore[arg-type]
-        return (
-            NoiseHandler(
-                list(compress(self.protocol, is_protocol)),
-                list(compress(self.options, is_protocol)),
+        protocol_matches: list = list()
+        if protocol == NoiseProtocol.READOUT:
+            protocol_matches = [p == protocol for p in self.protocol]
+        else:
+            protocol_matches = [isinstance(p, protocol) for p in self.protocol]  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
chMoussa: I think it's due to using StrEnum.

PR #612, opened 2024-11-14
[Bugfix] Fix noisy simulation pulser and test

chMoussa:
Related to #583. Now we return single-valued expectations.

### DIFF
--- a/qadence/backends/pulser/backend.py
+++ b/qadence/backends/pulser/backend.py
@@ -351,27 +351,14 @@ class Backend(BackendInterface):
                     noise=noise,
                 )
                 support = sorted(list(circuit.abstract.register.support))
-                # TODO: There should be a better check for batched density matrices.
-                if dms.size()[0] > 1:
-                    res_list = [
-                        [
-                            obs.native(
-                                dm.squeeze(), param_values, qubit_support=support, noise=noise
-                            )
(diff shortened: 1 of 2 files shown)


qadence/backends/pulser/backend.py
```diff
@@ -351,27 +351,14 @@ def expectation(
...
-                    )
-
-                else:
-                    res_list = [
-                        obs.native(dms, param_values, qubit_support=support) for obs in observable
+                res_list = [
+                    [
+                        obs.native(dm.squeeze(), param_values, qubit_support=support, noise=noise)
+                        for dm in dms
                     ]
```
jpmoutinho, line 358: Isn't this removing some functionality related to batching?

chMoussa: From the previous code, it did not seem batching was allowed since we were looping over `dms` and `observable`. But let me check this by in the appropriate test

chMoussa: We have two tests in tests/backends/pulser_basic/test_quantum_pulser where we test a non-batched case with one noise_probs value called `test_noisy_simulations`, and another case where we test several values called `test_batched_noisy_simulations`.

jpmoutinho (approved):

PR #616, opened 2024-11-19
[Bug] Fix GPSR differentiation mode

vytautas-a:
Currently GPSR differentiation mode with `pyqtorch` backend is not used even when it is selected. Also `h_norm` factor is used when parsing analog gates that is not correct. This PR fixes these issues.

### DIFF
--- a/qadence/backends/gpsr.py
+++ b/qadence/backends/gpsr.py
@@ -36,7 +36,11 @@ def general_psr(spectrum: Tensor, n_eqs: int | None = None, shift_prefac: float 
     sorted_unique_spectral_gaps = torch.tensor(list(sorted_unique_spectral_gaps)[:n_eqs])
 
     if n_eqs == 1:
-        return single_gap_psr
+        return partial(
+            single_gap_psr,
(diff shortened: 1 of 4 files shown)


qadence/backends/gpsr.py
```diff
@@ -110,7 +114,9 @@ def multi_gap_psr(
     batch_size = max(t.size(0) for t in param_dict.values())
 
     # get shift values
-    shifts = shift_prefac * torch.linspace(PI / 2 - PI / 5, PI / 2 + PI / 5, n_eqs)
+    shifts = shift_prefac * torch.linspace(
+        PI / 2 - PI / 4, PI / 2 + PI / 5, n_eqs
```
chMoussa, line 118: Why does it become PI / 4 instead of PI / 5 @vytautas-a ?

vytautas-a: Short answer - it allows to pass all the differentiation tests. Long answer - I suspect that due to some symmetries in matrices computed during GPSR calculations, results with some generators are not correct. Breaking this symmetry of shifts around `pi/2` helps in this regard.

Issue #369, opened 2024-03-19
Allow readout errors to be applied on the backend for differentiability


rajaiitp:
Requesting for readout errors to be applied without sampling 

tensordot can be used to apply readout errors if pytorch is computing the probability vector without directly using statevector for computing expectation values (not sure if it makes it slower, but is only usable for diagonal observables)

This would help with training models faster using AD while keeping the effects of readout errors. Something which the usecase people would require sooner or later for robustness studies.

Roland-djee:
Hey @rajaiitp thanks for this. Could you please clarify details here ?

rajaiitp:
From what i see, the add_noise method in readout.py allows for noise to implemented at the level of bitstrings rather than probability vectors. This means that sampling is happen before readout errors applied.

Adjoint differentiation isnt compatible with sampling, i understand. The goto method is thus PSR if one wants to implement readout errors. It would be way convenient if we can bypass sampling and apply readout errors directly on the probability vector instead.

I like to think that the best way to do is generate transformed probability vectors affected by readout matrix directly at the level of backend (instead of exact non noisy state vectors), so that diagonal observables can be measured with it

Roland-djee:
@gvelikova Could you please have a go at what this entails exactly ? That'd be helpful. Thanks.

Roland-djee:
@rajaiitp @gvelikova It would be great to start on this.

rajaiitp:
Missed it, I will start working on this soon

mlahariya:
Closing as not relevant anymore.

PR #620, opened 2024-11-25
[Feature] Correlated readout

### LINKED ISSUE #383: [Feature, Refactoring] Improve readout noise model
* allow for correlated noise following https://arxiv.org/abs/2006.14044

Other links:
https://journals.aps.org/pra/abstract/10.1103/PhysRevA.103.042605
An implementation is available at https://qiskit-community.github.io/qiskit-experiments/stubs/qiskit_experiments.library.characterization.CorrelatedReadoutError.html

chMoussa:
Closes #383 and #369. Two readout noise protocols are available from Pyqtorch: correlated readout and independent readout (that was in qadence originally).

- [x] Change types
- [x] Modify noiseHandler accordingly
- [x] Add tests
- [x] Modify docs

### DIFF
--- a/qadence/noise/protocols.py
+++ b/qadence/noise/protocols.py
@@ -197,7 +197,7 @@ class NoiseHandler:
         self.append(NoiseHandler(NoiseProtocol.ANALOG.DEPHASING, *args, **kwargs))
         return self
 
-    def independentreadout(self, *args: Any, **kwargs: Any) -> NoiseHandler:
+    def independent_readout(self, *args: Any, **kwargs: Any) -> NoiseHandler:
         self.append(NoiseHandler(NoiseProtocol.READOUT.INDEPENDENTREADOUT, *args, **kwargs))
         return self
 
--- a/qadence/noise/protocols.py
+++ b/qadence/noise/protocols.py
@@ -149,16 +149,22 @@ class NoiseHandler:
         return list(filter(lambda el: not el.startswith("__"), dir(cls)))
 
     def filter(self, protocol: NoiseEnum | str) -> NoiseHandler | None:
-        is_protocol: list = list()
         if protocol == NoiseProtocol.READOUT:
-            is_protocol = [p == protocol for p in self.protocol]
+
+            def filter_fn(p: NoiseEnum | str) -> bool:
+                return p == protocol
+
         else:
-            is_protocol = [isinstance(p, protocol) for p in self.protocol]  # type: ignore[arg-type]
+
+            def filter_fn(p: NoiseEnum | str) -> bool:
+                return isinstance(p, protocol)  # type: ignore[arg-type]
+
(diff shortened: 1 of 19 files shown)


qadence/types.py
```diff
@@ -470,16 +470,25 @@ class AnalogNoise(StrEnum):
     DEPHASING = "Dephasing"
 
 
+class ReadoutNoise(StrEnum):
+    """Type of readout protocol."""
+
+    INDEPENDENTREADOUT = "IndependentReadout"
```
mlahariya, line 476: It becomes a little verbose to specify READOUT.INDEPENDENTREADOUT and READOUT.CORRELATEDREADOUT. 

Maybe remove the readout at the end, so that they can be specified like READOUT.INDEPENDENT and READOUT.CORRELATED.

qadence/noise/protocols.py
```diff
@@ -201,6 +197,10 @@ def dephasing(self, *args: Any, **kwargs: Any) -> NoiseHandler:
         self.append(NoiseHandler(NoiseProtocol.ANALOG.DEPHASING, *args, **kwargs))
         return self
 
-    def readout(self, *args: Any, **kwargs: Any) -> NoiseHandler:
-        self.append(NoiseHandler(NoiseProtocol.READOUT, *args, **kwargs))
+    def independent_readout(self, *args: Any, **kwargs: Any) -> NoiseHandler:
```
mlahariya, line 200: Should we have a single default readout for the handler? I am not sure if this will be useful, but in case the user does not specify the type of readout, we can default it to Independent readout?

qadence/noise/protocols.py
```diff
@@ -201,6 +197,10 @@ def dephasing(self, *args: Any, **kwargs: Any) -> NoiseHandler:
...
+        self.append(NoiseHandler(NoiseProtocol.READOUT.INDEPENDENTREADOUT, *args, **kwargs))
+        return self
+
+    def correlated_readout(self, *args: Any, **kwargs: Any) -> NoiseHandler:
```
mlahariya, line 204: Maybe consider renaming these functions to readout_correlated, and readout_indipendent so that they align with how readout is specified using the protocol in the handler?

mlahariya (commented):
Hey @chMoussa , Great work. I have added a few comments related to nomenclature and having a default readout - in case the user does not (want to?) specify the type of readout.

chMoussa:
> Hey @chMoussa , Great work. I have added a few comments related to nomenclature and having a default readout - in case the user does not (want to?) specify the type of readout. 

Indeed you're right. I'll change this quickly.

qadence/noise/protocols.py
chMoussa: Hm these methods are mostly helpers. So I would let them as if

mlahariya (approved):

PR #621, opened 2024-11-25
[Feature] Time-dependent Hamevo with noise operators

### LINKED ISSUE #622: [Feature] Add noise_operators in Hamevo as in Pyqtorch
### Describe the feature

Perform noisy simulations with Hamevo as it is done in Pyqtorch

_No response_

### Additional context

_No response_

Yes

chMoussa:
Closes #622 . Adding the `noise_operators` argument to `Hamevo` as in Pyqtorch.

TODOs:

- [x] Add a `noise_operators` argument to Hamevo
- [x] Set the conversion to Pyqtorch appropriately (solver, steps, and noise_operators)
- [x] Add/Fix tests
- [x] Add docs

Note that one may need to adapt the NoiseHandler class later to set the `noise_operators` accordingly to the type of analog noise. However, this may require harmonising all noises in all backends. Thus this is left for future work.


docs/content/time_dependent.md
```diff
@@ -43,3 +45,36 @@ print(out_state)
...
+# Define block parameters
+t = TimeParameter("t")
+omega_param = FeatureParameter("omega")
+
+# Arbitrarily compose a time-dependent generator
+generator_td = omega_param * (t * X(0) + t**2 * Y(1))
+
+# Create parameterized HamEvo block
+noise_operators = [torch.eye(4, dtype=torch.complex128)]
+hamevo = HamEvo(generator_td, t, noise_operators = noise_operators)
```
jpmoutinho, line 71: Hmm I think the interface should be to pass Qadence blocks, no?

E.g. `noise_operators = [I(0) @ I(1))]`. What are the operators users will want to pass here, various combinations of Paulis?

jpmoutinho (commented):
Thanks @chMoussa. I think it would be more consistent to stick to qadence blocks when it comes to interfaces asking for operators.

docs/content/time_dependent.md
```diff
@@ -43,3 +45,36 @@ print(out_state)
...
+# Simulation parameters
+ode_solver = SolverType.DP5_ME  # time-dependent Schrodinger equation solver method
+n_steps_hevo = 500  # integration time steps used by solver
+
+# Define block parameters
+t = TimeParameter("t")
+omega_param = FeatureParameter("omega")
+
+# Arbitrarily compose a time-dependent generator
+generator_td = omega_param * (t * X(0) + t**2 * Y(1))
```
RolandMacDoland, line 67:
```suggestion
td_generator = omega_param * (t * X(0) + t**2 * Y(1))
```

docs/content/time_dependent.md
```diff
@@ -43,3 +45,36 @@ print(out_state)
...
+
+# Create parameterized HamEvo block
+noise_operators = [torch.eye(4, dtype=torch.complex128)]
```
RolandMacDoland, line 70: Do I understand correctly that noisy operators are simply identities ?

tests/backends/pyq/test_time_dependent_generator.py
```diff
@@ -25,24 +25,79 @@ def test_time_dependent_generator(
 ) -> None:
     n_steps = 500
 
+    psi_0 = qutip.basis(4, 0)
+    psi0_torch = (torch.tensor(psi_0.full()).to(torch.complex128)).reshape(1, -1)
```
RolandMacDoland, line 29: Why do you have to use a qutip object here ? Ah I see. I'd suggest to add them as fixtures.

docs/content/time_dependent.md
RolandMacDoland, line 71: Does it check dimensionalities match ?

RolandMacDoland (commented):
Hey @chMoussa thanks for this ! Unsure about how it works though. I thought you had to specify operators but instead you seem to be passing a list of identities ?

docs/content/time_dependent.md
chMoussa: Apparently, there are no restrictions on these noise matrices besides having the correct shape.

chMoussa: I can of course allow for a list of blocks that would require a call `block_to_tensor` and removing the batchsize.

chMoussa: I think not but I probably can add this in the init

chMoussa: Added

tests/backends/pyq/test_time_dependent_generator.py
chMoussa: Eventually I could remove the initial state `psi0_torch` as it corresponds to the state=None.

chMoussa: Why adding qutip object as fixture?

docs/content/time_dependent.md
jpmoutinho: My main question would be in practice what do users typically use. If it's mainly pauli strings, then it would be much more convenient to do it with Qadence blocks. If they want to use arbitrary matrices, it could also be a use for the `MatrixBlock`, which has not been used that much.

jpmoutinho: Btw, it should allow projectors as well. But I guess you will make it work with any block right?

chMoussa: @jpmoutinho Yes, btw I already modified it to work with a list of `AbstractBlock`.

chMoussa:
> My main question would be in practice what do users typically use. If it's mainly pauli strings, then it would be much more convenient to do it with Qadence blocks. If they want to use arbitrary matrices, it could also be a use for the `MatrixBlock`, which has not been used that much.

@jpmoutinho @RolandMacDoland I modified eventually the code to work with a list of blocks as noise operators. We won't support certain cases such as parametric blocks but I left a warning note in docs about using `MatrixBlock` instead. This is reflected in the tests now. Let me know what you think of it.

docs/content/time_dependent.md
```diff
@@ -43,3 +45,39 @@ print(out_state)
...
+
+config = {"ode_solver": ode_solver, "n_steps_hevo": n_steps_hevo}
+
+out_state = run(hamevo, values = values, configuration = config)
+
+print(out_state)
+```
+
+!!! warning "Noise operators definition"
+    Note it is not possible to define `noise_operators` with parametric operators. If you want to do so, we recommend obtaining the tensors via run and set `noise_operators` using `MatrixBlock`. Also, `noise_operators` should have the same or a subset of the qubit support of the `HamEvo` instance.
```
jpmoutinho, line 83: Two questions:
- I don't understand the suggestion to "obtain the tensors via run and set `noise_operators` using `MatrixBlock`" -- why is that similar to using parametric operators?
- Regarding the qubit support, what is the current behaviour regarding e.g. passing `X(0)` as a jump operator? Will it only affect qubit 0, or all qubits? 

In Pulser, only single-qubit jump operators are possible, and currently the behaviour is set to applying that same operator to all the qubits. To get that in Qadence, would one need to pass `[X(i) for i in range(n_qubits)]` ?

qadence/backends/pyqtorch/convert_ops.py
```diff
@@ -245,13 +245,26 @@ def convert_block(
...
 
+        # convert noise operators here
+        noise_operators: list = [
+            convert_block(noise_block, config=config)[0] for noise_block in block.noise_operators
+        ]
+        if len(noise_operators) > 0:
+            # squeeze batch size for noise operators
+            noise_operators = [
+                pyq_op.tensor(full_support=qubit_support).squeeze(-1) for pyq_op in noise_operators
+            ]
```
jpmoutinho, line 256: Do I understand correctly that this is blowing up the qubit support of each noise operator to the full qubit support of the evolution?

If yes, is this a current limitation of the `mesolve` ?

jpmoutinho (commented):
Thanks @chMoussa, looking good. I think my main question now is about how general / restrictive should the noise model be. In Pulser only single-qubit jump operators are available. If we don't have that restriction, is that ok?

docs/content/time_dependent.md
chMoussa: @jpmoutinho 

- My bad, I meant using `block_to_tensor` with the parameter values on the operator to convert it to a matrix that should be passed to a MatrixBlock.
- Similarly to applying an X gate to qubit 0, which will be converted to a matrix with filled identity to be applied with mesolve.

chMoussa: For Pulser, I guess passing `[X(i) for i in range(n_qubits)]` would be the same, but seems restrictive no?

qadence/backends/pyqtorch/convert_ops.py
chMoussa: @jpmoutinho Yes, it is a current limitation of mesolve.

docs/content/time_dependent.md
jpmoutinho: What I mean is that in Pulser only single-qubit jump operators are possible AND they automatically apply to all qubits. So if you pass X, you are essentially passing X(0), X(1), X(2)...

I just wanted to clarify if these restrictions are NOT in place in Qadence / PyQ currently.
- So if you pass X(0), it literally means that only qubit 0 will have a noise effect. 
- If someone wants to run on PyQ and produce the same effect that they would get in Pulser, they do have to write each individual jump operator affecting each individual qubit, e.g. `[X(i) for i in range(n_qubits)]`.

Correct?

chMoussa: I think that is correct, they need to specify `[X(i) for i in range(n_qubits)]`.

tests/backends/pyq/test_time_dependent_generator.py
chMoussa: Ok eventually skipping this as tests changed

docs/content/time_dependent.md
jpmoutinho: Can you change the example so it shows the syntax for applying the same operator to each qubit instead? (with an explicit comment that you do need to specify each operator for each qubit).

jpmoutinho (approved):
Thanks, final comment left on the docs but otherwise lgtm.

Issue #535, opened 2024-08-15
[Docs] Documentation improvements


jpmoutinho:
Here we list a few things that need to be made more explicit in the docs:

- The usage of `set_trainable` to change between feature and variational parameters in blocks. Maybe can be added in the [Parameters](https://pasqal-io.github.io/qadence/latest/content/parameters/) section.
- A few things on the usage of [QuantumModel](https://pasqal-io.github.io/qadence/latest/content/quantummodels/):
  - How to reset quantum model parameter values to specific value (`model.reset_vparams(...)`)
  - How to get derivatives with respect to trainable parameters (e.g., `autograd.grad(loss, qnn.parameters())`).
  - How to obtain the parameter dict at a given training epoch (`model._params`)
- The whole "model config" options need documentation, e.g., the quantum circuit dropout.

mlahariya:
If we would like to work on this issue further - can we reassign it to @sungwoo-pasqal.

@RolandMacDoland , @jpmoutinho - thoughts?

PR #623, opened 2024-11-26
[Feature] Add quantum dropout 

chMoussa:
Solves #530

### DIFF
--- a/qadence/backends/pyqtorch/config.py
+++ b/qadence/backends/pyqtorch/config.py
@@ -4,8 +4,6 @@ from dataclasses import dataclass
 from logging import getLogger
 from typing import Callable
 
-from pyqtorch.utils import DropoutMode, SolverType
-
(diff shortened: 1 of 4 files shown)


qadence/backends/pyqtorch/config.py
```diff
@@ -4,7 +4,7 @@
 from logging import getLogger
 from typing import Callable
 
-from pyqtorch.utils import SolverType
+from pyqtorch.utils import DropoutMode, SolverType
```
jpmoutinho, line 7: You need to expose the `DropoutMode` in Qadence by adding it to the `types` as well.

jpmoutinho:
I added a line here to keep track of documenting this: https://github.com/pasqal-io/qadence/issues/535

jpmoutinho (approved):
Thanks, lgtm :)

PR #624, opened 2024-11-27
[Refac] Refactor the ansatz constructors

### LINKED ISSUE #617: [Feature] Implement sDAQC strategy for Alt Ansatz
### Describe the feature

Following on from #565 add sDAQC strategy to Alt Ansatz constructor.

My plan is to simply pass the support based on the block that is defined using `m` and `n`.

It should be implemented for parity between the ansatze

### Additional context

_No response_

Yes

smitchaudhary:
Changing the scope of the PR:

Now, the PR only refactors the code for the different ansatze into appropriate files and has consistent interface and naming for them.

~Fixes #617~

~Following on from #565 add sDAQC strategy to Alt Ansatz constructor.~


smitchaudhary:
Not fully ready yet. only refactored some code, haven't added the sDA version yet.

Summary so far:

1. The 3 ansatze now live in 3 separate files instead of 2 of them in a singular big ansatze.py and the other in iia.py
2. Created an alias `iia` for `identity_initialized_ansatz` so that it can be used in the same as `hea` and `alt` are available without renaming the original function to not break anything.

jpmoutinho:
Hey @smitchaudhary, is this still something you want to pursue? Let me know if you want to discuss it.

smitchaudhary:
Hey @jpmoutinho 

I originally started working on implementing an alternating layer ansatz with SDAQC strategy, however, that would require us to redefine how analog blocks are defined currently. Namely, we would need the ability to specify `support` for them, the way we currently specify for digital blocks.

I am inclined to not go forward with this for the time being? One reason is that I understand qadence is under "relative" freeze and we are not looking to make too many changes. Besides, in terms of prioritising, I think there might not be a pressing need for this at this point. If some users want alt ansatz with SDAQC, we can maybe have a look at it again?

I am keeping the refactor here, let me know if we still want to do that, or just close the PR in favour of the existing state.

jpmoutinho (approved):
Hey @smitchaudhary, not much to say, I think the refactor is worth merging for general code tidiness. Thanks for thanking care of that!

PR #625, opened 2024-12-03
[Docs] Update docs on OS and WLS2 use.

### LINKED ISSUE #160: [Documentation] Installation instructions: mention OS, tips for Windows/Linux/MacOS ?
After discussing with some of the recent users of qadence, I realized we are not yet mentioning any specifics on installation. It is assumed that the user has Linux or MacOS running and then everything will be fine (I think). For GraphViz there are specific instructions for Ubuntu and MacOS.

Suggestion: Let's add an explanation in the README/documentation, which states that we have tested and support Linux and MacOS currently, and that we recommend either WSL2 or dual-booting for those running Windows. We could add a link to a trusted/clear WSL2 installation guide or a super short intro to it. What do you think?

RolandMacDoland:
Closes #160.

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -1,6 +1,6 @@
 ## Requirements
 
-Qadence is fully tested on Linux/MacOS operating systems. For Widows users, we recommend using [WSL2](https://learn.microsoft.com/en-us/windows/wsl/about) to install a Linux distribution of choice.
+Qadence is fully tested on Linux/MacOS operating systems. For Windows users, we recommend using [WSL2](https://learn.microsoft.com/en-us/windows/wsl/about) to install a Linux distribution of choice.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


docs/getting_started/installation.md
```diff
@@ -1,3 +1,9 @@
+## Requirements
+
+Qadence is fully tested on Linux/MacOS operating systems. For Widows users, we recommend using [WSL2](https://learn.microsoft.com/en-us/windows/wsl/about) to install a Linux distribution of choice.
```
chMoussa, line 3: `Windows` instead of `Widows`

mlahariya (approved):
Looks good. 
(Just the minor spell check for windows that Charles mentioned)

docs/getting_started/installation.md
RolandMacDoland: Done.

PR #627, opened 2024-12-04
[bug] Resume dictdataloader support for Trainer

mlahariya:


qadence/ml_tools/train_utils/base_trainer.py
```diff
@@ -330,25 +330,27 @@ def _compute_num_batches(self, dataloader: DataLoader) -> int:
...
         Validates the type of the DataLoader and raises errors for unsupported types.
 
         Args:
-            dataloader (DataLoader): The DataLoader to validate.
+            dataloader (DataLoader | DictDataLoader): The DataLoader to validate.
             dataloader_type (str): The type of DataLoader ("train", "val", or "test").
         """
         if dataloader is not None:
-            if not isinstance(dataloader, DataLoader):
+            if not (isinstance(dataloader, DataLoader) or isinstance(dataloader, DictDataLoader)):
```
chMoussa, line 344: We can just do a `not isinstance(dataloader, (DataLoader, DictDataLoader))`

qadence/ml_tools/train_utils/base_trainer.py
```diff
@@ -330,25 +330,27 @@ def _compute_num_batches(self, dataloader: DataLoader) -> int:
...
                 raise NotImplementedError(
                     f"Unsupported dataloader type: {type(dataloader)}."
                     "The dataloader must be an instance of DataLoader."
                 )
         if dataloader_type == "val" and self.config.val_every > 0:
-            if not isinstance(dataloader, DataLoader):
+            if not (isinstance(dataloader, DataLoader) or isinstance(dataloader, DictDataLoader)):
```
chMoussa, line 350: Same: `not isinstance(dataloader, (DataLoader, DictDataLoader))`

qadence/ml_tools/train_utils/base_trainer.py
mlahariya: Updated.

qadence/ml_tools/train_utils/base_trainer.py
mlahariya: Updated.

chMoussa (approved):
LGTM, do you think there is a need to add more docs on it?

mlahariya:
> LGTM, do you think there is a need to add more docs on it?

In the API documentation the dictdataloader support is already documented. 
We should definately add an example of using the dictdataloader with trainer - but I am not sure which example would be most relevant. So for now we can skip it, and create a issue for it to be added later on.

qadence/ml_tools/callbacks/writer_registry.py
```diff
@@ -274,6 +273,8 @@ def open(self, config: TrainConfig, iteration: int | None = None) -> ModuleType
         Returns:
             mlflow: The MLflow module instance.
         """
+        import mlflow
```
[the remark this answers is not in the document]
mlahariya: This is to ensure that mlflow works accurately as an optional dependency. (Which is the case in the pyproject.toml.)

If the user installs qadence - mlflow will not be installed (other than in a hatch dev env).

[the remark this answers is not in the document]
RolandMacDoland: Yeah, I would assume so. There seems to be low use for mlflow to justify it being a dep on its own.

Issue #145, opened 2023-10-27
[Realistic-sims, Benchmarking, Measurements] Improve performance for classical shadows method


madagra:
Classical shadows algorithm performance can be greatly improved by doing appropriate parallelization of the loop for shadow computation and in other places within the code (see comments throughout the code).

**UPDATE 06/11/23**: let's use this issue for testing also the performance of the bitstring corruption routines which might be quite inefficient. Up to how many qubits can we go?

Roland-djee:
Part of Ethan internship ?

chMoussa:
@mlahariya This will be revisited in protocols. I also have #608 open that fixes an issue with the current shadow implementation and improves a bit the runtime.

chMoussa:
Closing with #638

PR #638, opened 2024-12-30
[Testing] Reduce testing time of measurement protocols

chMoussa:
Solves #145 by:

- [x] Refactoring tests to use several observables in shadows
- [x] Reduce testing time of tomography. It turns out that QNN tests are a major bottleneck, especially a second-order cross-derivative. Also the `zz_hamiltonian` is quite long for the parameterized test  `test_forward_and_backward_passes_with_qnn`. In my laptop, tests went from 5 minutes to 1 with these changes.

### DIFF
--- a/tests/qadence/test_measurements/test_shadows.py
+++ b/tests/qadence/test_measurements/test_shadows.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import json
 import os
-from typing import Callable
 
 import pytest
 import torch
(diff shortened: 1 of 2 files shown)


tests/qadence/test_measurements/test_tomography.py
```diff
@@ -457,7 +457,7 @@ def test_basic_tomography_for_parametric_circuit_forward_pass(
     "observable, acceptance",
     [
         (total_magnetization(4), MIDDLE_ACCEPTANCE),
-        (zz_hamiltonian(4), MIDDLE_ACCEPTANCE),
```
jpmoutinho, line 460: Assuming this test is still heavy, it could also be changed to 3 qubits instead of 4

jpmoutinho (approved):
Thanks @chMoussa

tests/qadence/test_measurements/test_shadows.py
```diff
@@ -161,17 +152,19 @@ def test_chemistry_hamiltonian_1() -> None:
...
     param_values = {"theta_0": torch.tensor([1.0])}
 
     model = QuantumModel(
         circuit=circuit,
-        observable=hamiltonian,
+        observable=hamiltonians,  # type: ignore[arg-type]
         backend=BackendName.PYQTORCH,
-        diff_mode=DiffMode.GPSR,
+        diff_mode=DiffMode.AD,
     )
```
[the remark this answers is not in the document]
chMoussa: Yes, that was the point.

tests/qadence/test_measurements/test_tomography.py
chMoussa: Still takes 1 minute, mostly due to quite many hamiltonian terms to evaluate.

Issue #618, opened 2024-11-19
[Feature] Add support for Alt Ansatz in config constructors for QNN


smitchaudhary:
### Describe the feature

Add the option to specify the Alt ansatz through config constructors (`QNN.from_configs`)

### It should be implemented because

Since the configs were added before the Alt Ansatz was added in #565, there is currently no way to specify Alt Ansatz via configs. 

### Would you like to work on this issue?

Yes

jpmoutinho:
Thanks @smitchaudhary. Regarding this one and https://github.com/pasqal-io/qadence/issues/617, do you plan to work on it soon?

smitchaudhary:
Hey @jpmoutinho Yeah I am working on it now. #617 first and then this. I will open a draft PR soon already.

smitchaudhary:
Solved in #658

PR #658, opened 2025-01-29
Add config constructor to use QNN with ALA ansatz

smitchaudhary:
Solves #618

### DIFF
--- a/qadence/ml_tools/constructors.py
+++ b/qadence/ml_tools/constructors.py
@@ -601,6 +601,16 @@ def _create_ala_digital(
     num_qubits: int,
     config: AnsatzConfig,
 ) -> AbstractBlock:
+    """
+    Create the Digital Alternating Layer Ansatz based on the configuration.
+
+    Args:
+        num_qubits (int): The number of qubits.
+        config (AnsatzConfig): The configuration for the ansatz.
+
+    Returns:
+        AbstractBlock: The Digital Alternating Layer Ansatz.
+    """
     operations = config.strategy_args.get("operation", [RX, RY, RX])
     entangler = config.strategy_args.get("entangler", CNOT)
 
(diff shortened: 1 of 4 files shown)


smitchaudhary:
Hello hello! I want your opinion on something.

I am implementing this [Alternating Layer Ansatz](https://github.com/pasqal-io/qadence/blob/main/qadence/constructors/ala.py#L13): in configs. For this, we need to do the following:

1. Allow the user to pass `AnsatzType.ALA` if they want this ansatz.
2. Construct it in the `ml_tools.constructors`. 

So far, so good. But as you can see in the above link for ALA, it needs another argument called `m_block_qubits`. And I don't know where I should put that in the `AnsatzConfig` [here](https://github.com/pasqal-io/qadence/blob/main/qadence/ml_tools/config.py#L429):

Because, the AnsatzConfig has a number of arguments, but all of these arguments are valid/necessary for all types of ansatzae. While this new argument is required only for this specific type of ansatz. What is a good design choice in your opinion? Should I make it one of the fields of the `AnsatzConfig` dataclass? And only be concerned with it if it the ALA ansatz? Or do you think some other way would be better?

qadence/ml_tools/constructors.py
```diff
@@ -596,6 +597,35 @@ def _create_hea(
...
+) -> AbstractBlock:
+    if config.ansatz_strategy == Strategy.DIGITAL:
+        return _create_ala_digital(num_qubits=num_qubits, config=config)
+    else:
+        raise ValueError(
+            f"Invalid ansatz strategy {config.ansatz_strategy} provided. Only `Strategy.DIGITAL` \
+                allowed"
+        )
+
+
```
debrevitatevitae, line 628: Is this check necessary, given that the same one happens in the `AnsatzConfig`?

tests/ml_tools/test_qnn.py
```diff
@@ -366,6 +367,44 @@ def test_config_qnn(diff_mode: DiffMode) -> None:
...
+        obs_config=observable_config,
+        fm_config=fm_config,
+        ansatz_config=ansatz_config,
+        diff_mode=diff_mode,
+        backend=backend,
+    )
+
+    assert isinstance(qnn, QNN)
+    assert qnn._diff_mode == diff_mode
+    assert qnn._backend_name == backend
```
debrevitatevitae, line 383: Any way to parametrize this test with the `AnsatzType`?

debrevitatevitae (commented):
Nice work @smitchaudhary , only minor comments :)

qadence/ml_tools/constructors.py
smitchaudhary: If the user always calls the `QNN.from_configs` then it is not necessary. I put these checks since there is no real way of preventing anyone form calling these functions directly.

tests/ml_tools/test_qnn.py
smitchaudhary: I separated the test function instead of parametrizing it because I wanted a bigger register in case of ALA, and it also needs an additional argument in the ansatz config so just a separate test is probably better. WDYT?

debrevitatevitae: Sounds good to me.

debrevitatevitae (approved):

PR #665, opened 2025-02-07
[Refactor] Remove additional run() calls when importing qadence

### LINKED ISSUE #663: [Bug] A few 1-qubit circuits are run on qadence import
### Short description

This is a weird one. Apparently on importing qadence there are a few actual circuit executions on the PyQ backend? I randomly found this while debugging something else. To reproduce it, try the following:

- Go to `qadence/execution.py`, look for line 79 just after `with no_grad():` and before calling `bknd.run(...` and add e.g. `print("HELLO")`.
- Open python and simply run `import qadence`.
- You should see the print four times, which means it is running something on `bknd.run(...)` 4 times...
- If you instead do `print(conv.circuit)` you will see that it runs `I(0)`, `I(0)`, `X(0)`, `X(0)` on PyQ.
- This seems to be only in the `run`, and not on `sample` or `expectation`.

Suggestion from @awennersteen: add a `breakpoint()` and when dropped into the PDB session just use the `up` command to go backwards until you see where this comes from.

### What is the expected result?

No execution on import!

### What is the actual result?

_No response_

### Steps/Code to reproduce

Added above

### Tracebacks (optional)

_No response_

### Environment details (optional)

_No response_

None

sungwoo-pasqal:
Closes #663 

Remove additional run() calls by deleting `P0_MATRIX` and `P1_MATRIX`

### DIFF
--- a/qadence/utils.py
+++ b/qadence/utils.py
@@ -290,5 +290,3 @@ def one_qubit_projector_matrix(state: str) -> Tensor:
 
 P0 = partial(one_qubit_projector, "0")
 P1 = partial(one_qubit_projector, "1")
-P0_MATRIX = one_qubit_projector_matrix("0")
-P1_MATRIX = one_qubit_projector_matrix("1")


RolandMacDoland (approved):
NIce ! LGTM.

PR #679, opened 2025-02-14
qcnn_constructor

arthurfaria:
Hello everyone,
I'm pushing the QCNN constructor. Please check it first and, if everyone agrees, I'll update the documentation accordingly. 
Cheers!

### DIFF
--- a/qadence/constructors/qcnn.py
+++ b/qadence/constructors/qcnn.py
@@ -1,0 +1,370 @@
+class qcnn(QNN):
+    def __init__(
+        self,
+        n_inputs: int,
+        n_qubits: int,
+        depth: list[int],
+        operations: list[Any],
+        entangler: Any,
+        random_meas: bool,
+        use_dagger: bool = True,  # True for using the adjoint operation
+        **kwargs: Any,
+    ) -> None:
+        """
+        Initialize the qcnn class.
+
+        Args:
+            n_inputs (int): Number of input features.
+            n_qubits (int): Total number of qubits.
+            depth (list[int]): List defining the depth (repetitions) of each layer.
+            operations (list[Any]): List of quantum operations to apply in the gates
+                (e.g., [RX, RZ]).
+            entangler (Any): Entangling operation, such as CNOT or CZ.
+            random_meas (bool): If True, applies random weighted measurements.
+            use_dagger (bool): If True, includes the corresponding adjoint operations
+                (gates with negative angles).
+            **kwargs (Any): Additional keyword arguments for the parent QNN class.
+        """
+        self.n_inputs = n_inputs
+        self.n_qubits = n_qubits
+        self.depth = depth
+        self.operations = operations
+        self.entangler = entangler
+        self.random_meas = random_meas
+        self.use_dagger = use_dagger  # Store the flag
+
+        # Create circuit and observables
+        circuit, last_target_qubits = self.create_circuit(
+            self.n_inputs, self.n_qubits, self.depth, self.operations, self.entangler
+        )
+        obs = self.create_obs(self.n_qubits, self.random_meas, last_target_qubits)
+
+        super().__init__(
+            circuit=circuit,
+            observable=obs,
+            backend=BackendName.PYQTORCH,
+            diff_mode=DiffMode.AD,
+            inputs=[f"\u03C6_{i}" for i in range(self.n_inputs)],
+            **kwargs,
+        )
+
+    def get_param(self, params: dict, layer: int, rep: int, pos: int) -> Parameter:
+        """
+        Retrieves or creates a parameter key for the given layer, repetition, and position.
+
+        Args:
+            params (dict): Dictionary to store and retrieve parameters.
+            layer (int): The index of the current layer.
+            rep (int): The index of the current repetition in the layer.
+            pos (int): Position of the qubit in the layer.
+
+        Returns:
+            Parameter: The created or retrieved parameter.
+        """
+        key = f"\u03B8_{layer}_{rep}_{pos}"
+        if key not in params:
+            params[key] = Parameter(key)
+        return params[key]
+
+    def create_gate_sequence(
+        self,
+        params: dict,
+        operations: list[Any],
+        entangler: Any,
+        layer: int,
+        rep: int,
+        control: int,
+        target: int,
+        spacing: int = 0,
+        n_qubits: int = 8,
+    ) -> AbstractBlock:
+        """
+        Creates a sequence of gates for a control-target pair.
+
+        Args:
+            params (dict): Dictionary to store and retrieve parameters.
+            operations (list[Any]): List of gate operations to apply (e.g., [RX, RZ]).
+            entangler (Any): Entangling operation, such as CNOT or CZ.
+            layer (int): The index of the current layer.
+            rep (int): The index of the current repetition in the layer.
+            control (int): Index of the control qubit.
+            target (int): Index of the target qubit.
+            spacing (int, optional): Number of qubits to include as padding. Defaults to 0.
+            n_qubits (int, optional): Total number of qubits. Defaults to 8.
+
+        Returns:
+            AbstractBlock: The sequence of gates as a quantum block.
+        """
+        pad = [
+            I(q)
+            for q in range(control - spacing, control + spacing + 1)
+            if q != control and q != target and 0 <= q < n_qubits
+        ]
+        gates = []
+        # Apply user-defined operations
+        for op in operations:
+            gates.append(
+                kron(
+                    *pad,
+                    op(control, self.get_param(params, layer, rep, control)),
+                    op(target, self.get_param(params, layer, rep, target)),
+                )
+            )
+        gates.append(entangler(control, target))
+
+        # Conditionally apply negative gates in reverse order
+        if self.use_dagger:  # Only apply negative gates if use_dagger is True
+            for op in reversed(operations):
+                gates.append(
+                    kron(
+                        *pad,
+                        op(control, -self.get_param(params, layer, rep, target)),
+                        op(target, -self.get_param(params, layer, rep, control)),
+                    )
+                )
+            gates.append(entangler(control, target))
+
+        return chain(*gates)
+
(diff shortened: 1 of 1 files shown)


qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,403 @@
+from __future__ import annotations
+from typing import Any
+
+import torch
+from qadence import (
```
jpmoutinho, line 5: Importing directly from qadence namespace should be done only by users. In the source code it's best to keep the full module path as much as possible to avoid circular imports.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,403 @@
...
+    Z,
+    add,
+    chain,
+    feature_map,
+    kron,
+    tag,
+)
+
+####
+class qcnn(QNN):
```
jpmoutinho, line 25: Classes should follow CamelCase convention.

Also, this is a model, so maybe it should go in the models submodule? (which currently is in `ml_tools` hosting only the `QNN`, so maybe some further reorganization would be needed)

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,403 @@
...
+            w1 = [Parameter(f"w{i}") for i in range(n_qubits)]
+            obs = add(Z(i) * w for i, w in zip(range(n_qubits), w1))
+        else:
+            obs = add(Z(i) for i in range(n_qubits))
+
+        print(obs)
+        return obs
+
+
+class qcnn_msg_passing(torch.nn.Module):
```
jpmoutinho, line 354: Also CamelCase convention

jpmoutinho (commented):
Thanks, just quick comments from my side.

I feel like there might be some decoupling needed in the code between what is a `constructor` (i.e., a composition of blocks) and a `model` (the usage of blocks in a certain manner / for a certain application). But that probably only makes sense in case there are construction functions here that would be useful for other models.

qadence/constructors/qcnn.py
RolandMacDoland: I fully agree. Maybe the QCNN should move to `./ml_tools/models.py` and here only stands a function that returns an instance of it ?

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,403 @@
...
+                        *pad,
+                        op(control, -self.get_param(params, layer, rep, target)),
+                        op(target, -self.get_param(params, layer, rep, control)),
+                    )
+                )
+            gates.append(entangler(control, target))
+
+        return chain(*gates)
+
+    def create_layer(
```
RolandMacDoland, line 153: Maybe I'm wrong but couldn't you use the `hea` constructor here ?

RolandMacDoland (commented):
Thanks @arthurfaria. I think what @jpmoutinho makes total sense. I guess the best way to proceed is to move the QCNN class to `ml_tools` and only retain here a function that constructs the sequence of your QCNNs. Does it make sense ?

qadence/constructors/qcnn.py
RolandMacDoland: Hm, I guess that's up to you. The `QuantumModel` -> `QNN` -> `QCNN` lineage somehow makes sense to me.

debrevitatevitae: Also, maybe we can add here a reference to the paper where this implementation comes from?

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+        self.n_inputs = n_inputs
+        self.n_qubits = n_qubits
+        self.depth = depth
+        self.operations = operations
+        self.entangler = entangler
+        self.random_meas = random_meas
+        self.use_dagger = use_dagger  # Store the flag
+
+        # Create circuit and observables
+        circuit, last_target_qubits = self.create_circuit(
```
debrevitatevitae, line 61: Creating circuit and observable here introduces unnecessary coupling between components of `QNN` and the `QNN` itself.
 See the comment below about separating the constructor.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+        super().__init__(
+            circuit=circuit,
+            observable=obs,
+            backend=BackendName.PYQTORCH,
+            diff_mode=DiffMode.AD,
+            inputs=[f"\u03C6_{i}" for i in range(self.n_inputs)],
+            **kwargs,
+        )
+
+    def get_param(self, params: dict, layer: int, rep: int, pos: int) -> Any:
```
debrevitatevitae, line 75: This should be a private method (`_get_param`), since it’s only used in this class.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+        Args:
+            params (dict): Dictionary to store and retrieve parameters.
+            layer (int): The index of the current layer.
+            rep (int): The index of the current repetition in the layer.
+            pos (int): Position of the qubit in the layer.
+
+        Returns:
+            Parameter: The created or retrieved parameter.
+        """
+        key = f"\u03B8_{layer}_{rep}_{pos}"
```
debrevitatevitae, line 88: Unclear parameter names.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+        if key not in params:
+            params[key] = Parameter(key)
+        return params[key]
+
+    def create_gate_sequence(
```
debrevitatevitae, line 93: Can be a private method

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+        Args:
+            params (dict): Dictionary to store and retrieve parameters.
+            operations (list[Any]): List of gate operations to apply (e.g., [RX, RZ]).
+            entangler (Any): Entangling operation, such as CNOT or CZ.
+            layer (int): The index of the current layer.
+            rep (int): The index of the current repetition in the layer.
+            control (int): Index of the control qubit.
+            target (int): Index of the target qubit.
+            spacing (int, optional): Number of qubits to include as padding. Defaults to 0.
+            n_qubits (int, optional): Total number of qubits. Defaults to 8.
```
debrevitatevitae, line 117: Why default to 8?

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+                        *pad,
+                        op(control, -self.get_param(params, layer, rep, target)),
+                        op(target, -self.get_param(params, layer, rep, control)),
+                    )
+                )
+            gates.append(entangler(control, target))
+
+        return chain(*gates)
+
+    def create_layer(
```
debrevitatevitae, line 153: Long function. Consider splitting

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+
+        Returns:
+            tuple[AbstractBlock, list[int]]: A tuple containing the quantum block for
+                the layer and the target indices for the next layer.
+        """
+        current_layer = []
+        next_indices = []  # To store the targets for the next layer
+        spacing = layer_index  # Define spacing based on layer index
+
+        if layer_index in [0, 1]:  # Special behavior for first two layers
```
debrevitatevitae, line 183: This special case can be isolated into another (private) function.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+                        spacing=spacing,
+                        n_qubits=n_qubits,
+                    )
+                    current_layer.append(gate_sequence)
+
+        # Update `next_indices` with the **targets** of the current layer
+        next_indices = current_indices[1::2]
+        return chain(*current_layer), next_indices
+
+    def create_circuit(
```
debrevitatevitae, line 236: This function (or functions that compose this one) can be the constructor.
In this way there's no need to create a module specifically for the QCNN and you can directly pass the circuit to a `QNN`.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+            n_qubits (int): Total number of qubits.
+            depth (list[int]): List defining the depth (repetitions) of each layer.
+            operations (list[Any]): List of quantum operations to apply in the gates.
+            entangler (Any): Entangling operation, such as CNOT or CZ.
+
+        Returns:
+            tuple[QuantumCircuit, list[int]]: A tuple containing the quantum circuit
+                and the final target indices.
+        """
+        # Feature map (FM)
```
debrevitatevitae, line 258: Feature map can be a separate function.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+                    fm_type="Fourier",
+                    support=tuple(range(start, end)),
+                )
+            )
+            start = end  # Update starting index for next FM
+
+        fm = kron(*fm_temp)
+        tag(fm, "FM")
+
+        ansatz_layers = []  # To store each layer of the ansatz
```
debrevitatevitae, line 284: Ansatz can be a separate function.

qadence/constructors/qcnn.py
```diff
@@ -0,0 +1,395 @@
...
+            w1 = [Parameter(f"w{i}") for i in range(n_qubits)]
+            obs = add(Z(i) * w for i, w in zip(range(n_qubits), w1))
+        else:
+            obs = add(Z(i) for i in range(n_qubits))
+
+        print(obs)
+        return obs
+
+
+class qcnn_msg_passing(torch.nn.Module):
```
debrevitatevitae, line 348: This class belongs to `ml_tools/models.py`. Make sure to change to CamelCase.

debrevitatevitae: Also, this class holds for generic `nn.Module`s and for more than just QCNNs. Please update the names accordingly.

debrevitatevitae (commented):
Thanks for the contribution @arthurfaria! My main comments are about context of the new constructors (in line with @jpmoutinho and @RolandMacDoland ), length of the methods and coupling.

PR #684, opened 2025-02-21
Fix wrong warning from TrainConfig

### LINKED ISSUE #674: TrainConfig reports wrong log
### Short description

There is an erroneous log when the user provides only `root_folder`, but no `log_folder`.

### What is the expected result?

No log or log different than the one reported below in the issue.

### What is the actual result?

See above.

### Steps/Code to reproduce

MWE:
```python

import torch
from torch.utils.data import DataLoader

from qadence.constructors import hea
from qadence.constructors.feature_maps import feature_map
from qadence.constructors.hamiltonians import hamiltonian_factory
from qadence.ml_tools.config import TrainConfig
from qadence.ml_tools.data import to_dataloader
from qadence.ml_tools.models import QNN
from qadence.ml_tools.trainer import Trainer
from qadence.operations.primitive import Z

n_qubits = 2
ansatz_depth = 1


def dataloader(batch_size: int = 25) -> DataLoader:
    x = torch.linspace(0, 1, batch_size).reshape(-1, 1)
    y = torch.cos(x)
    return to_dataloader(x, y, batch_size=batch_size, infinite=True)


obs = hamiltonian_factory(register=n_qubits, detuning=Z)

data = dataloader()
fm = feature_map(n_qubits, param="x")

model = QNN(
    QuantumCircuit(n_qubits, fm, hea(n_qubits, ansatz_depth)),
    observable=obs,
    inputs=["x"],
)

optimizer = torch.optim.Adam(model.parameters(), lr=0.1)

config = TrainConfig(max_iter=100, checkpoint_every=0, root_folder="./tmp")

trainer = Trainer(model=model, optimizer=optimizer, config=config)
trainer.fit(data)

``` 

### Tracebacks (optional)

<!-- Uploading "Screenshot 2025-02-13 at 16.02.51.png"... —>

### Environment details (optional)

qadence==1.10.1

Yes

debrevitatevitae:
Closes #674

### DIFF
--- a/qadence/ml_tools/config.py
+++ b/qadence/ml_tools/config.py
@@ -112,11 +112,11 @@ class TrainConfig:
     the latest checkpoint if one exists in the specified log folder.
     """
 
-    custom_log_folder: Path | None = None
+    log_folder: Path | None = None
     """The log folder for saving checkpoints and tensorboard logs.
 
     This stores the path where all logs and checkpoints are being saved
-    for this training session. `custom_log_folder` takes precedence over `root_folder`,
+    for this training session. `log_folder` takes precedence over `root_folder`,
     but it is ignored if `create_subfolders_per_run=True` (in which case, subfolders
     will be spawned in the root folder).
     """
@@ -200,14 +200,6 @@ class TrainConfig:
     root folder.
 
     Each subfolder is of structure `<id>_<timestamp>_<PID>`.
-    """
-
-    log_folder: Path = field(init=False)
-    """The final logging folder, defined according to the logic of.
-
-    the `qadence.ml_tools.ConfigManager`.
-
-    Not set by the user.
     """
 
 
--- a/qadence/ml_tools/config.py
+++ b/qadence/ml_tools/config.py
@@ -94,7 +94,7 @@ class TrainConfig:
     model's state will be logged, useful for model versioning and reproducibility.
     """
 
-    root_folder: Path = Path("./qml_logs")
+    root_folder: Path | None = None
     """The root folder for saving checkpoints and tensorboard logs.
 
     The default path is "./qml_logs"
@@ -112,7 +112,7 @@ class TrainConfig:
     the latest checkpoint if one exists in the specified log folder.
     """
 
-    log_folder: Path = Path("./")
+    log_folder: Path | None = None
     """The log folder for saving checkpoints and tensorboard logs.
 
     This stores the path where all logs and checkpoints are being saved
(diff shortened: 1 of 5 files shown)


qadence/ml_tools/config.py
```diff
@@ -112,14 +112,14 @@ class TrainConfig:
...
+    log_folder: Path | None = None
     """The log folder for saving checkpoints and tensorboard logs.
 
     This stores the path where all logs and checkpoints are being saved
     for this training session. `log_folder` takes precedence over `root_folder` and
     `create_subfolder_per_run` arguments. If the user specifies a log_folder,
-    all checkpoints will be saved in this folder and `root_folder` argument
-    will not be used.
+    all checkpoints will be saved in this folder and the`root_folder` and
+    `create_subfolder_per_run` arguments will be ignored.
```
mlahariya, line 122: This is incorrect. 

The hierarchy of arguments is 
1)  create_subfolder_per_run
2) log_folder
3) root_folder. 
4) default root folder

if log_folder is specified - root folder will be ignored. 
if create_subfolder_per_run == True is specified - log_folder will be ignored. 
if root folder is specified, default root folder will be ignored. 

--

Thus - in case log_folder specification does not ignore  create_subfolder_per_run.

mlahariya (commented):
LGTM!

Just minor comments below. Thanks.

qadence/ml_tools/train_utils/config_manager.py
```diff
@@ -175,9 +184,9 @@ def _log_warnings(self) -> None:
                 "`Checkpoint_best_only` is only available when `validation_criterion` is provided."
                 "No checkpoints will be saved."
             )
-        if self.config.log_folder != Path("./") and self.config.root_folder != Path("./qml_logs"):
+        if self.config.log_folder is not None and self.config.root_folder is not None:
```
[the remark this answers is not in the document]
debrevitatevitae: Fixed in 4f26aef.

qadence/ml_tools/config.py
debrevitatevitae: Fixed in a8f98af.

debrevitatevitae:
@mlahariya, the failed linting job is due to the new type of `log_folder` and `root_folder` (which is `Path | None`). 

I think the problem is that the static type checker cannot see that the `self.config.log_fodler` and `self.config.root_folder`  always become `Path`s after processing with `ConfigManager`.

I was wondering if we could merge `ConfigManager` into `TrainConfig`, that is move all the private methods to `TrainConfig` and handle [config initialization](https://github.com/pasqal-io/qadence/blob/fbcd4759f1b2c3c01fbd7cbb41b31d15b3b135d5/qadence/ml_tools/train_utils/config_manager.py#L37) directly in the `__post_init__` of `TrainConfig`.

WDYT?

mlahariya (approved):
LGTM! 
Thanks.

PR #687, opened 2025-02-25
[Testing] Projector tests

### LINKED ISSUE #224: [Testing] Better testing the projectors
[jpmoutinho](https://github.com/jpmoutinho) [32 minutes ago](https://github.com/pasqal-io/qadence/pull/208#discussion_r1406407929)
Oh man those get a bit... verbose 😅. I would suggest mostly testing the projector in terms of expected results coming out of circuits and observables instead of directly comparing matrices. E.g.:

For some random initial state checking that a random projector as a block in the circuit will lead to a unnormalized wavefunction with only a component in the respective basis state.
Checking that HamEvo with a generator written with only projectors gives the same result as one written with Paulis (e.g., X_proj = |0><1| + |1><0| and you can similarly rewrite the other paulis with the projectors and build an Hamiltonian out of it).
Same thing for using the projectors as the observable in the QuantumModel.

chMoussa:
This closes #224

### DIFF
--- a/tests/qadence/test_matrices.py
+++ b/tests/qadence/test_matrices.py
@@ -76,99 +76,6 @@ def _calc_mat_vec_wavefunction(
     "projector, exp_projector_mat",
     [
         (
-            Projector(bra="1", ket="1", qubit_support=0),
-            torch.tensor([[0.0 + 0.0j, 0.0 + 0.0j], [0.0 + 0.0j, 1.0 + 0.0j]]),
-        ),
-        (
-            Projector(bra="10", ket="01", qubit_support=(1, 2)),
-            torch.tensor(
-                [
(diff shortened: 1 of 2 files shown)


tests/qadence/test_matrices.py
```diff
@@ -76,99 +76,6 @@ def _calc_mat_vec_wavefunction(
 @pytest.mark.parametrize(
```
RolandMacDoland, line 76: Do you think we can also check the multiqubit gate matrices from the projector definition ? Ie. CNOT = Id x proj0 + X x proj1 ?

RolandMacDoland (approved):
Thanks @chMoussa a lonely comment from me otherwise looking good.

PR #693, opened 2025-03-04
set_trainable update

### LINKED ISSUE #535: [Docs] Documentation improvements

sungwoo-pasqal:
closes #535.

- [x] Update [Parameters](https://pasqal-io.github.io/qadence/latest/content/parameters/) documentation
  - [x] `set_trainable` feature
- [ ] Update [QuantumModel](https://pasqal-io.github.io/qadence/latest/content/quantummodels/) documentation
  - [ ] `reset_vparams` feature
  - [ ] getting derivatives w.r.t trainable parameters
  - [ ] obtaining parameter dict at a given training epoch
- [ ] Whole `model config` option documentation

### DIFF
--- a/docs/content/parameters.md
+++ b/docs/content/parameters.md
@@ -51,6 +51,25 @@ print(f"{wf = }") # markdown-exec: hide
 
 Since a batch of input values was passed, the `run` function returns a batch of output states. Note that `FeatureParameter("x")` and `VariationalParameter("x")` are simply aliases for `Parameter("x", trainable = False)` and `Parameter("x", trainable = True)`.
 
+## Switching Between Feature Parameters and Variational Parameters
+
+If you want to switch between **feature parameters** and **variational parameters**, you can use `set_trainable` to modify the `trainable` attribute of the variable. Additionally, `set_trainable` can apply the same `trainable` option to multiple blocks if the input is provided in the format `List[blocks]`.
+
+```python exec="on" source="material-block" result="json"
+from qadence import RX, VariationalParameter, set_trainable
+from qadence.blocks.utils import parameters # markdown-exec: hide
+
+variational_block = [RX(0, VariationalParameter("theta")), RX(1, VariationalParameter("phi"))]
+
+for gate in variational_block: # markdown-exec: hide
+    print(f"trainable is set to {parameters(gate)[0].trainable} in {gate}")  # markdown-exec: hide
+
+feature_block = set_trainable(variational_block, False)
+
+for gate in feature_block: # markdown-exec: hide
+    print(f"trainable is set to {parameters(gate)[0].trainable} in {gate}")  # markdown-exec: hide
+```
+
 ## Multiparameter expressions and analog integration
 
 The integration with Sympy becomes useful when one wishes to write arbitrary parameter compositions. Parameters can also be used as scaling coefficients in the block system, which is essential when defining arbitrary analog operations.
--- a/qadence/transpile/block.py
+++ b/qadence/transpile/block.py
@@ -50,7 +50,7 @@ def set_trainable(
         blocks (AbstractBlock | list[AbstractBlock]): Block or list of blocks for which
             to set the trainable attribute
         value (bool, optional): The value of the trainable attribute to assign to the input blocks
-        inplace (bool, optional): Whether to modify the block(s) in place or not. Currently, only
+        inplace (bool, optional): Whether to modify the block(s) in place or not. Currently, only `True` is available.
 
     Raises:
         NotImplementedError: if the `inplace` argument is set to False, the function will


sungwoo-pasqal:
Can you review this PR @jpmoutinho? I also have a question regarding the whole `model config` in the issue. Could you share what this is referring to? I found multiple `config.py` files within the Qadence folder tree, and most of the models themselves come from either the QNN class or Torch. I’d like to confirm whether the `whole config` is meant to cover a broad range or if there is a specific target class in focus.

docs/content/quantummodels.md
```diff
@@ -63,19 +63,22 @@ print(f"ex = {ex.detach()}") # markdown-exec: hide
...
+import numpy as np
+
 # To pass onto a torch optimizer
 parameter_generator = model.parameters()
 
 # Number of variational parameters
 num_vparams = model.num_vparams
 
-# Dictionary to easily inspect variational parameter values
-vparams_values = model.vparams
```
jpmoutinho, line 76: We shouldn't delete this instruction, it shows how to use the `vparams` property.

docs/content/quantummodels.md
```diff
@@ -63,19 +63,22 @@ print(f"ex = {ex.detach()}") # markdown-exec: hide
...
+print(f"old {model.vparams = }") # markdown-exec: hide
 
-print(f"{vparams_values = }") # markdown-exec: hide
+# To reset current variational parameter to other values
+model.reset_vparams([np.random.rand()])
+print(f"new {model.vparams = }") # markdown-exec: hide
```
jpmoutinho, line 81: Maybe you can showcase this using a rand from torch instead of importing numpy?

jpmoutinho:
> Can you review this PR @jpmoutinho? I also have a question regarding the whole `model config` in the issue(#535). Could you share what this is referring to? I found multiple `config.py` files within the Qadence folder tree, and most of the models themselves come from either the QNN class or Torch. I’d like to confirm whether the `whole model config` still needs to cover some specific target class or if previous document updates have resolved it.

Yes, what it refers to is the optional `configuration` argument that is available when initializing a `QuantumModel`. This is a field that allows users to pass extra options to their models. The actual options available are backend dependent, and do a variety of things. For example, you can see that each backend folder inside `qadence.backends` has a `config.py` file with a `dataclass`. This is the default backend cofiguration used. If the user wishes to change one of this options, they should pass it in the `configuration` argument. For example, for something running in the PyQ backend, and changing one of the options, I think the syntax would be:
```
model = QuantumModel(..., configuration = {"use_single_qubit_composition": True})
```
However, I don't think any of this is properly documented, and probably deserves some section in the quantum model page.

docs/tutorials/qml/qcl.md
```diff
@@ -107,12 +107,18 @@ for i in range(n_epochs):
...
+    # value_gen = model._params.values()
+    # print(list(value_gen)[:3])
+
     if (i+1) % 20 == 0:
         print(f"Epoch {i+1} - Loss: {loss.item()}")
 
 assert loss.item() < 1e-3
 ```
 
+If you want to check the parameter values during the epochs, you can access and utilize the `model._params` variable as shown in the example above.
```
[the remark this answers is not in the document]
sungwoo-pasqal: Thanks for the comment. It looks like `model.vparams` is properly retrieving variational parameters from `model._params` using `@property`. Currently, I have restructured it with `@property` to allow external access to all variables, not just the variational ones.
Additionally, If you think that a setter or any direct modification of parameters is necessary, please feel free to suggest it. From what I have reviewed so far, adding this might improve convenience, but I am cautious as it could potentially disrupt relationships between other functions.

sungwoo-pasqal:
Hi @jpmoutinho. Could you review this PR again? Please let me know for any more comments.

docs/content/quantummodels.md
```diff
@@ -63,7 +63,7 @@ print(f"ex = {ex.detach()}") # markdown-exec: hide
 
 By default, the `forward` method of `QuantumModel` calls `model.run()`. To define custom quantum models, the best way is to inherit from `QuantumModel` and override the `forward` method, as typically done with custom PyTorch Modules.
 
-The `QuantumModel` class provides convenience methods to manipulate parameters. Being a `torch.nn.Module`, all torch methods are also available.
+The `QuantumModel` class provides convenience methods to manipulate parameters. Being a `torch.nn.Module`, all torch methods are also available. From below example, you can pass, check and reset the model parameters. When entering new values for the `VariationalParameter`, they must match the number of existing variables.
```
jpmoutinho, line 66:
```suggestion
The `QuantumModel` class provides convenience methods to manipulate parameters. Being a `torch.nn.Module`, all torch methods are also available. As shown in the example below, you can pass, check and reset the model parameters. When entering new values for the `VariationalParameter`, they must match the number of existing variables.
```

docs/content/quantummodels.md
```diff
@@ -72,10 +72,60 @@ parameter_generator = model.parameters()
...
+# Choose your backend (PYQTORCH or PULSER)
+backend=BackendName.PYQTORCH
+# backend=BackendName.PULSER
+
+model = QuantumModel(circuit, backend=backend, diff_mode=DiffMode.GPSR)
+config = model.backend.backend.config
+
+# Check backend confugration options
+bknd_options = BackendConfiguration.available_options(config)
+print(f"{bknd_options = }") # markdown-exec: hide
```
jpmoutinho, line 115: I would suggest adding a `@property` to the `QuantumModel` class for example called `config_options` that automatically gives the user this information, instead of the user having to write all of this. I would do that in this MR since it's a quick add and then you can already fix it here in the docs.

jpmoutinho: Also, could be useful to get a method to change the config. Then the user could do something like:
```
model = QuantumModel(circuit, backend=backend, diff_mode=DiffMode.GPSR)

# Get the available config options and their current set values
model.current_config

# Change the model configuration
new_config = {"dropout_probability": 0.1}
model.change_config(new_config)
```
Seems more natural then having to initialize a model just to check the available options, and then having to reinitialize it just to change them.

docs/content/quantummodels.md
```diff
@@ -72,10 +72,60 @@ parameter_generator = model.parameters()
...
+```
+
+The configuration of the quantum model can be changed by passing `options_names` and `value` in dictionary format. The detailed descriptions are available in [PyQTorch Backend Configuration](https://github.com/pasqal-io/qadence/blob/main/qadence/backends/pyqtorch/config.py) and [Pulser Backend Configuration](https://github.com/pasqal-io/qadence/blob/main/qadence/backends/pulser/config.py) here.
```
jpmoutinho, line 118: Maybe linking to the `.py` source code files is not the best. Is it needed? For example, isn't the information the user will get when running the code above already sufficient to know all the possible options? In that case I would remove the links and then just exemplify the interface as done in the snippet below.

jpmoutinho (commented):
Thanks @sungwoo-pasqal. I would suggest one more small change that I think makes the config interface a bit nicer.

docs/content/quantummodels.md
jpmoutinho: If the `change_config` is added I think there should be a test for it.

sungwoo-pasqal: I updated the features with showing and changing config with `QuantumModel` class.

jpmoutinho: Did you perhaps forget to push the changes?

sungwoo-pasqal: Oh I'm doing it now and having issues with previous merges and rebasing. I'll poke you on slack when I'm ready to get reviewed.

PR #694, opened 2025-03-05
Add Support for Tag Attribute in ObservableConfig

minutogiovanni:
- [x] Add support for a tag attribute in the ObservableConfig class.
- [x] Modify the str method of qnn.from_config to include the obs.tag if it is present 
- [x] Add new method in quantummodel to print the mathematical representation of the observable


qadence/model.py
```diff
@@ -568,6 +569,28 @@ def to(self, *args: Any, **kwargs: Any) -> QuantumModel:
             logger.warning(f"Unable to move {self} to {args}, {kwargs} due to {e}.")
         return self
 
+    def to_pauli_list(self) -> str:
```
chMoussa, line 572: @minutogiovanni Shall we call it rather `observables_to_expression`? The observable is not necessarily a pauli operator (as you can use any list of AbstractBlocks). The rest is fine to me

chMoussa (changes requested):

qadence/model.py
minutogiovanni: e96326fd2a2b9e50f5b28692e297b41ef2587cb4

chMoussa (approved):

Issue #676, opened 2025-02-14
[Feature] Add QCNN constructor and docs


RolandMacDoland:
### Describe the feature

For the purposes of the Blaise Pascal challenge we'd like:
- [x] To implement a QCNN constructor in the `/constructors` folder.
- [ ] Add a corresponding docs section.

### Would you like to work on this issue?

None

PR #696, opened 2025-03-05
issue 676

arthurfaria:
Hello everyone,
I've submitted a PR to address this issue: https://github.com/pasqal-io/qadence/issues/676#event-16310565698
I'm pushing the QCNN constructor. Please check it first and, if everyone agrees, I'll update the documentation accordingly.
Cheers!

obs: For some reason, the last MR was closed. However, please consider this one now. I'm also following your suggestions about incorporating some of the qcnn functions into ml_tools.constructors, for example.

### DIFF
--- a/qadence/ml_tools/constructors.py
+++ b/qadence/ml_tools/constructors.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Callable
+from typing import Any, Callable
 import numpy as np
 from sympy import Basic
 
@@ -8,7 +8,8 @@ from qadence.backend import BackendConfiguration
 from qadence.blocks import chain, kron
 from qadence.blocks.abstract import AbstractBlock
 from qadence.blocks.composite import ChainBlock, KronBlock
-from qadence.blocks.utils import add, tag
+from qadence.blocks.utils import tag, add
+from qadence.parameters import Parameter
 from qadence.circuit import QuantumCircuit
 from qadence.constructors import (
     analog_feature_map,
@@ -24,7 +25,7 @@ from qadence.constructors.hea import hea_digital, hea_sDAQC
 from qadence.constructors.iia import iia
 from qadence.measurements import Measurements
 from qadence.noise import NoiseHandler
-from qadence.operations import CNOT, RX, RY, I, N, Z
+from qadence.operations import CNOT, RX, RY, RZ, I
 from qadence.register import Register
 from qadence.types import (
     AnsatzType,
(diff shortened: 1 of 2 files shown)


qadence/ml_tools/constructors.py
debrevitatevitae, line 1: The `__<method>`s should be single underscore (`_<method>`)

qadence/ml_tools/models.py
```diff
@@ -505,3 +510,81 @@ def _from_dict(cls, d: dict, as_torch: bool = False) -> QNN:
             logger.warning(f"Unable to deserialize object {d} to {cls.__name__} due to {e}.")
 
         return qnn
+
+
+class QCNN(QNN):
```
debrevitatevitae, line 515: Correct me if I’m wrong, but do we really need the `QCNN` class? Why not:
1. Build `QCNN_circuit` in the main script
2. Pull out `qcnn_deferred_obs` from this class and build it also in the main script.
3. Pass circuit and observable to the standard QNN model.

qadence/ml_tools/constructors.py
debrevitatevitae, line 1: IMO, functions longer than 50 lines (rule of thumb) should be split.

debrevitatevitae (commented):
Cool stuff @arthurfaria ;) Left some comments

qadence/ml_tools/constructors.py
arthurfaria: ok, but to which function are referring to? Can you point it to me please?

qadence/ml_tools/models.py
```diff
@@ -20,6 +20,11 @@
 from qadence.register import Register
 from qadence.types import BackendName, DiffMode, Endianness, InputDiffMode, ParamDictType
 from qadence.utils import block_to_mathematical_expression
+from qadence.parameters import Parameter
+from qadence.blocks.utils import add
+from qadence.operations import Z
+
+from qadence.ml_tools.constructors import QCNN_circuit
```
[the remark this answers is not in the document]
arthurfaria: Thanks!

qadence/ml_tools/models.py
debrevitatevitae: What I’m thinking is something along these lines. Assuming this is your main script:
```python

from qadence import CZ, QNN, RX, RZ
from qadence.constructors.hamiltonians import qcnn_deferred_obs
from qadence.ml_tools.constructors import QCNN_circuit


def main():
    circuit = QCNN_circuit(
        n_inputs=2,
        n_qubits=6,
        depth=3,
        operations=[RX, RZ],
        entangler=CZ,
        is_opt=False,
    )

    observable = qcnn_deferred_obs(n_qubits=6, random_meas=True)

    qnn = QNN(circuit, observable)

    # train and do whatever else...


if __name__ == "__main__":
    main()
```

As you see, I have no need for a QCNN _model_, but just for the QCNN _circuit_ and _observable_.

qadence/ml_tools/constructors.py
debrevitatevitae: I signaled them into separate comments.

Issue #700, opened 2025-03-07
Deprecation warning in `nx.node_link_data()`


inafergra:
### Short description

We get a deprecation warning coming from the `networkx` package in the `nx.node_link_data()` function. This function is called when running the `._to_dict()` method of the `Register` class. The deprecation warning is pasted below. According to the warning, we need to set `edges="links"` in the arguments to guarantee backwards compatibility.

### Steps/Code to reproduce

```
from qadence import QuantumModel, QuantumCircuit, hea, total_magnetization

ansatz = hea(n_qubits=5)
circ = QuantumCircuit(5, ansatz)
model = QuantumModel(circ, observable=total_magnetization(n_qubits=5))
model.save("/tmp", "ckpt.pt")
```

### Tracebacks (optional)

```
The default value will be `edges="edges" in NetworkX 3.6.

To make this warning go away, explicitly set the edges kwarg, e.g.:

  nx.node_link_data(G, edges="links") to preserve current behavior, or
  nx.node_link_data(G, edges="edges") for forward compatibility.
```

### Would you like to work on this issue?

Yes

PR #702, opened 2025-03-07
[Bug] Fix `networkx` deprecation warning 

inafergra:
Solves #700


inafergra:
The `python3.9` tests fails because the `networkx` version installed there is too old and that does not include the `edge` parameter in the function. @chMoussa any suggestions on how to deal with this? We could just raise this again when we drop support for 3.9

PR #705, opened 2025-03-13
[Refactor] Remove post optimization loss

mlahariya:


chMoussa (approved):

qadence/ml_tools/trainer.py
```diff
@@ -534,11 +534,10 @@ def run_train_batch(
         # Because the loss/metrics are returned before the optimization. To sync
         # model state and current loss/metrics we calculate them again after optimization.
         # This is not strictly necessary.
-        # TODO: Should be removed if loss can be logged at an unoptimized model state
-        with torch.no_grad():
-            post_update_loss_metrics = self.loss_fn(self.model, batch)
+        # with torch.no_grad():
```
RolandMacDoland, line 537: Why not simply removing it ?

RolandMacDoland (commented):
Thanks @mlahariya. Why not simply removing it ?

qadence/ml_tools/trainer.py
mlahariya: Could do that as well. Just removed it.

RolandMacDoland (approved):

PR #716, opened 2025-03-20
Add aGPSR functionality

### LINKED ISSUE #701: [Feature] Implement aGPSR in Qadence
### Describe the feature

I want to be able to use aGPSR as another diffmode in Qadence.

It will enable users to run simulations with better performance.

### Additional context

_No response_

None

vytautas-a:
Incorporated aGPSR functionality into existing GPSR DiffMode.

Closes #701


qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,92 @@
...
+from functools import lru_cache
+from typing import Callable
+
+import numpy as np
+import torch
+from scipy.optimize import minimize
+from torch import Tensor
+
+
+def create_M_matrix(shifts: torch.Tensor, spectral_gaps: torch.Tensor) -> torch.Tensor:
```
RolandMacDoland, line 12: Docstring ?

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    n_eqs = len(spectral_gaps)
+    M = torch.empty((n_eqs, n_eqs), dtype=torch.double)
+    for i in range(n_eqs):
+        for j in range(n_eqs):
+            M[i, j] = 4 * torch.sin(shifts[i] * spectral_gaps[j] / 2)
```
RolandMacDoland, line 17: It feels like a dense matrix, is it ?

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    return M
+
+
+def variance(shifts: Tensor, spectral_gaps: Tensor) -> Tensor:
+    M = create_M_matrix(shifts, spectral_gaps)
+
+    # calculate iverse of M
+    M_inv = torch.linalg.pinv(M)
```
RolandMacDoland, line 25: Isn't it a costly operation?

qadence/backends/parameter_shift_rules.py
```diff
@@ -8,9 +8,17 @@
...
 from qadence.types import PI
 from qadence.utils import _round_complex
+from qadence.backends.agpsr_utils import calculate_optimal_shifts
 
 
-def general_psr(spectrum: Tensor, n_eqs: int | None = None, shift_prefac: float = 0.5) -> Callable:
+def general_psr(
+    spectrum: Tensor,
+    n_eqs: int | None = None,
+    shift_prefac: float | None = 0.5,
```
RolandMacDoland, line 17: Is it necessary to unionize with `None` when defaulted to 0.5 ?

tests/backends/test_gpsr.py
```diff
@@ -267,3 +292,76 @@ def test_hamevo_gpsr(n_qubits: int, generator: AbstractBlock) -> None:
     dfdx_gpsr = torch.autograd.grad(exp_gpsr, xs, torch.ones_like(exp_gpsr), create_graph=True)[0]
 
     assert torch.allclose(dfdx_ad, dfdx_gpsr, atol=GPSR_ACCEPTANCE)
+
+
+@pytest.mark.slow
```
RolandMacDoland, line 297: If it is marked as slow, it won't be picked up by `test_fast` on CI.

tests/metrics.py
```diff
@@ -17,6 +17,7 @@
 JS_ACCEPTANCE = 7.5e-2
 PSR_ACCEPTANCE = 1e-5
 GPSR_ACCEPTANCE = 1e-4
+AGPSR_ACCEPTANCE = 1e-1
```
RolandMacDoland, line 20: Why is it so low ?

RolandMacDoland (commented):
Thanks @vytautas-a. Not much to say apart from minor comments. Also, could you please update the docs ?

qadence/backends/parameter_shift_rules.py
vytautas-a: it is possible to pass `None` and then the corresponding shifts will be calculated using an optimization function

qadence/backends/agpsr_utils.py
vytautas-a: Well, I haven't benchmarked it, but here it's unavoidable - I have to get the inverse, even though sometimes `M` matrix can be singular

qadence/backends/agpsr_utils.py
vytautas-a: yes, it is

tests/metrics.py
RolandMacDoland: hm what second derivative exactly ?

qadence/backends/agpsr_utils.py
vytautas-a: done

tests/backends/test_gpsr.py
vytautas-a: fixed

qadence/types.py
```diff
@@ -180,7 +180,7 @@ class AnsatzType(StrEnum):
     """Alternating Layer Ansatz."""
```
jpmoutinho, line 180: Shouldn't there be an AGPSR diff mode?

jpmoutinho (commented):
Thanks @vytautas-a. It's not clear to me how to pick AGPSR, it seems like it's not a separate diff mode? Also, the current tests are only for normal GPSR right? 

EDIT: sorry I missed one of the test files somehow

tests/backends/test_gpsr.py
```diff
@@ -267,3 +289,75 @@ def test_hamevo_gpsr(n_qubits: int, generator: AbstractBlock) -> None:
...
+    expval = diff_backend.expectation(pyq_circ, pyq_obs, embedding_fn(params, values))
+    dexpval_psr_x = torch.autograd.grad(
+        expval, values["x"], torch.ones_like(expval), create_graph=True
+    )[0]
+    dexpval_psr_xx = torch.autograd.grad(
+        dexpval_psr_x, values["x"], torch.ones_like(dexpval_psr_x), create_graph=True
+    )[0]
+    dexpval_psr_theta = torch.autograd.grad(
+        expval, list(params.values())[0], torch.ones_like(expval)
+    )[0]
```
jpmoutinho, line 357: Ok this kind of answers my previous question. It would be better to have AGPSR accessible through the `QuantumModel` directly.

qadence/types.py
vytautas-a: Well, that's up for debate. Currently I decided not to introduce a separate diff mode since it's very similar in everything to GSPR. Essentially, you pass `n_eqs` and you use aGPSR.

jpmoutinho: I see, maybe an easy option could be to have the `n_eqs` added to the possible model configs, and have it by default set to use the full GPSR.

jpmoutinho: Similarly to how you can change the solver method for the time-dependent blocks.

tests/metrics.py
RolandMacDoland: Oh right, so PRS can be used to compute higher order derivatives ? I didn't know that. OK thanks.

qadence/types.py
RolandMacDoland: Hm, I'd rather not silently fall back onto smtg a user might not _explicitly_ asked for if I'm totally honest.

RolandMacDoland:
Thanks @jpmoutinho for the hints. @vytautas-a do you think you can push that next sprint and close it off ?

qadence/types.py
jpmoutinho: Agreed but if I am understanding the code correctly, and since there is no explicit `aGPSR` diff mode, it seems like this would be an ok usage of the model config. If the user asks for `GPSR` they get the full `GPSR`, and they want the approximate one, they change the model config with the number of equations they would like to use.

RolandMacDoland: Alright but it has to be clearly documented.

vytautas-a: @RolandMacDoland, what do you mean?

vytautas-a:
Added (a)GPSR args to the backend configs - now we have separate explicit `AGPSR` and `GPSR` diff modes that work with helper functions as well as `QuantumModel`.

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,112 @@
...
+        spectral_gaps (Tensor): tensor containing spectral gap values
+
+    Returns:
+        Tensor: 2D M matrix
+    """
+    n_eqs = len(spectral_gaps)
+    M = torch.empty((n_eqs, n_eqs), dtype=torch.double)
+    for i in range(n_eqs):
+        for j in range(n_eqs):
+            M[i, j] = 4 * torch.sin(shifts[i] * spectral_gaps[j] / 2)
```
jpmoutinho, line 28: Isn't this something like `M = 4 * torch.sin(torch.outer(shifts, spectral_gaps)/2)`

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    Returns:
+        Tensor: variance tensor
+    """
+    M = create_M_matrix(shifts, spectral_gaps)
+
+    # calculate iverse of M
+    M_inv = torch.linalg.pinv(M)
+
+    # calculate variance of derivative estimation
+    a = torch.matmul(spectral_gaps.reshape(1, -1), M_inv)
```
jpmoutinho, line 48: Since what you actually need is $M^{-1}.b$, you should be able to use a linear solver instead. It's usually much more efficient.

https://www.johndcook.com/blog/2010/01/19/dont-invert-that-matrix/

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    # specify constraints for solutions to be unique
+    for i in range(n_eqs - 1, 0, -1):
+        for j in range(i):
+
+            def fn(x, i=i, j=j):  # type: ignore [no-untyped-def]
+                return np.abs(x[i] - x[j]) - 0.02
+
+            constraints.append({"type": "ineq", "fun": fn})
+
+    init_guess = torch.linspace(lb, ub, n_eqs)
```
jpmoutinho, line 98: I also feel like these looks could be optimized but maybe that's not so important since it's just computing the initial guess.

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,112 @@
...
+        res = minimize(
+            fun=var_fn,
+            x0=init_guess,
+            args=(spectral_gaps,),
+            method="COBYLA",
+            constraints=constraints,
+        )
+        return torch.tensor(res.x)
+
+    return minimize_variance(variance)
```
jpmoutinho, line 112: Since you are running the `variance` function on a minimization loop I think the previous comments on the outer and skipping the inverse could make a big difference in performance here.

qadence/engines/torch/differentiable_backend.py
```diff
@@ -75,11 +75,20 @@ def expectation(
...
+            if self.diff_mode == DiffMode.AGPSR:
+                assert (
+                    self.psr_args.get("n_eqs") is not None
+                ), "n_eqs argument must not be None for AGPSR differentiation mode"
+
+            # ensure n_eqs is None when GPSR model is selected
+            if self.diff_mode == DiffMode.GPSR:
+                assert (
+                    self.psr_args.get("n_eqs") is None
+                ), "n_eqs must be None for GPSR differentiation mode"
```
jpmoutinho, line 89: Best to raise actual errors instead of using the `assert`.

tests/metrics.py
jpmoutinho: I would suggest making the acceptance as low as you can for the first derivative, and then increasing it only for the second derivative test. Maybe `sqrt(AGPSR_ACCEPTANCE)` would make sense in the second derivative test, or some other expression if you know how the error scales with multiple applications of the method.

jpmoutinho (commented):
Thanks @vytautas-a! A few more comments from my side.

qadence/backends/agpsr_utils.py
vytautas-a: yes, but I don't solve a linear system here, in contrast with the actual (a)GPSR algo.

qadence/backends/agpsr_utils.py
vytautas-a: correct, changed that

qadence/backends/agpsr_utils.py
vytautas-a: yeah, I wouldn't bother for now with this function - in most cases users will use predefined shifts

tests/metrics.py
vytautas-a: done

qadence/backends/agpsr_utils.py
vytautas-a: done

qadence/backends/agpsr_utils.py
vytautas-a: done

qadence/backends/agpsr_utils.py
vytautas-a: done

vytautas-a:
@RolandMacDoland, pipeline passes now - could you take a look?

qadence/backends/parameter_shift_rules.py
```diff
@@ -93,30 +109,39 @@ def multi_gap_psr(
...
 
-    # get shift values
-    shifts = shift_prefac * torch.linspace(
-        PI / 2 - PI / 4, PI / 2 + PI / 5, n_eqs
-    )  # breaking the symmetry of sampling range around PI/2
+    # get shift values - values minimize the variance of expectation
+    if shift_prefac is not None:
+        # set shift values manually
+        shifts = shift_prefac * torch.linspace(
+            PI / 2 - PI / 4, PI / 2 + PI / 5, n_eqs
```
RolandMacDoland, line 139:
```suggestion
            PI / 4, 7 * PI / 10, n_eqs
```
It would be good to elaborate a little more on where these numbers come from.

qadence/backends/pulser/config.py
```diff
@@ -66,6 +66,24 @@ class Configuration(BackendConfiguration):
     FIXME: To be deprecated.
     """
 
+    n_eqs: int = 2
+    """Number of equations to use in aGPSR calculations."""
+
+    shift_prefac: float = 0.5
+    """Prefactor governing the magnitude of parameter shift values -.
```
RolandMacDoland, line 73: Formatting.

qadence/backends/pyqtorch/config.py
```diff
@@ -70,3 +70,21 @@ class Configuration(BackendConfiguration):
     """Quantum dropout probability (0 means no dropout)."""
     dropout_mode: DropoutMode = DropoutMode.ROTATIONAL
     """Type of quantum dropout to perform."""
+
+    n_eqs: int = 2
+    """Number of equations to use in aGPSR calculations."""
+
+    shift_prefac: float = 0.5
+    """Prefactor governing the magnitude of parameter shift values -.
```
RolandMacDoland, line 78: Same.

qadence/types.py
```diff
@@ -180,11 +180,13 @@ class AnsatzType(StrEnum):
...
 
 
-class _DiffMode(StrEnum):
+class DiffMode(StrEnum):
     """Differentiation modes to choose from."""
 
     GPSR = "gpsr"
     """Basic generalized parameter shift rule."""
+    AGPSR = "agpsr"
```
RolandMacDoland, line 188: I am a little confused now. Is this still necessary ? I mean, we discussed that if you add a value to the `n_eq` argument, you will fall through aGPSR. Therefore, it doesn't feel like this is needed anymore ?

qadence/engines/torch/differentiable_backend.py
```diff
@@ -75,11 +75,20 @@ def expectation(
...
         elif self.diff_mode == DiffMode.ADJOINT:
             expectation = differentiable_expectation.adjoint
-        else:
-            try:
-                fns = get_gpsr_fns()
-                psr_fn = fns[self.diff_mode]
-            except KeyError:
-                raise ValueError(f"{self.diff_mode} differentiation mode is not supported")
-            expectation = partial(differentiable_expectation.psr, psr_fn=psr_fn, **self.psr_args)
+        elif self.diff_mode in [DiffMode.GPSR, DiffMode.AGPSR]:
```
RolandMacDoland, line 78: I don't think this is needed anymore ? Unless I am missing smtg.

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        shifts (Tensor): shifts to apply for each spectral gap
+        spectral_gaps (Tensor): tensor containing spectral gap values
+
+    Returns:
+        Tensor: variance tensor
+    """
+
+    # calculate inverse of M
```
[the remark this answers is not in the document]
vytautas-a: I'll add a comment with paper reference

qadence/backends/agpsr_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    for i in range(n_eqs):
+
+        def fn_lb(x, i=i):  # type: ignore [no-untyped-def]
+            return x[i] - lb
+
+        def fn_ub(x, i=i):  # type: ignore [no-untyped-def]
+            return ub - x[i]
+
+        constraints.append({"type": "ineq", "fun": fn_lb})
+        constraints.append({"type": "ineq", "fun": fn_ub})
```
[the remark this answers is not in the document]
vytautas-a: This is the standard syntax of constraint definition with `scipy`'s `minimize` function. It doesn't matter what's the order of the individual constraint added to the list - `minimize` recognizes them by the actual constraint functions passed in `fun` key.

qadence/engines/torch/differentiable_backend.py
vytautas-a: reverted

qadence/types.py
vytautas-a: removed

qadence/backends/pyqtorch/config.py
vytautas-a: fixed

qadence/backends/pulser/config.py
vytautas-a: fixed

qadence/backends/parameter_shift_rules.py
```diff
@@ -93,30 +109,39 @@ def multi_gap_psr(
...
+        )  # breaking the symmetry of sampling range around PI/2
```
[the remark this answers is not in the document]
vytautas-a: expanded the comment a bit

qadence/backends/agpsr_utils.py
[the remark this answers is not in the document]
RolandMacDoland: Yep, just a quick comment would help.

tests/backends/test_gpsr.py
```diff
@@ -267,3 +287,59 @@ def test_hamevo_gpsr(n_qubits: int, generator: AbstractBlock) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ["n_qubits", "batch_size", "circuit_fn", "shift_prefac", "n_eqs", "lb", "ub"],
+    [
+        (3, 1, circuit_hamevo_tensor_agpsr, 0.5, 10, 0.1, 1.0),
+        (3, 1, circuit_hamevo_tensor_agpsr, None, 10, 0.01, 0.6),
+    ],
+)
+def test_expectation_agpsr(
```
RolandMacDoland, line 299: How much do we expect it to compare with GPSR for a reduced `n_eqs` set ?

RolandMacDoland (approved):
Thanks @vytautas-a. Looking good. Minor comments from me for the sake of discussion mainly.

tests/backends/test_gpsr.py
vytautas-a: what do you mean by this?

RolandMacDoland: Say, if we run this test for n_eqs=3,4,5,6 do we expect to increase the results with respect to the GPSR one ?

vytautas-a: the more equations you include, the closer is the result to GPSR

Issue #220, opened 2023-11-26
[Bug] when using `N` block scaled by some constant with GPSR


vytautas-a:
When generator has `w*N(i)` terms and we want to use GPSR - an error is raised due to lack of `eigenvalues_generator` property for `N` block.

Solution: modify `uuid_to_block` function so that `N` blocks are not assigned uuids and included in the dict of parameters passed to GPSR functions.

dominikandreasseitz:
hm, not that easy @vytautas-a . we need the uuids in the dict, i guess this needs to be refactorted. tagging @Roland-djee

chMoussa:
Solves with #728

PR #728, opened 2025-04-07
[BugFix] N observables and GPSR

chMoussa:
Solves #220


chMoussa:
> I think the general idea makes sense since `N` doesn't have a generator. It might be a matter of taste but checking if `b` has attribute `eigenvalues_generator` could be more elegant.

Actually, it has this attribute but it is implemented as raising the `ValueError`, so that's why we try except.

PR #729, opened 2025-04-08
[Feature] Add Parametric + multiple observables for Adjoint Differentiation

chMoussa:
Solves #185


tests/backends/test_adjoint.py
```diff
@@ -13,21 +13,24 @@
 from qadence.parameters import VariationalParameter
 from qadence.types import DiffMode
 
+list_obs = [[Z(0)], [Z(0), X(1), "xobs" * Z(0)]]
```
RolandMacDoland, line 16: Is it worth randomly generating them ?

tests/backends/test_adjoint.py
```diff
@@ -13,21 +13,24 @@
...
     n_qubits = 2
-    observable: list[AbstractBlock] = [Z(0)]
     circ = QuantumCircuit(n_qubits, chain(RX(0, 3 * "x"), CPHASE(0, 1, "y")))
 
     bknd = backend_factory(backend="pyqtorch", diff_mode=DiffMode.ADJOINT)
     pyqtorch_circ, pyqtorch_obs, embeddings_fn, params = bknd.convert(circ, observable)
 
     inputs_x = torch.rand(batch_size, requires_grad=True)
     inputs_y = torch.rand(batch_size, requires_grad=True)
+    inputs_xobs = inputs_x.clone().detach()
```
RolandMacDoland, line 30: Is there a reason to use `inputs_x` here as opposed to randomly generating them ?

tests/backends/test_adjoint.py
RolandMacDoland: Other point: since it is used multiple times, it can become a fixture.

RolandMacDoland (approved):
Thanks @chMoussa. Minor comments here. Could you also update the docs accordingly ?

tests/backends/test_adjoint.py
chMoussa: Hm probably not worth it. But I made it a fixture

tests/backends/test_adjoint.py
chMoussa: Nope, will randomize it with `requires_grad=False`

PR #735, opened 2025-05-02
[Feature] Separated circuit and observable parameters for AD

### LINKED ISSUE #188: [Feature, Refactoring, Differentiability] Split embedded params of circuit and observable
To support parametric observables for adjoint and GPSR, we can not natively compute gradients w.r.t. observable parameters. Observable parameters are purely classical and they should be treated separately using automatic differentiation. 

### Proposed solution

If we maintain two separate dictionaries for both, we can use AD on the observable. Notice that this refactoring will imply changes into the inner workings of `qadence` parametrization and will likely be quite extensive.

chMoussa:
Closes #188.

So far only for the PyQtorch backend.

### DIFF
--- a/qadence/backend.py
+++ b/qadence/backend.py
@@ -236,7 +236,6 @@ class Backend(ABC):
                 conv_obs.append(c_obs)
 
             def embedding_fn_dict(a: dict, b: dict) -> dict:
-
                 if "circuit" in b:
                     embedding_dict = {"circuit": circ_embedding_fn(a, b), "observables": dict()}
                     if "observables" in b:
@@ -331,7 +330,11 @@ class Backend(ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def assign_parameters(self, circuit: ConvertedCircuit, param_values: dict[str, Tensor]) -> Any:
+    def assign_parameters(
+        self,
+        circuit: ConvertedCircuit,
+        param_values: dict[str, Tensor] | dict[str, dict[str, Tensor]],
+    ) -> Any:
         raise NotImplementedError
 
     @staticmethod
--- a/qadence/backend.py
+++ b/qadence/backend.py
@@ -236,9 +236,16 @@ class Backend(ABC):
                 conv_obs.append(c_obs)
 
             def embedding_fn_dict(a: dict, b: dict) -> dict:
-                embedding_dict = circ_embedding_fn(a, b)
-                for o in obs_embedding_fn_list:
-                    embedding_dict.update(o(a, b))
+
+                if "circuit" in b:
+                    embedding_dict = {"circuit": circ_embedding_fn(a, b), "observables": dict()}
+                    if "observables" in b:
+                        for o in obs_embedding_fn_list:
+                            embedding_dict["observables"].update(o(a, b))
+                else:
+                    embedding_dict = circ_embedding_fn(a, b)
+                    for o in obs_embedding_fn_list:
+                        embedding_dict.update(o(a, b))
                 return embedding_dict
 
             return Converted(conv_circ, conv_obs, embedding_fn_dict, params)
--- a/qadence/backend.py
+++ b/qadence/backend.py
@@ -32,7 +32,6 @@ from qadence.types import (
     Endianness,
     Engine,
     ParamDictType,
-    SeparatedParamDictType,
 )
 
 logger = getLogger(__name__)
@@ -223,7 +222,7 @@ class Backend(ABC):
         if observable is not None:
             observable = observable if isinstance(observable, list) else [observable]
             conv_obs = []
-            obs_embedding_fn_list = []
+            obs_embedding_fns = []
 
             for obs in observable:
                 obs = check_observable(obs)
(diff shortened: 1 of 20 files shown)


qadence/types.py
```diff
@@ -431,6 +431,7 @@ class ReadOutOptimization(StrEnum):
 
 
 ParamDictType = dict[str, ArrayLike]
+SeparatedParamDictType = dict[str, dict[str, ArrayLike]]
```
RolandMacDoland, line 434: Hm, not sure you really need another custom type. Can't you get away with just:

```python
ParamDictType = dict[str, ArrayLike | dict[str, ArrayLike]]
```
?

qadence/backend.py
```diff
@@ -25,7 +25,15 @@
...
 from qadence.parameters import stringify
-from qadence.types import ArrayLike, BackendName, DiffMode, Endianness, Engine, ParamDictType
+from qadence.types import (
+    ArrayLike,
+    BackendName,
+    DiffMode,
+    Endianness,
+    Engine,
+    ParamDictType,
+    SeparatedParamDictType,
```
RolandMacDoland, line 35: Not sure you really need yet another type. See below.

qadence/backend.py
```diff
@@ -228,9 +236,15 @@ def check_observable(obs_obj: Any) -> AbstractBlock:
...
             def embedding_fn_dict(a: dict, b: dict) -> dict:
-                embedding_dict = circ_embedding_fn(a, b)
-                for o in obs_embedding_fn_list:
-                    embedding_dict.update(o(a, b))
+                if "circuit" in b:
+                    embedding_dict = {"circuit": circ_embedding_fn(a, b), "observables": dict()}
+                    if "observables" in b:
+                        for o in obs_embedding_fn_list:
+                            embedding_dict["observables"].update(o(a, b))
+                else:
```
RolandMacDoland, line 244: Do I understand this clause as being "if the passed dict of parameters is not a dict of dicts" ? In that case, I'd make it more explicit because what if a user wants to differentiate over hamiltonian parameters only ?

qadence/backend.py
```diff
@@ -228,9 +236,15 @@ def check_observable(obs_obj: Any) -> AbstractBlock:
...
+                    embedding_dict = circ_embedding_fn(a, b)
+                    for o in obs_embedding_fn_list:
```
RolandMacDoland, line 246:
```suggestion
                    for obs_embedding_fn in obs_embedding_fns:
```

I'd use the plural as opposed to a `_list` suffix.

qadence/backends/horqrux/backend.py
```diff
@@ -135,18 +135,29 @@ def expectation(
             A jax.Array of shape (batch_size, n_observables)
         """
         observable = observable if isinstance(observable, list) else [observable]
-        batch_size = max([arr.size for arr in param_values.values()])
+        if "observables" in param_values or "circuit" in param_values:
+            raise NotImplementedError("The HORQRUX backend does not support separated parameters.")
```
RolandMacDoland, line 139:
```suggestion
            raise NotImplementedError("The Horqrux backend does not support separated parameters.")
```

qadence/backends/horqrux/backend.py
```diff
@@ -135,18 +135,29 @@ def expectation(
...
+        def _expectation(params: ParamDictType | SeparatedParamDictType) -> ArrayLike:
+            param_circuits = params["circuit"] if "circuit" in params else params
+            param_observables = params["observables"] if "observables" in params else params
             out_state = self.run(
-                circuit, params, state, endianness, horqify_state=True, unhorqify_state=False
+                circuit,
+                param_circuits,
+                state,
+                endianness,
+                horqify_state=True,
```
RolandMacDoland, line 153: Fair enough.

qadence/backends/horqrux/backend.py
```diff
@@ -135,18 +135,29 @@ def expectation(
...
+                unhorqify_state=False,
             )
-            return jnp.array([o.native(out_state, params) for o in observable])
+            return jnp.array([o.native(out_state, param_observables) for o in observable])
```
RolandMacDoland, line 156:
```suggestion
            return jnp.array([observable.native(out_state, param_observables) for observable in observables])
```

I guess we don't have to be stingy on names ;)

qadence/backends/pulser/backend.py
```diff
@@ -322,38 +322,45 @@ def expectation(
         self,
         circuit: ConvertedCircuit,
         observable: list[ConvertedObservable] | ConvertedObservable,
-        param_values: dict[str, Tensor] = {},
+        param_values: dict[str, Tensor] | dict[str, dict[str, Tensor]] = {},
```
RolandMacDoland, line 325: No use of a `ParamDictType` ?

qadence/backends/pyqtorch/backend.py
```diff
@@ -200,15 +205,15 @@ def _batched_expectation(
...
+            [obs.native.expectation(state, param_observables).reshape(-1, 1) for obs in observable]
         )
         return _expectation
 
     def _looped_expectation(
         self,
         circuit: ConvertedCircuit,
         observable: list[ConvertedObservable] | ConvertedObservable,
-        param_values: dict[str, Tensor] = {},
+        param_values: dict[str, Tensor] | dict[str, dict[str, Tensor]] = {},
```
RolandMacDoland, line 216: `ParamDictType` ?

qadence/backends/pyqtorch/backend.py
```diff
@@ -242,7 +256,7 @@ def expectation(
...
```
RolandMacDoland, line 259: `ParamDictType` ?

RolandMacDoland (commented):
Thanks @chMoussa. Generally speaking, I don't think the introduction of yet another custom type makes it simpler. Comments from my side.

qadence/types.py
chMoussa: Yes I can!

qadence/types.py
```diff
@@ -430,7 +430,7 @@ class ReadOutOptimization(StrEnum):
     CONSTRAINED = "constrained"
 
 
-ParamDictType = dict[str, ArrayLike]
+ParamDictType = dict[str, Union[ArrayLike, dict[str, ArrayLike]]]
```
RolandMacDoland, line 433: The more I see this, the more I think we should extend `ParamDictType` to a dataclass that will enforce the user passing circuit parameters but would default observables to an empty dict. This would avoid potential syntactic errors while initialising parameters.

RolandMacDoland (approved):
Thanks @chMoussa LGTM. If you agree with my comment, can you raise an issue about it ?

qadence/types.py
chMoussa: Sure, I raised #739

